PR #31, opened 2014-12-10
Allow to run Harbour RPM validation script from Qt Creator

martyone:
[...] Attention: To try this change SDK virtual machine must be unset and
re-set in qtc options. Also existing project configuration is not
touched, so projects must be reconfigured. [...]

### DIFF
--- a/src/plugins/mer/mer.pro
+++ b/src/plugins/mer/mer.pro
@@ -12,6 +12,7 @@ RESOURCES += \
     mer.qrc
 
 FORMS += \
+    mergeneraloptionswidget.ui \
     meroptionswidget.ui \
     mersdkdetailswidget.ui \
     mersdkselectiondialog.ui \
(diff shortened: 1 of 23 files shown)


tkoskinen:
The pull request works as advertised. The _Preferences/Mer_ section gains a new _General_ tab where _Enable RPM validation step initially_ is the only option.

The _Enable RPM validation step initially_ option works as expected, i.e., if checked, newly created projects have the _RPM Validation_ deployment step enabled. If the option is unchecked when a new project is created, the validation step is present in the configuration but is disabled by default.

Also, the implementation looks good to me, aside from a few "this is hack" comments.

Tested on OS X 10.10.3, built against Qt 5.4.1.

rpm-validation-screenshot-1-20150505

rpm-validation-screenshot-2-20150505


__Notes__
* Even though the build scripts use Qt 5.2.1 by default, the pull request was built against Qt 5.4.1 because the lowest version that builds without errors on OS X 10.10 is Qt 5.4.0.
* I was unable to *actually run the RPM validation step*, as I have no device to run it on. The comments here are based on reviewing the code and testing all included functionality except deploying to a device.

jlehtoranta:
Ok, lgtm too.

PR #34, opened 2015-05-25
Resize virtual machine disk image from SDK details

tkoskinen:
The SDK details widget now has a spin box for showing and setting the maximum capacity of a virtual machine disk image file. Resizing is enabled by entering a value greater than the current capacity.

`MerVirtualBoxManager` now has support for querying the current capacity of the primary disk image associated with a virtual machine.

Parts of `MerOptionsWidget` and `MerVirtualBoxManager` were also refactored to take advantage of new private utility methods introduced as a part of this pull request.

**Note:** This feature is not particularly useful until corresponding support for detecting and acting on increased disk capacity is added the build engine virtual machine. For this reason, the resizing functionality can be disabled by defining `MER_NO_BUILD_ENGINE_RESIZE` when building Qt Creator.

Disk image capacity option in SDK details

Illegal capacity warning dialog

Close running virtual machine dialog

### DIFF
--- a/src/plugins/mer/meroptionswidget.cpp
+++ b/src/plugins/mer/meroptionswidget.cpp
@@ -66,6 +66,7 @@ MerOptionsWidget::MerOptionsWidget(QWidget *parent)
     connect(m_ui->sdkDetailsWidget, SIGNAL(sshTimeoutChanged(int)), SLOT(onSshTimeoutChanged(int)));
     connect(m_ui->sdkDetailsWidget, SIGNAL(headlessCheckBoxToggled(bool)), SLOT(onHeadlessCheckBoxToggled(bool)));
     connect(m_ui->sdkDetailsWidget, SIGNAL(srcFolderApplyButtonClicked(QString)), SLOT(onSrcFolderApplyButtonClicked(QString)));
+    connect(m_ui->sdkDetailsWidget, SIGNAL(resizeDiskImageButtonClicked(QString,int,int)), SLOT(onResizeDiskImageButtonClicked(QString,int,int)));
     onSdksUpdated();
 }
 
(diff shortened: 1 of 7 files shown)


JussiPakkanen:
Looking nice. Are there plans when the corresponding build engine resizing lands?

As a minor nitpick, there's this bit which is needlessly complicated:

    QPointer<QMessageBox> questionBox = new QMessageBox(...

The simpler way is to just do this:

    QMessageBox questionBox(...); // parent needs to be nullptr rather than this.

It does not need a delete, and is guaranteed to be destroyed no matter how the scope is exited.

tkoskinen:
Good point about the overly complex way to create `QMessageBox` instances. This was a copy-and-paste relic from the original implementation of the `onSrcFolderApplyButtonClicked()` method.

I simplified the dialog creation to use the style suggested by @JussiPakkanen.

src/plugins/mer/meroptionswidget.cpp
```diff
@@ -249,60 +250,110 @@ void MerOptionsWidget::onSrcFolderApplyButtonClicked(const QString &newFolder)
...
+}
+
+void MerOptionsWidget::onResizeDiskImageButtonClicked(const QString &uuid, int capacity, int newCapacity)
+{
+    MerSdk *sdk = m_sdks[m_virtualMachine];
+
+    if (newCapacity == capacity) {
+        QMessageBox::information(this,
+                                 tr("Choose a Larger Capacity"),
+                                 tr("The given capacity (%1 MB) is the current capacity of the disk image. "
```
thp: Or just return silently for "success" (if the user wants to have the same size that already exists, the resize goes "very quickly")? Even better: Disable the "Resize" button as long as `newCapacity == capacity` to avoid this error condition.

tkoskinen:
@thp is correct in pointing out that the warning dialog about the current and new capacities being the same is redundant.

This implementation was chosen mostly to keep the functionality in line with the other inputs on the same form. For example the source folder "Change" button shows a dialog when preconditions for actually changing the folder are not met. I suspect this is done to avoid storing SDK related values (i.e. the current source folder) in the details widget.

However, since the widget already has all disk image related information, I changed the implementation to the more usable form of simply disabling the "Resize" button until the entered capacity is valid.

PR #35, opened 2015-05-29
Use active project path for deployment

tkoskinen:
For deployment steps, the source path of the active project is now always used as the value of the `MER_SSH_PROJECT_PATH` environment variable instead of the expanded value of the project explorer variable "%{CurrentProject:Path}".

Note that when multiple projects are open, deploying a project via the _Build | Deploy Project "myproject"_ menu action causes a technically incorrect project path value (i.e. the path of the project that contains the directory or file that is currently highlighted in the project explorer view, __not__ the path of the project set as active) to be passed in `MER_SSH_PROJECT_PATH` for the `qmake` and `make` MerSSH commands.

For these two commands, the path set in `MER_SSH_PROJECT_PATH` has to exists but the value seems to be otherwise ignored.


thp:
LGTM.

PR #36, opened 2015-06-03
Mer: Clean up SDK details widget layout (Fixes JB#29309)

tkoskinen:
Made the layout of `mersdkdetailswidget.ui` narrower by splitting the widest form lines “Alt source folder” and “SSH key” into two lines.

As a result, the Preferences > Mer > SDK view now only scrolls vertically, not both vertically and horizontally as before.

Further cleaned up the widget by giving all child widgets and layouts proper names and setting all form labels and form widgets to use identical size policies and layout alignments.

sdk-widget-layout-screenshot-1-20150603

sdk-widget-layout-screenshot-2-20150603

PR #37, opened 2015-06-30
Make web UI port editable in SDK details widget (Fixes JB#30222)

tkoskinen:
Replaced the static web UI port label with a spin box in `MerSdkDetailsWidget` to make it possible for the user to change the port of the build engine web UI.

qt-creator-mer-change-www-port

The initial value for the spin box is the port configured by the installer (8080). The user can specify any port in the range [1024..65535].

No attempt is made to verify that the port is actually free. The SDK configuration and the NAT port forwarding rule of the “MerSDK” virtual machine are updated as soon as the user pressed the “Change” button.

In `MerOptionsWidget`, the new slot `onWwwPortApplyButtonClicked(int)` takes care of closing the virtual machine for the duration of the port update operation and restarting it, if necessary.

The existing slot `onSrcFolderApplyButtonClicked()` was refactored to use the new private utility methods `showCloseVirtualMachineDialog()` and `showStartVirtualMachineDialog()`, as the functionality of these two slots is very similar.

Note that this pull request also includes a narrower layout for the SDK details widget requiring no horizontal scrolling. This fixes [JB#29309](https://bz.jollamobile.com/show_bug.cgi?id=29309) and obsoletes sailfish-sdk/sailfish-qtcreator pull request #36.

### DIFF
--- a/src/plugins/mer/meroptionswidget.cpp
+++ b/src/plugins/mer/meroptionswidget.cpp
@@ -291,36 +291,15 @@ void MerOptionsWidget::onWwwPortApplyButtonClicked(int port)
 {
     MerSdk *sdk = m_sdks[m_virtualMachine];
 
-    const bool vmIsOff = sdk->connection()->isVirtualMachineOff();
-    const bool vmWillBeClosed = !vmIsOff && showCloseVirtualMachineDialog(tr("Virtual machine must be closed before the web port can be changed."));
-    bool updateOk = false;
-
-    if (vmIsOff || vmWillBeClosed) {
-        // Disable the WWW port "Change" button so that it cannot be clicked
-        // multiple times while the port change is in progress.
-        m_ui->sdkDetailsWidget->setWwwPortApplyButtonEnabled(false);
-
-        if (sdk->connection()->lockDown(true)) {
-            updateOk = MerVirtualBoxManager::updateWwwPort(m_virtualMachine, port);
-            sdk->connection()->lockDown(false);
-        }
-
-        if (updateOk) {
-            m_ui->sdkDetailsWidget->setWwwPort(port);
-            sdk->setWwwPort(port);
-
-            if (vmWillBeClosed && showStartVirtualMachineDialog(tr("Web port successfully changed to %1.").arg(port))) {
-                sdk->connection()->connectTo();
-            }
-        } else {
-            QMessageBox::warning(this,
-                                 tr("Changing web port failed!"),
-                                 tr("Unable to change web port to %1").arg(port));
-        }
-    }
-
-    if (!updateOk) {
+    if (MerVirtualBoxManager::updateWwwPort(m_virtualMachine, port)) {
+        m_ui->sdkDetailsWidget->setWwwPort(port);
+        sdk->setWwwPort(port);
+    } else {
         m_ui->sdkDetailsWidget->resetWwwPort();
+
+        QMessageBox::warning(this,
+                             tr("Changing web port failed!"),
+                             tr("Unable to change web port to %1").arg(port));
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/plugins/mer/meroptionswidget.cpp
```diff
@@ -249,60 +250,107 @@ void MerOptionsWidget::onSrcFolderApplyButtonClicked(const QString &newFolder)
...
-                QMessageBox::Yes | QMessageBox::No,
-                this);
-        questionBox->setInformativeText(
-                tr("Virtual machine must be closed before the source folder can be changed."));
-        if (questionBox->exec() != QMessageBox::Yes) {
-            // reset the path in the chooser
-            m_ui->sdkDetailsWidget->setSrcFolderChooserPath(sdk->sharedSrcPath());
-            return;
+    const bool vmIsOff = sdk->connection()->isVirtualMachineOff();
+    const bool vmWasClosed = !vmIsOff && showCloseVirtualMachineDialog(tr("Virtual machine must be closed before the source folder can be changed."));
```
thp: Shouldn't this do another `isVirtualMachineOff()` check after the dialog? `vmWasClosed` is actually misleading, it's true when the user clicked on the affirmative action on the dialog, no closing needs to be done?

src/plugins/mer/meroptionswidget.cpp
```diff
@@ -249,60 +250,107 @@ void MerOptionsWidget::onSrcFolderApplyButtonClicked(const QString &newFolder)
...
         m_ui->sdkDetailsWidget->setSrcFolderChooserPath(sdk->sharedSrcPath());
-        return;
     }
+}
 
-    bool ok = MerVirtualBoxManager::updateSharedFolder(m_virtualMachine,
-            QLatin1String("src1"), newFolder);
+void MerOptionsWidget::onWwwPortApplyButtonClicked(int port)
+{
+    MerSdk *sdk = m_sdks[m_virtualMachine];
```
thp: Can we be sure that `sdk` is never `NULL`? Should we add a check to be sure?

src/plugins/mer/meroptionswidget.cpp
tkoskinen: The `isVirtualMachineOff()` call is only used to find out if the "Virtual machine must be closed..." dialog should be shown to the user.

Another call is unnecessary, as `sdk->connection()->lockDown(true)` will return `true`if the virtual machine is already off, allowing the update to proceed, in case the VM was shut down manually while the dialog was on screen.

True, `vmWasClosed` is slightly misleading at the point of declaration. It gets this meaning on line 274, where it is used to determine if the corresponding "Start 'MerSDK' virtual machine?" dialog should be shown. I renamed the variable to `vmWillBeClosed` to be more clear that no VM shutdown has taken place yet when the variable is declared.

src/plugins/mer/meroptionswidget.cpp
tkoskinen: The value of `m_sdks[m_virtualMachine]` is assumed to be non-NULL throughout the class. 

This would appear to be a fairly safe assumption, as `m_virtualMachine` is set to one of the values of the "Virtual Machine" combo box, which in turn is populated with the return value of `m_sdks.keys()`.

In case the last SDK is removed, `m_virtualMachine` is cleared and the SDK details widget is hidden. Because `onWwwPortApplyButtonClicked()` is called via a signal fired by the (now hidden) SDK details widget, `m_sdks[m_virtualMachine]` will not be accessed if there are no SDKs.

jlehtoranta:
The Virtualbox NAT port forwarding changes can also be done live without shutting down the VM by using the "controlvm" command instead of "modifyvm". A small change, which adds to the user experience.

tkoskinen:
Excellent point about the `controlvm` subcommand.

Updating the web UI port is now done using `VBoxManage controlvm` instead of `VBoxManage modifyvm`. This simplifies the implementation, as all VM shutdown and restart checks related to updating the port value were also removed.

PR #38, opened 2015-07-14
Allow deployment to tablet

martyone:


JussiPakkanen:
A few of the classes (at least merdeploysteps.cpp) include other files before their own header. It is considered good practice to include your own header first to ensure it works properly standalone. Please fix.

In MerPrepareTargetStep::init, line 383 the if else chain may lead to null pointer dereference if a new machine type is added in the future. Please either add a null pointer check after that, print an error and exit gracefully or a Q_ASSERT.

In merdevicefactory.cpp method MerDeviceFactory::create both branches of the ifelse statement begin with the same dialog boilerplate:

            MerHardwareDeviceWizard wizard(ICore::dialogParent());
            if (wizard.exec() != QDialog::Accepted)
                return ProjectExplorer::IDevice::Ptr();

Move it out from the branches to a single definition. Similarly at the end of the ifelse is an empty return statement which is dead code.

This is not related to your changes but the method has three duplicated places where it moves ssh settings from the wizard to an sshconnection object. While we're here let's move the common items (host, username, port, timeout, etc) to a static helper function like copySshSettingsFromWizard(MerEmulatorDeviceWizard &wizard, QSsh::SshConnectionParameters &sshParams) and have the different branches just override the ones that are different. (see also below, too lazy to edit this text again)

meremulatordevice.cpp has #include over its copyright declaration.

meremulatordevicewidget.cpp has weird checks for casts. It checks that the cast works with dynamic_cast but the result is used with static_cast. Shouldn't the cast type be the same for both? (same for mersdkmanager)

And going further, the ssh settings copy code is in MerHardwareDeviceWizardSelectionPage::handleTestConnectionClicked too. Hmmm. Maybe that should have a method that returns a qsshconnectionparameters object instead?

The connect calls at the end of MerHardwareDeviceWizardSetupPage::MerHardwareDeviceWizardSetupPage define signal parameters as QString when they really are const QString &.

Phew. That's it for now. I did not actually try running it, that will happen next.

martyone:
Foreword: In general I prefer not to break git history with code polishing unles it means a significant improvement of some sort (and only do this in separate commit/pull request). Working git-blame is an invaluable tool when trying to understand codes sometimes :)

> A few of the classes (at least merdeploysteps.cpp) include other files before their own header. It is considered good practice to include your own header first to ensure it works properly standalone. Please fix.

Unless I missed something, this is the original code.

> In MerPrepareTargetStep::init, line 383 the if else chain may lead to null pointer dereference if a new machine type is added in the future. Please either add a null pointer check after that, print an error and exit gracefully or a Q_ASSERT.

In both branches m_impl is initialized - what do you mean if not this?

> In merdevicefactory.cpp method MerDeviceFactory::create both branches of the ifelse statement begin with the same dialog boilerplate:
> 
>         MerHardwareDeviceWizard wizard(ICore::dialogParent());
>         if (wizard.exec() != QDialog::Accepted)
>             return ProjectExplorer::IDevice::Ptr();
> 
> Move it out from the branches to a single definition. Similarly at the end of the ifelse is an empty return statement which is dead code.
> 
> This is not related to your changes but the method has three duplicated places where it moves ssh settings from the wizard to an sshconnection object. While we're here let's move the common items (host, username, port, timeout, etc) to a static helper function like copySshSettingsFromWizard(MerEmulatorDeviceWizard &wizard, QSsh::SshConnectionParameters &sshParams) and have the different branches just override the ones that are different. (see also below, too lazy to edit this text again)

Notice there are two wizard classes - it is not really a boiler plate code.

The empty return statement at the end might have been needed by some (broken) compiler.

> meremulatordevice.cpp has #include over its copyright declaration.

This is original code.

> meremulatordevicewidget.cpp has weird checks for casts. It checks that the cast works with dynamic_cast but the result is used with static_cast. Shouldn't the cast type be the same for both? (same for mersdkmanager)

The static_cast does not really do any check here. During testing, the dynamic_cast is active. Same for mersdkmanager.

> And going further, the ssh settings copy code is in MerHardwareDeviceWizardSelectionPage::handleTestConnectionClicked too. Hmmm. Maybe that should have a method that returns a qsshconnectionparameters object instead?

This is the original code. The function originates from the second page from which I moved it.

> The connect calls at the end of MerHardwareDeviceWizardSetupPage::MerHardwareDeviceWizardSetupPage define signal parameters as QString when they really are const QString &.

This is the so called canonical signature (Qt meta type system point of view). It wouldn't be OK to use full C++ signature here.

Jussi, prefer using inline comments :) (available from the diff view).

JussiPakkanen:
Qt docs are poor on this but according to docs (http://doc.qt.io/qt-5/signalsandslots.html) at least pointers should be SIGNAL(foobar(QObject*)) not SIGNAL(foobar(QObject)). This seems to indicate that refs should be SIGNAL(foobar(QObject&)). Their sample code does not have refs.

I tried that and it compiles, also for (const QObject &). I'm not sure what the correct thing here is. There's a newer type safe connection method but I'm not sure if we can use it.

martyone:
It's really not well docummented https://bugreports.qt.io/browse/QTBUG-8961 :) but there are tons of example code to deduce that fact from. If you do not use canonical signature, QObject::connect() will try with QMetaObject::normalizedSignature() which costs some performance. Pointer types must be indicated in a normalized signature, references not - because C++ code needs no change when you substitute an argument of type T with another one of type T&.

This is not subject to compile-time checking. It can only fail at runtime. Regarding memfun based connections, which do compile time type checking, I found (only) the winrt plugin using them (winrt/winrtdevicefactory.cpp) so we may consider to use them in new code.

JussiPakkanen:
I finally got my installation sorted and ran this. Deployment works nicely and everything seems to be working fine. The cleanups mentioned can be done later. I'm going to go through this one more time to check for issues and if nothing is found we can merge.

JussiPakkanen:
When clicking on the start sdk button it starts it in visual mode e.g. the window with the SDK comes up. In the previous version it was started headless. Are my settings screwed up or did this change?

martyone:
This should not have been changed. Please check your settings in Tools->Options->Mer->No Window.

JussiPakkanen:
For some reason that was off. Anyhow, everything looks fine, so let's merge this.

PR #39, opened 2015-07-23
Allow to control emulator's video mode

martyone:
After all the refactorization done with the first two commits in not really needed, but it took some time to find the error so good to record it..

### DIFF
--- a/src/plugins/coreplugin/fancyactionbar.cpp
+++ b/src/plugins/coreplugin/fancyactionbar.cpp
@@ -321,6 +321,7 @@ void FancyActionBar::addProjectSelector(QAction *action)
     FancyToolButton* toolButton = new FancyToolButton(this);
     toolButton->setDefaultAction(action);
     connect(action, SIGNAL(changed()), toolButton, SLOT(actionChanged()));
+    toolButton->actionChanged();
     m_actionsLayout->insertWidget(0, toolButton);
 
 }
@@ -329,6 +330,7 @@ void FancyActionBar::insertAction(int index, QAction *action)
     FancyToolButton *toolButton = new FancyToolButton(this);
     toolButton->setDefaultAction(action);
     connect(action, SIGNAL(changed()), toolButton, SLOT(actionChanged()));
+    toolButton->actionChanged();
     m_actionsLayout->insertWidget(index, toolButton);
 }
 
--- a/src/plugins/coreplugin/fancyactionbar.h
+++ b/src/plugins/coreplugin/fancyactionbar.h
@@ -58,6 +58,7 @@ public:
     void setFader(float value) { m_fader = value; update(); }
 
 private slots:
+    friend class FancyActionBar;
     void actionChanged();
 };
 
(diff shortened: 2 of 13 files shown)


JussiPakkanen:
QMap<QString, QSize> MerEmulatorDevice::availableDeviceModels() const
    {
        static QMap<QString, QSize> modes {
            { tr("Jolla Phone"), QSize(540, 960) },
            { tr("Jolla Tablet"), QSize(1536, 2048) },
        };
        return modes;
    }

Shouldn't this return const QMap<QString, QSize>& instead? Or not define the modes variable as static. Currently it always creates a full copy on every method call.

There are a few uses of the auto keyword. Testing on the build machine says that its gcc (4.4.4) does have c++0x and that the build uses it. Once the trial build finishes we know if it actually works.

JussiPakkanen:
Does not compile:

https://etherpad.jollamobile.com/pastebin/1437655659/

martyone:
It seems I never presses the green button bellow, so here is my late reply :)

> Shouldn't this return const QMap& instead? Or not define the modes variable as static. Currently it always creates a full copy on every method call.

http://doc.qt.io/qt-5/implicit-sharing.html

Also updated with the dereferencing of QPointers.

JussiPakkanen:
The fact that Qt does implicit sharing behind the scenes is not really relevant. The point is that the function returns data that all clients should treat as const as a mutable value. This may lead to bugs where they alter the value by accident. Returning const ref makes this impossible so calling code can't do the wrong thing. If they absolutely must make a copy, they have to do that explicitly.

JussiPakkanen:
Assuming you can do auto x = getAvailableDeviceModels(). A quick grepping says it is not used much in QtC code. It compiled on 1004 but does it work on OSX? Full c++11 support landed only in Mavericks and the build machine is mountain lion AFAICR.

martyone:
> Returning const ref makes this impossible so calling code can't do the wrong thing. If they absolutely must make a copy, they have to do that explicitly.

That's not the "Qt way" of designing APIs :) You would get a return value of a reference type only where performance may be critical http://doc.qt.io/qt-5/qpoint.html - and only when you explicitly require it AFAIK.

martyone:
Why is the "close and comment" button there and so big???

martyone:
> does it work on OSX?

Let's try build it? :)

JussiPakkanen:
If this is established Qt practice then I guess we can go with it. But at least for shared_ptr the performance cost of _not_ passing shared_ptrs by ref can be substantial: https://stackoverflow.com/questions/2502394/the-cost-of-passing-by-shared-ptr

That's probably because it uses atomics. I'm not sure if Qt's containers use those behind the scenes.

martyone:
Yes, Qt containers use atomics http://doc.qt.io/qt-5/threads-modules.html#threads-and-implicitly-shared-classes, but I am sure we are rarely writing such a performance critical code to worry about.

> If this is established Qt practice then I guess we can go with it.

Definitely. I always look at the surrounding code first. (QtC has not the shiniest codes, though:))

martyone:
Merge?

PR #40, opened 2015-07-28
Core: Fix FancyToolButton initial visibility

martyone:
This fixes the case when action is set not visible _before_ adding with
FancyActionBar::insertAction.

(cherry picked from commit 5b7d80c032cda37ac68134a6c81ee704a1a89693)


JussiPakkanen:
LGTM

PR #45, opened 2015-08-11
Mer: Avoid useless auto reload of mgmt web view

martyone:
Observed on Windows. Management web view flashing when SDK VM is not
running. Screen blinks every 5 seconds - page auto refresh interval.

Limit auto reload to the case of VM running but not responding. For
other cases notification signals are available and already handled.


jlehtoranta:
lgtm

PR #51, opened 2015-09-07
Mini target selector: Make the kit area look less like a header

martyone:
With the light background it was not obvious this is an interactive
widget; it looked like a non-interactive header of the popup widget.


jlehtoranta:
lgtm

PR #69, opened 2015-11-19
Wizards

martyone:

PR #70, opened 2015-11-20
Fix segfault: focus SSH key line edit in VM options and hit escape

martyone:
Notice that all but the very last commit belongs to PR #69.


jlehtoranta:
lgtm

PR #79, opened 2016-07-25
Help: Fix font in QtWebKitHelpView too small

martyone:
Most probably no one ever noticed this because Qt docs set font size
explicitly in their CSS
https://code.qt.io/cgit/qt/qtbase.git/tree/doc/global/template/style/offline.css#n2


jlehtoranta:
lgtm

PR #80, opened 2016-07-26
Mer: Query SSU to detect device name

martyone:
This way devices with the same /etc/hw-release but different brand can
be distinguished.


jlehtoranta:
lgtm

PR #87, opened 2016-10-26
Jb7833

martyone:


jpetrell:
Awesome! :) I'll give it a try.

martyone:
@jpetrell you will also need sailfish-sdk/libsailfishapp#23 on your device - I'm sure you noticed

PR #92, opened 2016-12-21
Mer 4.2

Kaffeine:


martyone:
@Kaffeine we would need you to reorganize your changes to follow the merging strategy we have been using so far.

We are using mer-X.Y branches to collect our upstreamable changes on top of upstream X.Y branches.

These are the steps we use for updating to never upstream:

1. Branch mer-X.Y from an upstream tag vX.Y.Z
1. Ch-p our changes from previous mer-X.Y on top of the new one (those which we didn't send upstream yet)
1. Revert our changes to core which we do not need anymore
1. Temporarily revert our changes to core which would lead to merge conflict
1. Merge mer-X.Y
1. Reintroduce changes temporarily reverted prior merging (fixing conflicts)
1. Stabilize with additional changes

Usually this is done on our 'next' branch and merged to 'master' when it is considered stable.

Check this with 'gitk qt/master..sailfish/master'. Look around a01e9b63a161f64893687b64fce1bbb983551a26 (Merge branch 'mer-3.5' into next).

- Example of commit from step 3: a9d025464448adaae5c885dbbaf5a39d7e9231ff (Revert "UpdateInfo: Do not crash randomly when activating timers")
- Example of commit from step 4: 75c129426c1ac6053fb6883d0beea25eacd0c922 (Temporarily revert "Remotelinux: Fix deploy config widget geometry")

If you then look at 'gitk qt/master..kaffeine/mer-4.2' what I do not like is:

1. 'Fixed conflict in ...' commits
1. Upstreamable commits are left back in history (related to point 1)
1. "Merge tag 'v4.2.0' into mer-4.2" is confusing w.r.t the approach being used till now

**So what I would recomment is to follow the approach we have been using up to step 6 and then ch-p all your "Mer: Port ...." commits as step 7.**

Kaffeine:
First of all, as I said in IRC, I'm proposing to update to 4.1 for now — I think that there is a problem with Compiler Language selection in 4.2 port.

I will update this PR once I'll re-check the port to 4.2.

>We are using mer-X.Y branches to collect our upstreamable changes on top of upstream X.Y branches.

By the way, does anyone work on upstreaming?

>Branch mer-X.Y from an upstream tag vX.Y.Z
>Ch-p our changes from previous mer-X.Y on top of the new one (those which we didn't send upstream yet)

So, a new mer branch is not based previous one. Ok.

> Temporarily revert our changes to upstream code which would lead to merge conflict

So, you would like to have changes applied/reverted for every update.

I'll update the #93 with instruction from this commit and after that, I'll work on this PR again.

martyone:
> By the way, does anyone work on upstreaming?

If you mean upstreaming changes to upstream code, then yes. If you mean upstreaming the Mer plugin and related stuff, this is on the list but not actively worked on ATM.

> So, a new mer branch is not based previous one. Ok.

Yes, mer-X.Y is vanilla X.Y with our upstreamable changes only.

> So, you would like to have changes applied/reverted for every update.

Yes, those which would result in conflict with upstream changes. This way you maintain atomicity of all changes to upstream code. (This could be also achieved using an approach based on "ours" merge strategy, but considering the quite long history of our branch I prefer temporary reverting as a more lightweight alternative.)

You will find more details from the Git history. Like that upstream changes to dist/gdb/ are discarded to keep the version which matches the gdb version on device.

> I'll update the #93 with instruction from this commit and after that, I'll work on this PR again.

Thank you.

Kaffeine:
>Ch-p our changes from previous mer-X.Y on top of the new one (those which we didn't send upstream yet)

This operation named "rebase". So instead of actual cherry-pick, we can do
```
git checkout mer-4.0
git checkout -b mer-4.1
git rebase upstream/4.1
```

martyone:
In an ideal case yes. I wish Git has implemented a mechanism like Gerrit's change-ids natively... the commit as accepted by upstream usually differs a bit from the original in our tree, in which case git-patch-id fails delivering the expected results.

martyone:
QtC 4.2 finally arrived to our master with #106

PR #97, opened 2017-02-03
RemoteLinux: Quiet device process termination

martyone:
NOTE: Upstream has already applied this change in scope of
0bef624 (RemoteLinux: Allow killing
processes by ID)

This is to prevent message like this to appear in Application Output
pane when the process is terminated using the Stop button in Application
Output pane:

 bash: line 1:  3477 Terminated  ENV_VAR=val... /usr/bin/my_app
 Application finished with exit code 143.

With this change it would print just this:

 Process killed by signal

Which is not only shorter but also more informative (the application did
not handle the TERM signal).


martyone:
Looking into it now I would prefer NOT merging this to prevent unnecessary merge conflict with 4.2 prepared in 'next' branch.

PR #122, opened 2017-12-11
Merssh: pass --pedantic to mb2

martyone:
With this mb2 will issue a warning when qtc_* macros are missing from
the .spec file.


wurzer:
So, what do I have to do to build the lipstick project? Is it only possible to build th library within the SailfishOS Plattform SDK? 

- [ ] Should I wait for an update of Sailfish QtCreator?
- [ ] Should I install the SailfishOS platform SDK and build the project within this SDK?
- [ ] Could I manually reference the installed pkg-config binary of macOS
- [ ] Could I manually reference the testlib? Is the testlib part of the QtCreator distribution?

I think the lestlib is for unit testing and should be a part of QtCreator as an essential quality assurance approach.

martyone:
Yes, without qtc_* macros in the .spec it can only be build within the platform SDK. No, Sailfish Qt Creator update will not change this. Yes, testlib is available in the application SDK. Proper fix is to add qtc_* macros to the .spec file of lipstick - feel free to open a pull request to the lipstick repository. Check a .spec file of any project created inside your application SDK to learn about these macros (File > New File or Project > Application > Sailfish OS Qt Quick Application).

PR #133, opened 2018-07-25
Fix setting GDB's solib-search-path with relative DESTDIR

martyone:


martyone:
Note that all but the very last commit belong to different PRs

PR #134, opened 2018-08-07
Allow to forward environment variables to build engine

martyone:


martyone:
Note that all but the two very last commits belong to other PRs

PR #135, opened 2018-08-09
Use progress dialogs instead of "retry?" question boxes

martyone:


martyone:
Note that all but the very last three commits belong to other PRs

PR #136, opened 2018-08-20
 Mer: Remove space from "user name" and "host name" 

martyone:


martyone:

PR #137, opened 2018-08-20
 Mer: Ask before starting/closing VM optionally

martyone:


martyone:

PR #138, opened 2018-08-20
Mer: Prevent crash when updating targets while building

martyone:


martyone:

PR #146, opened 2018-10-19
Allow to change VM system parameters

etaishev:
Added the ability to change the size of VDI, memory and CPU count of VMs and dialog for this. Add open button for dialog to each VM dialog in QtCreator.

One more question, when i resize VDI, partition table of VM don't change. Whee need do something or leave it as done in the VirtualBox?


src/plugins/mer/mervirtualmachinedialog.cpp
```diff
@@ -0,0 +1,231 @@
...
+    ui->vdiSizeMbSlider->setValue(vdiCapacity);
+
+    ui->sizeOnDiskLabel->setText(QString("%1 Gb").arg(QString::number(currentVdi.sizeOnDisk / 1024.0, 'f', 3)));
+    ui->vdiPathEdit->setText(currentVdi.path);
+}
+
+int getTotalMemorySize()
+{
+    // default for now return 3.5GB
+    return 3584;
```
etaishev: If really need it, can do the calculation of the available PC memory

QT does not have its own implementation for this, it will be platform dependent code for win, OS X and linux

martyone:
@etaishev I had the review prepared before you updated the commit but due to some temporary failure on github side I couldn't submit until now, so navigation to the source from the comments is unfortunately broken - please check if those are still valid.

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -502,6 +525,91 @@ void MerVirtualBoxManager::setVideoMode(const QString &vmName, const QSize &size
...
+    arguments.append(QLatin1String(RESIZE));
+    arguments.append(QString::number(sizeMb));
+
+    QTime timer;
+    timer.start();
+
+    VBoxManageProcess process;
+
+    bool running = process.runSynchronously(arguments);
+    bool successfull = process.exitCode() == 0;
```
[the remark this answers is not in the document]
etaishev: Add this checking for other. I think this need, because the command, for example "modifyvm", can succeed finish but with error code.

src/plugins/mer/meroptionswidget.cpp
```diff
@@ -159,6 +165,46 @@ void MerOptionsWidget::store()
                 ok = false;
             }
         }
+        if (m_memorySize.contains(sdk)) {
+            if (m_memorySize[sdk] != sdk->memorySize()) {
```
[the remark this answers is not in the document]
etaishev: You mean that the settings are stored in xml and don't updated from VirtualBox?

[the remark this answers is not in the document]
martyone: Yes, QtC works with cached values. Users are not expected to fiddle with the VM settings using VBox interfaces. Of course users are naughty at times and for that reason all values are updated unconditionally from qtc side to allow restoring consistency somehow conveniently in case of troubles.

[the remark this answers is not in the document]
martyone: What do you mean by every time? After sdk upgrade? That should be handled by the installer. This is also something we will have to have a closer look at later, what happens with these values during upgrade. More fun waiting for us yet :)

etaishev:
Fix comments, merge master

src/plugins/mer/mer.pro
```diff
@@ -28,7 +28,7 @@ FORMS += \
     merrpmvalidationstepconfigwidget.ui \
     merrunconfigurationaspectqmllivedetailswidget.ui \
     mersdkdetailswidget.ui \
-    mersdkselectiondialog.ui \
+    mersdkselectiondialog.ui
```
martyone: Please preserve the trailing backslash - it allows to append items without the need to touch the last item. More occurrences below.

src/plugins/mer/meremulatordevice.cpp
```diff
@@ -692,6 +765,23 @@ bool MerEmulatorDeviceManager::restorePorts(const MerEmulatorDevice::Ptr &device
     return true;
 }
 
+bool MerEmulatorDeviceManager::restoreSystemSettings(const MerEmulatorDevice::Ptr &device)
```
martyone: Unless I missed something, this should be merged with restorePorts(), using restoreSystemSettings as the new name. All comments should be updated accordingly (there is also a comment where the function is called)

src/plugins/mer/meremulatordevice.cpp
```diff
@@ -751,6 +856,24 @@ void MerEmulatorDeviceManager::onDeviceListReplaced()
             MerVirtualBoxManager::updateEmulatorQmlLivePorts(merEmulator->virtualMachine(),
                     merEmulator->qmlLivePortsList());
         m_deviceQmlLivePortsCache.insert(merEmulator->id(), nowQmlLivePorts);
+
+        const int nowMemorySize = merEmulator->memorySize();
+        if (nowMemorySize != oldMemorySizeCache.value(merEmulator->id())) {
+            MerVirtualBoxManager::setMemorySize(merEmulator->virtualMachine(), nowMemorySize);
+        }
```
martyone: Curly braces should not be used for one-line blocks. More occurrences below.

src/plugins/mer/meremulatordevice.h
```diff
@@ -101,6 +101,18 @@ class MerEmulatorDevice : public MerDevice
     void setSubnet(const QString& subnet);
     QString subnet() const;
 
+    void setMemorySize(int sizeMb);
+    int memorySize() const;
+
+    void setVdiInfo(const QString &vdiPath, int vdiCapacity, int vdiSizeOnDisk);
```
martyone: It would be nice to use the Mb/Gb suffix as you did for setMemorySize. Ideally member variable, access function and/or it's parameter name would use the same words. `setMeMorySizeMb(memorySizeMb) { m_memorySizeMb = memorySizeMb; }`. I accept that it does not look nice for very long names but this is not the case and it makes code searching/manipulation easier.

src/plugins/mer/mersdkdetailswidget.cpp
```diff
@@ -79,14 +79,21 @@ MerSdkDetailsWidget::MerSdkDetailsWidget(QWidget *parent)
...
+    });
     connect(m_ui->wwwProxyDisabledButton, &QRadioButton::toggled,
             this, &MerSdkDetailsWidget::onWwwProxyDisabledToggled);
     connect(m_ui->wwwProxyAutomaticButton, &QRadioButton::toggled,
             this, &MerSdkDetailsWidget::onWwwProxyAutomaticToggled);
     connect(m_ui->wwwProxyManualButton, &QRadioButton::toggled,
             this, &MerSdkDetailsWidget::onWwwProxyManualToggled);
     connect(m_ui->wwwProxyServersLine, &QLineEdit::textEdited,
-            this, &MerSdkDetailsWidget::onWwwProxyServersEdited);
+                this, &MerSdkDetailsWidget::onWwwProxyServersEdited);
```
martyone: Unwanted space change

src/plugins/mer/mersdkdetailswidget.cpp
```diff
@@ -139,6 +146,26 @@ void MerSdkDetailsWidget::setSdk(const MerSdk *sdk)
     }
 
     m_ui->userNameLabelText->setText(sdk->userName());
+
+    auto handler = [this](int hostMemorySize) {
+        int maxMemorySize = m_ui->memorySpinBox->value() > hostMemorySize ? m_ui->memorySpinBox->value() : hostMemorySize;
```
martyone: There seems to be quite a lot of code duplicated between MerEmulatorDeviceWidget and MerSdkDetailsWidget, and between MerEmulatorDevice and MerSdk at least. We should share the code somehow.

src/plugins/mer/mersdkkitinformation.cpp
```diff
@@ -208,7 +208,7 @@ QWidget *MerSdkKitInformationWidget::buttonWidget() const
 
 void MerSdkKitInformationWidget::handleSdksUpdated()
 {
-    const QList<MerSdk*>& sdks = MerSdkManager::sdks();
+    const QList<MerSdk*> &sdks = MerSdkManager::sdks();
```
martyone: Code polishing should not be mixed with regular code changes. Better leave unchanged, there are more issues in the legacy code.

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -500,6 +515,79 @@ void MerVirtualBoxManager::setVideoMode(const QString &vmName, const QSize &size
...
+
+    qCDebug(Log::vms) << "Resizing VDI took" << timer.elapsed() << "milliseconds";
+
+    return true;
+}
+
+// It is an error to call this function when the VM vmName is running
+bool MerVirtualBoxManager::setMemorySize(const QString &vmName, int sizeMb)
+{
+    qCDebug(Log::vms) << "Changing memory size of " << vmName << "to" << sizeMb << "MB";
```
martyone: Unnecessary space after "of". More occurrences elsewhere.

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -581,7 +700,48 @@ QStringList listedVirtualMachines(const QString &output)
...
+    if (!process.runSynchronously(arguments))
+        return;
+
+    vdiInfoFromOutput(QString::fromLocal8Bit(process.readAllStandardOutput()), virtualMachineInfo);
+}
+
+void vdiInfoFromOutput(const QString &output, VirtualMachineInfo &virtualMachineInfo)
+{
+    //2 vdi size
+    //3 vdi size on disk
```
martyone: It is common to put space after `//` unless it is a code snippet that is commented out

src/plugins/mer/meremulatordevicewidget.cpp
```diff
@@ -242,6 +270,26 @@ void MerEmulatorDeviceWidget::initGui()
 
     connect(MerEmulatorDeviceManager::instance(), &MerEmulatorDeviceManager::storedDevicesChanged,
             this, &MerEmulatorDeviceWidget::onStoredDevicesChanged);
+
+    auto handler = [this](int hostMemorySize) {
+        int maxMemorySize = m_ui->memorySpinBox->value() > hostMemorySize ? m_ui->memorySpinBox->value() : hostMemorySize;
+        m_ui->memorySpinBox->setRange(4, maxMemorySize);
```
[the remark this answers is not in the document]
etaishev: I think no. I did that by default set 3.5Gb max value and update whet fetch host memory size finished. 
It looks better than changing the item's availability a few seconds after opening dialog

src/plugins/mer/mersdkdetailswidget.cpp
etaishev: Move Gui code in own widget. In MerEmulatorDevice and MerSdk now only get and set functions.

src/plugins/mer/mervirtualmachinesettingswidget.ui
```diff
@@ -0,0 +1,146 @@
...
+     <y>60</y>
+     <width>85</width>
+     <height>26</height>
+    </rect>
+   </property>
+   <property name="text">
+    <string>Vdi capacity:</string>
+   </property>
+  </widget>
+  <widget class="QLineEdit" name="vdiPathEdit">
```
martyone: QLabel is used for displaying paths elsewhere.

src/plugins/mer/meremulatordevicewidget.ui
```diff
@@ -7,343 +7,351 @@
...
+         </property>
+         <property name="text">
+          <string/>
+         </property>
+        </widget>
+       </item>
+      </layout>
+     </item>
+     <item row="13" column="0">
+      <layout class="QHBoxLayout" name="virtualMachineInfoLayout"/>
```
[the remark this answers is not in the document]
etaishev: Hmm, ok. You mean like this?

src/plugins/mer/mervirtualmachinesettingswidget.ui
```diff
@@ -0,0 +1,142 @@
...
+   <item row="0" column="1">
+    <widget class="QSpinBox" name="memorySpinBox">
+     <property name="sizePolicy">
+      <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
+       <horstretch>0</horstretch>
+       <verstretch>0</verstretch>
+      </sizepolicy>
+     </property>
+     <property name="suffix">
+      <string> Mb</string>
```
martyone: In messages it should be MB, GB, and similarly VDI. More occurrences elsewhere.

src/plugins/mer/meremulatordevicewidget.ui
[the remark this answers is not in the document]
martyone: Better now.

src/plugins/mer/mervirtualboxmanager.h
```diff
@@ -73,8 +80,18 @@ class MerVirtualBoxManager : public QObject
     static bool updateEmulatorSshPort(const QString &vmName, quint16 port);
     static bool updateEmulatorQmlLivePorts(const QString &vmName, const QList<Utils::Port> &ports);
     static void setVideoMode(const QString &vmName, const QSize &size, int depth);
+    static bool setVdiSizeMb(const QString &path, int sizeMb);
+    static bool setMemorySizeMb(const QString &vmName, int sizeMb);
+    static bool setCpuCount(const QString &vmName, int count);
+
     static QString getExtraData(const QString &vmName, const QString &key);
 
+    static void fetchHostTotalMemorySizeMb();
```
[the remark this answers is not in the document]
etaishev: It was a bug when `MerVirtualMachineSettingsWidget` delete before `fetchHostTotalMemorySizeMb` finished. Fixed, returned back  the solution with lambda. Fixed

etaishev:
In the case of using the sdktool addMerSdk, i think better load VMs params in `MerSdkManager::restoreSdks` than  use `MerVirtualBoxManager` in **sdktool**

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -77,11 +81,22 @@ const char SDK_WWW_NATPF_RULE_NAME[] = "guestwww";
...
+const char COLLECT[] = "collect";
+const char TOTAL_RAM[] = "RAM/Usage/Total";
 
 namespace Mer {
 namespace Internal {
 
-static VirtualMachineInfo virtualMachineInfoFromOutput(const QString &output);
+static VirtualMachineInfo virtualMachineInfoFromOutput(const QString &output, bool fillVdiInfo);
+static void vdiInfoFromOutput(const QString &output, VirtualMachineInfo &virtualMachineInfo);
+static void fetchVdiInfo(VirtualMachineInfo &virtualMachineInfo);
```
martyone: virtualMachineInfo should be passed by pointer. Please also swap the two declarations: fetchVdiInfo first, vdiInfoFromOutput second.

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -581,7 +703,48 @@ QStringList listedVirtualMachines(const QString &output)
     return vms;
 }
 
-VirtualMachineInfo virtualMachineInfoFromOutput(const QString &output)
+void fetchVdiInfo(VirtualMachineInfo &virtualMachineInfo)
```
martyone: Please move fetchVdiInfo and vdiInfoFromOutput after virtualMachineInfoFromOutput.

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -581,7 +703,48 @@ QStringList listedVirtualMachines(const QString &output)
...
+    if (!process.runSynchronously(arguments))
+        return;
+
+    vdiInfoFromOutput(QString::fromLocal8Bit(process.readAllStandardOutput()), virtualMachineInfo);
+}
+
+void vdiInfoFromOutput(const QString &output, VirtualMachineInfo &virtualMachineInfo)
+{
+    // 2 vdi size
+    // 3 vdi size on disk
```
martyone: 2, 3 -> 1, 2?

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -590,11 +753,18 @@ VirtualMachineInfo virtualMachineInfoFromOutput(const QString &output)
...
+    // 14 SATA paths
+    // 15 Memory size
+    // 16 Count of CPU
     QRegExp rexp(QLatin1String("(?:Forwarding\\(\\d+\\)=\"(\\w+),(\\w+),(.*),(\\d+),(.*),(\\d+)\")"
                                "|(?:SharedFolderNameMachineMapping\\d+=\"(\\w+)\"\\W*"
                                "SharedFolderPathMachineMapping\\d+=\"(.*)\")"
                                "|(?:macaddress\\d+=\"(.*)\")"
-                               "|(?:Session(Type|Name)=\"(.*)\")"));
+                               "|(?:Session(Type|Name)=\"(.*)\")"
+                               "|(\"SATA(-\\d-\\d)\"=\"(.*)\")"
```
martyone: Why is this line not using "?:" as the others do

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -581,7 +703,48 @@ QStringList listedVirtualMachines(const QString &output)
...
+        return;
+
+    vdiInfoFromOutput(QString::fromLocal8Bit(process.readAllStandardOutput()), virtualMachineInfo);
+}
+
+void vdiInfoFromOutput(const QString &output, VirtualMachineInfo &virtualMachineInfo)
+{
+    // 2 vdi size
+    // 3 vdi size on disk
+    QRegExp rexp(QLatin1String("(Capacity:\\s+(\\d+)\\s+\\w+)"
```
martyone: Why is this line not using "?:" as the others do

src/plugins/mer/mervirtualboxmanager.cpp
martyone: I think you should use "?:" in the first parentheses to fix this

src/plugins/mer/meremulatordevice.cpp
```diff
@@ -481,7 +547,17 @@ void MerEmulatorDevice::updateAvailableDeviceModels()
...
+        if (!(deviceModelData.contains(MEMORY_SIZE_MB) || deviceModelData.contains(CPU_COUNT) ||
+              deviceModelData.contains(VDI_PATH) || deviceModelData.contains(VDI_SIZE_MB) || deviceModelData.contains(VDI_CAPACITY_MB))) {
+            QString vmName = deviceModelData.value(QLatin1String(VM_NAME)).toString();
+            VirtualMachineInfo info = MerVirtualBoxManager::fetchVirtualMachineInfo(vmName, true);
+            deviceModelData.insert(QLatin1String(MEMORY_SIZE_MB), info.memorySizeMb);
+            deviceModelData.insert(QLatin1String(CPU_COUNT), info.cpuCount);
+            deviceModelData.insert(QLatin1String(VDI_PATH), info.vdiPath);
+            deviceModelData.insert(QLatin1String(VDI_CAPACITY_MB), info.vdiCapacity);
+            deviceModelData.insert(QLatin1String(VDI_SIZE_MB), info.vdiSizeOnDisk);
+        }
```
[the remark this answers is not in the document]
etaishev: Add params memory size, CPU count and vdi capacity to command line. Vdi path and vdi size on disk still need from `vboxmanage`. 

Vdi capacity can't be less than real vdi capacity, need check command line argument for this?

src/tools/sdktool/addmersdkoperation.cpp
```diff
@@ -300,6 +346,16 @@ QVariantMap AddMerSdkOperation::initializeSdks(int version)
     return map;
 }
 
+void addVirtualMachineParameters(Operation::KeyValuePairList *data, const QString &sdk, const QString &virtualMachineName)
+{
+    if (!virtualMachineName.isEmpty()) {
+        Mer::Internal::MerVirtualBoxManager virtualBoxManager;
+        const auto info = virtualBoxManager.fetchVirtualMachineInfo(virtualMachineName, true);
+        *data << Operation::KeyValuePair(QStringList() << sdk << QLatin1String(Mer::Constants::VDI_SIZE_MB), QVariant(info.vdiSizeOnDisk));
+        *data << Operation::KeyValuePair(QStringList() << sdk << QLatin1String(Mer::Constants::VDI_PATH), QVariant(info.vdiPath));
```
martyone: Similarly to addDeviceOperation, these should be expected on command line

src/tools/sdktool/sdktool.pro
```diff
@@ -3,6 +3,7 @@ include(../../qtcreatortool.pri)
 QT -= concurrent gui widgets testlib
 
 QT += xmlpatterns
+QT += core
```
martyone: Really needed?

src/tools/sdktool/sdktool.pro
```diff
@@ -70,12 +74,16 @@ HEADERS += \
     rmmertargetoperation.h \
     addmersdkoperation.h \
     ../../plugins/mer/mertargetsxmlparser.h \
+    ../../plugins/mer/mervirtualboxmanager.h \
+    ../../plugins/mer/merlogging.h \
+    ../../libs/utils/port.cpp \
```
martyone: port.cpp seems incorrect here.

src/plugins/mer/mervirtualboxmanager.h
```diff
@@ -52,6 +54,11 @@ class VirtualMachineInfo
     QList<quint16> qmlLivePorts;
     QStringList macs;
     bool headless;
+    int memorySizeMb;
+    int cpuCount;
+    QString vdiPath;
+    int vdiCapacity;
+    int vdiSizeOnDisk;
```
martyone: Unless I missed something we only query vdiSizeOnDisk once, then store it in QtC settings and never update. That's not right because it changes continuously as the VM is used. I would prefer removing that instead of fixing - it is not important to our use case anyway. The only moment the user would care about size-on-disk is when the disk gets full, but then there is more to check than our two VMs and more suitable tools are likely to be used. Note that this value generally does not correspond to the remaining disk space inside the VM - if this was your motivation for adding this we would need to ssh into the device and use `du` instead.

src/tools/sdktool/adddeviceoperation.cpp
```diff
@@ -414,6 +422,23 @@ bool AddDeviceOperation::exists(const QVariantMap &map, const QString &id)
...
+        if (value.key.contains(QString::fromStdString(Mer::Constants::MER_DEVICE_VIRTUAL_MACHINE))){
+            virtualMachineName = value.value.toString();
+            break;
+        }
+    }
+    if (!virtualMachineName.isEmpty()) {
+        Mer::Internal::MerVirtualBoxManager virtualBoxManager;
+        const auto vmInfo = virtualBoxManager.fetchVirtualMachineInfo(virtualMachineName, true);
+        dev->append(Operation::KeyValuePair(QLatin1String(Mer::Constants::VDI_SIZE_MB), QVariant(vmInfo.vdiSizeOnDisk)));
+        dev->append(Operation::KeyValuePair(QLatin1String(Mer::Constants::VDI_PATH), QVariant(vmInfo.vdiPath)));
```
martyone: I think these should be passed on command line as well. Do not forget the sdktool is meant to be used by installer, not by human user.

src/plugins/mer/meremulatordevice.cpp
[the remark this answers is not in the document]
martyone: No need to check it. Also I realized we should omit size-on-disk from the VM info we handle - see the other comments.

src/plugins/mer/merconstants.h
```diff
@@ -170,6 +175,9 @@ const char MER_SSH_PORT[] = "MER_SSH_PORT";
 const char MER_SSH_ENGINE_NAME[] = "MER_SSH_ENGINE_NAME";
 const char SAILFISH_OS_SDK_ENVIRONMENT_FILTER[] = "SAILFISH_OS_SDK_ENVIRONMENT_FILTER";
 const char MER_DEVICE_VIRTUAL_MACHINE[] = "MER_DEVICE_VIRTUAL_MACHINE";
+const char MER_VIRTUAL_MACHINE_MEMORY_SIZE[] = "MER_VIRTUAL_MACHINE_MEMORY_SIZE";
+const char MER_VIRTUAL_MACHINE_CPU_COUNT[] = "MER_VIRTUAL_MACHINE_CPU_COUNT";
+const char MER_VIRTUAL_MACHINE_VDI_SIZE[] = "MER_VIRTUAL_MACHINE_VDI_SIZE";
```
martyone: Unused?

src/plugins/mer/mervirtualmachinesettingswidget.ui
```diff
@@ -0,0 +1,133 @@
...
+       <horstretch>0</horstretch>
+       <verstretch>0</verstretch>
+      </sizepolicy>
+     </property>
+    </widget>
+   </item>
+   <item row="2" column="0">
+    <widget class="QLabel" name="vdiCapacityLabel">
+     <property name="text">
+      <string>Vdi capacity:</string>
```
martyone: "Vdi" -> "VDI"

src/plugins/mer/mervirtualmachinesettingswidget.ui
```diff
@@ -0,0 +1,133 @@
...
+     </property>
+     <property name="buddy">
+      <cstring>vdiCapacityGbSpinBox</cstring>
+     </property>
+    </widget>
+   </item>
+   <item row="2" column="1">
+    <widget class="QDoubleSpinBox" name="vdiCapacityGbSpinBox">
```
martyone: I would add a tool tip to this widget: "The size of the virtual disk image - it is only possible to increase the value here!"

src/plugins/mer/mervirtualboxmanager.h
```diff
@@ -73,8 +79,16 @@ class MerVirtualBoxManager : public QObject
     static bool updateEmulatorSshPort(const QString &vmName, quint16 port);
     static bool updateEmulatorQmlLivePorts(const QString &vmName, const QList<Utils::Port> &ports);
     static void setVideoMode(const QString &vmName, const QSize &size, int depth);
+    static bool setVdiSizeMb(const QString &path, int sizeMb);
```
martyone: VDI "size" and "capacity" terms are used inconsistently on a number of places.

src/plugins/mer/meremulatordevice.cpp
```diff
@@ -314,6 +325,10 @@ QVariantMap MerEmulatorDevice::toMap() const
     map.insert(QLatin1String(Constants::MER_DEVICE_DEVICE_MODEL), m_deviceModel);
     map.insert(QLatin1String(Constants::MER_DEVICE_ORIENTATION), m_orientation);
     map.insert(QLatin1String(Constants::MER_DEVICE_VIEW_SCALED), m_viewScaled);
+    map.insert(QLatin1String(MEMORY_SIZE_MB), memorySizeMb());
+    map.insert(QLatin1String(CPU_COUNT), cpuCount());
+    map.insert(QLatin1String(VDI_PATH), vdiPath());
+    map.insert(QLatin1String(VDI_CAPACITY_MB), vdiCapacityMb());
```
martyone: Any reason not to access the member variables directly as the above code does?

src/plugins/mer/mervirtualboxmanager.h
```diff
@@ -52,6 +54,10 @@ class VirtualMachineInfo
     QList<quint16> qmlLivePorts;
     QStringList macs;
     bool headless;
+    int memorySizeMb;
+    int cpuCount;
+    QString vdiPath;
+    int vdiCapacity;
```
martyone: The unit suffix is missing here. "Mb"?

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -634,12 +762,76 @@ VirtualMachineInfo virtualMachineInfoFromOutput(const QString &output)
...
+    if (!process.runSynchronously(arguments))
+        return;
+
+    vdiInfoFromOutput(QString::fromLocal8Bit(process.readAllStandardOutput()), virtualMachineInfo);
+}
+
+void vdiInfoFromOutput(const QString &output, VirtualMachineInfo *virtualMachineInfo)
+{
+    // 1 vdi size
+    QRegExp rexp(QLatin1String("(?:Capacity:\\s+(\\d+)\\s+\\w+)"));
```
martyone: Instead of "\\w+" we should clearly state we expect "MBytes" here, similarly to what you do in ramSizeFromOutput

etaishev:
> 1. After increasing VDI capacity, the minimum value of the spin box was not updated.
> 2. Now that I tried the UI I realized displaying VDI path in the UI is not adding any value for the user. I would remove that. While doing that I would also remove VDI path handling from most of the code, changing `bool MerVirtualBoxManager::setVdiSizeMb(const QString &path, int sizeMb)` so that it accepts `vmName` instead of `path` and works on the single expected VDI of the given VM.
> 3. There should be the info icon with tool tip "Stop the virtual machine to unlock this field for editing" next to each of the VM param widgets. Similar to how it is done for SSH/WWW port spin boxes.
> 4. We still miss the part that would enlarge the file system inside the VM when the VDI is enlarged.

4. Yes. It's looks problems to change file system inside, there is no LVM

martyone:
>     4. Yes. It's looks problems to change file system inside, there is no LVM

You can use the following shell code to grow the partition

```
sfdisk --force -N 1 /dev/sda <<<", +"
newsize=$(sfdisk --dump /dev/sda \
	|sed -n 's,^/dev/sda1 : .*size=[[:space:]]*\([0-9]\+\).*,\1,p')
resizepart /dev/sda 1 "$newsize"
resize2fs /dev/sda1
```

The code may grow a bit, so I would store it in a separate file (like it is done for https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/plugins/mer/reset-ambience.sh).

Check this for an example how to execute commands on build engine https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/tools/merssh/enginectlcommand.cpp#L204

For some reason for me the VM got paused during the resize2fs command when I tested it - I had to resume it manually.

etaishev:
Need show some message while resizing? It take about 10-15 second

martyone:
> Need show some message while resizing? It take about 10-15 second

Yup, there should be a message. You can use the code of MerEmulatorDevice::doFactoryReset() as a template. The function is not in master yet, added here https://github.com/sailfishos/sailfish-qtcreator/pull/168/commits/958f57c9a24d1b66d7bab53b63b56ceb35f0110d#diff-cd920f2cd1271468760ae68a9d94b449R554

etaishev:
Add show message while all settings VM apllying

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -500,6 +518,91 @@ void MerVirtualBoxManager::setVideoMode(const QString &vmName, const QSize &size
...
+    arguments.append(QLatin1String(MODIFYMEDIUM));
+    arguments.append(virtualMachineInfo.vdiPath);
+    arguments.append(QLatin1String(RESIZE));
+    arguments.append(QString::number(sizeMb));
+
+    QTime timer;
+    timer.start();
+
+    VBoxManageProcess process;
+    if (!process.runSynchronously(arguments)) {
```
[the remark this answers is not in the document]
etaishev: Yes, however, the virtual machine restarts during this function. It may be don't obvious for user why change VM states. Do it anyway?

src/plugins/mer/meremulatordevice.cpp
```diff
@@ -722,23 +784,41 @@ void MerEmulatorDeviceManager::onDeviceRemoved(Core::Id id)
...
 }
 
 void MerEmulatorDeviceManager::onDeviceListReplaced()
 {
+    QProgressDialog progress(tr("Applying virtual machine settings"), tr("Abort"), 0, 0, ICore::dialogParent());
+    progress.setWindowModality(Qt::WindowModal);
+    QPushButton cancelButton(tr("Abort"));
+    cancelButton.setDisabled(true);
+    progress.setCancelButton(&cancelButton);
+    progress.show();
```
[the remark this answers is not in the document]
etaishev: I don't set QProgressDialog::minimumDuration because we can't correctly cancel resize vdi operation

src/plugins/mer/mervirtualmachinesettingswidget.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+void MerVirtualMachineSettingsWidget::setCpuCount(int count)
+{
+    ui->cpuCountSpinBox->setValue(count);
+}
+
+void MerVirtualMachineSettingsWidget::setVdiCapacityMb(int vdiCapacityMb)
+{
+    // the minimum size is the current size because VBoxManager can't reduce storage size
+    ui->vdiCapacityGbSpinBox->setRange(vdiCapacityMb / 1024.0, DEFAULT_MAX_VDI_SIZE_GB);
```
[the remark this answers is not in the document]
etaishev: Done.

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -500,6 +518,91 @@ void MerVirtualBoxManager::setVideoMode(const QString &vmName, const QSize &size
...
 }
 
+bool MerVirtualBoxManager::setVdiCapacityMb(MerConnection *connection, int sizeMb)
+{
+    const QString &vmName = connection->virtualMachine();
+    qCDebug(Log::vms) << "VM:" << vmName << "Changing vdi size to" << sizeMb << "MB";
+
+    const VirtualMachineInfo virtualMachineInfo = fetchVirtualMachineInfo(vmName, true);
+    if (sizeMb < virtualMachineInfo.vdiCapacityMb)
+        return false;
```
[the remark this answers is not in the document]
etaishev: Add warning

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -500,6 +518,91 @@ void MerVirtualBoxManager::setVideoMode(const QString &vmName, const QSize &size
...
+    timer.start();
+
+    VBoxManageProcess process;
+    if (!process.runSynchronously(arguments)) {
+        qWarning() << "VBoxManage failed to" << MODIFYMEDIUM << virtualMachineInfo.vdiPath << RESIZE << sizeMb
+                   << "for VM" << vmName;
+        return false;
+    }
+
+    if (!increaseVmPatrition(connection)) {
```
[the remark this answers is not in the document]
etaishev: May be replace script like this to avoid restarting the device.
 And set autostart service on system start
```
#!/bin/bash

set -o nounset

resize() {
	# Compare available and current partition sizes
	pre=$(sfdisk -s /dev/sda1)
	post=$(sfdisk -s /dev/sda)

    if ! diff -q $pre $post >/dev/null 2>&1; then
    	echo "RESIZE:"
	    sfdisk --force -N 1 --no-reread /dev/sda <<<", +" || return
	    newsize=$(sfdisk --dump /dev/sda |sed -n 's,^/dev/sda1 : .*size=[[:space:]]*\([0-9]\+\).*,\1,p')
	    resizepart /dev/sda 1 "$newsize" || return
	    resize2fs /dev/sda1
    fi
}

resize || exit
```

src/plugins/mer/mervirtualboxmanager.cpp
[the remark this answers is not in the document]
martyone: Asynchronous so that it does not block the event loop. The VM will be locked down using MerConnection::lockDown, so starting from qtc will not be possible. Also the progress dialog will be modal, so the user will not be able to interact with qtc at all. It is just about to not block repainting the UI.

src/plugins/mer/meremulatordevice.cpp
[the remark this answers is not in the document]
martyone: Unless I missed something there is no conflict between setting minimumDuration and inability to cancel. The minimumDuration serves just as a mean to delay displaying the progress dialog. It is set by default - to 4000ms. I would prefer setting 2000ms instead, because we want to show the message that the operation was successfully done at the end, and 4s feels quite a long period of seamless inactivity.

src/plugins/mer/mervirtualboxmanager.cpp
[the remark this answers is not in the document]
martyone: yup, that's my plan

src/plugins/mer/mervirtualmachinesettingswidget.cpp
```diff
@@ -0,0 +1,108 @@
...
+    ui->cpuCountSpinBox->setValue(count);
+}
+
+void MerVirtualMachineSettingsWidget::setVdiCapacityMb(int vdiCapacityMb)
+{
+    const int vdiCapacityGb = static_cast<int>(vdiCapacityMb / 1024.0);
+    // Prohibit adding more than 10GB in one step, because there is no way to go back
+    int maxVdiSize = vdiCapacityGb + 10 > DEFAULT_MAX_VDI_SIZE_GB ? DEFAULT_MAX_VDI_SIZE_GB : vdiCapacityGb + 10;
+    // the minimum size is the current size because VBoxManager can't reduce storage size
+    ui->vdiCapacityGbSpinBox->setRange(vdiCapacityGb, maxVdiSize);
```
martyone: Wouldn't this be enough?

```
ui->vdiCapacityGbSpinBox->setRange(vdiCapacityGb, vdiCapacityGb + MAX_VDI_SIZE_INCREMENT_GB);
```

IIUC one would have to increase the size approximately 200 times to get over DEFAULT_MAX_VDI_SIZE_GB the regular way :)

src/plugins/mer/mervirtualboxmanager.cpp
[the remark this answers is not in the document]
etaishev: Now resize vdi takes time as changing memory and cpu. It take a lot time when restart VM and change partition table inside, but now it was on start VM.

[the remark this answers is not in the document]
martyone: OK, hopefully it will be always that fast, even when the VDI is bigger and/or the VM is used for longer time.

src/plugins/mer/meremulatordevice.cpp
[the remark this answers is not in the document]
etaishev: Hmm. In this function no asynchronous function and QProgressDialog::minimumDuration don't open dialog automatically, only called explicitly work...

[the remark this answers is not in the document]
martyone: Yeah, that's the problem with the blocking call - it newer returns to the event loop and freezes the thread. So if you are sure it will never take longer than a second or two to resize the VDI, just omit the progress dialog here and in MerOptionsWidget. Otherwise it would have to be reimplemented in a non-blocking way.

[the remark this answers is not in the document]
etaishev: Add asynchronous to MerVirtualBoxManager::setVdiCapacityMb

etaishev:
Resolve conflicts with master

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -500,6 +518,94 @@ void MerVirtualBoxManager::setVideoMode(const QString &vmName, const QSize &size
...
 
+void MerVirtualBoxManager::setVdiCapacityMb(const QString &vmName, int sizeMb, QObject *context, std::function<void(bool)> slot)
+{
+    qCDebug(Log::vms) << "Changing vdi size of" << vmName << "to" << sizeMb << "MB";
+
+    const VirtualMachineInfo virtualMachineInfo = fetchVirtualMachineInfo(vmName, true);
+    if (sizeMb < virtualMachineInfo.vdiCapacityMb) {
+        qWarning() << "VBoxManage failed to" << MODIFYMEDIUM << virtualMachineInfo.vdiPath << RESIZE << sizeMb
+                   << "for VM" << vmName << ". Can't reduce VDI. Current size:" << virtualMachineInfo.vdiCapacityMb;
+        slot(false);
```
martyone: Use `QTimer::singleShot(0, context, [slot]() { slot(false); })` here instead, and you can get rid of the `needLoop` hack where this is used.

src/plugins/mer/meremulatordevice.cpp
```diff
@@ -737,6 +819,37 @@ void MerEmulatorDeviceManager::onDeviceListReplaced()
...
+            MerVirtualBoxManager::setVdiCapacityMb(merEmulator->virtualMachine(), nowVdiCapacityMb, &loop, [&loop, &needLoop] (bool ok) {
+                Q_UNUSED(ok);
+                if (!loop.isRunning())
+                    needLoop = false;
+                else
+                    loop.exit();
+            });
+
+            if (needLoop) {
+                loop.exec();
```
martyone: Handle failure?

src/plugins/mer/mervirtualmachinesettingswidget.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+void MerVirtualMachineSettingsWidget::setCpuCount(int count)
+{
+    ui->cpuCountSpinBox->setValue(count);
+}
+
+void MerVirtualMachineSettingsWidget::setVdiCapacityMb(int vdiCapacityMb)
+{
+    const double vdiCapacityGb = vdiCapacityMb / 1024.0;
+    // Prohibit adding more than 10GB in one step, because there is no way to go back
```
martyone: Hard-coded number in a comment :)

src/plugins/mer/mervirtualmachinesettingswidget.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // the minimum size is the current size because VBoxManager can't reduce storage size
+    ui->vdiCapacityGbSpinBox->setRange(vdiCapacityGb, vdiCapacityGb + MAX_VDI_SIZE_INCREMENT_GB);
```
martyone: Doesn't the compiler issue a warning about narrowing here?

src/plugins/mer/mervirtualmachinesettingswidget.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    const int maxCpuCount = MerVirtualBoxManager::getHostTotalCpuCount();
+    ui->cpuCountSpinBox->setRange(1, maxCpuCount);
+
+    connect(ui->memorySpinBox, SIGNAL(valueChanged(int)),
+            this, SIGNAL(memorySizeChanged(int)));
+    connect(ui->cpuCountSpinBox, SIGNAL(valueChanged(int)),
+            this, SIGNAL(cpuCountChanged(int)));
+    connect(ui->vdiCapacityGbSpinBox, QOverload<double>::of(&QDoubleSpinBox::valueChanged), [this](double value) {
+        emit vdiCapacityChnaged(static_cast<int>(value * 1024));
```
martyone: This seems error prone. Imagine when `setVdiCapacityMb(1024+512)` is first called, then the Gb spin box is incremented to 2 and returned back to 1 - nothing changed from user point of view, but this will emit `vdiCapacityChanged(1024)`.

etaishev: Hm, i can't repeat error. 
In first call of `setVdiCapacityMb(1024+512)`  will be set range `ui->vdiCapacityGbSpinBox->setRange(1.5, 11.5)`. With this range signal `QDoubleSpinBox::valueChanged` doesn't call  with 1

src/plugins/mer/mervirtualmachinesettingswidget.cpp
etaishev: No, `setRange` gets double values

src/plugins/mer/meremulatordevice.cpp
etaishev: When apply emulator settings the error is not processed anywhere. Add it?

src/plugins/mer/mervirtualmachinesettingswidget.cpp
martyone: ah, I missed it's a QDoubleSpinBox

src/plugins/mer/mervirtualmachinesettingswidget.cpp
martyone: ah, I missed it's a QDoubleSpinBox

src/plugins/mer/meremulatordevice.cpp
martyone: Now I noticed it is not handled in other cases either, so let's forget about.

src/plugins/mer/mervirtualboxmanager.h
```diff
@@ -37,10 +37,12 @@ namespace Utils {
 namespace Mer {
 namespace Internal {
 
+class MerConnection;
```
martyone: Remnant?

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -24,10 +24,13 @@
 
 #include "merconstants.h"
 #include "merlogging.h"
+#include "merconnection.h"
+#include "tools/merssh/merremoteprocess.h"
```
martyone: Remnant?

src/plugins/mer/mer.pro
```diff
@@ -158,7 +161,9 @@ HEADERS += \
     merdevicemodelcombobox.h \
     meremulatormodeoptionspage.h \
     meremulatormodeoptionswidget.h \
-    meremulatormodedetailswidget.h
+    meremulatormodedetailswidget.h \
+    mervirtualmachinesettingswidget.h \
+    ../../tools/merssh/merremoteprocess.h \
```
martyone: Remnant?

src/plugins/mer/mervirtualboxmanager.h
etaishev: Yes, removed

etaishev:
Split commits. Move commit `MerVirtualBoxManager: Add check that process exit normaly` into second place.

src/plugins/mer/merdevicefactory.cpp
```diff
@@ -31,6 +31,7 @@
 #include "mersdkmanager.h"
 #include "mersshkeydeploymentdialog.h"
 #include "mersettings.h"
+#include "mervirtualboxmanager.h"
```
martyone: Remnant?

src/plugins/mer/meremulatordevice.cpp
```diff
@@ -737,6 +819,29 @@ void MerEmulatorDeviceManager::onDeviceListReplaced()
...
+        if (nowCpuCount != oldCpuCountCache.value(merEmulator->id()))
+            MerVirtualBoxManager::setCpuCount(merEmulator->virtualMachine(), nowCpuCount);
+
+        m_deviceCpuCountCache.insert(merEmulator->id(), nowCpuCount);
+
+        const int nowVdiCapacityMb = merEmulator->vdiCapacityMb();
+        if (nowVdiCapacityMb != oldVdiInfoCache.value(merEmulator->id())) {
+            QEventLoop loop;
+            MerVirtualBoxManager::setVdiCapacityMb(merEmulator->virtualMachine(), nowVdiCapacityMb, &loop, [&loop] (bool ok) {
+                Q_UNUSED(ok);
```
martyone: Now I noticed for build engine we issue the "Some changes could not be saved!" message in case of errors (`MerOptionsWidget::store()`). We should add the same error handling here. Please use a separate commit for extending the original code, ordered before this commit.

src/tools/sdktool/adddeviceoperation.cpp
```diff
@@ -443,8 +450,6 @@ Operation::KeyValuePairList AddDeviceOperation::createDevice(const QString &id,
     dev.append(KeyValuePair(QLatin1String("Timeout"), QVariant(timeout)));
     dev.append(KeyValuePair(QLatin1String("Uname"), QVariant(uname)));
     dev.append(KeyValuePair(QLatin1String("Version"), QVariant(version)));
-
     dev.append(extra);
-
```
martyone: I would preserve the spacing

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -634,13 +771,78 @@ VirtualMachineInfo virtualMachineInfoFromOutput(const QString &output)
...
+        }
+    }
+    return 0;
+}
+
 } // Internal
 } // VirtualBox
 
 #include "mervirtualboxmanager.moc"
+
```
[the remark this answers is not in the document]
etaishev: Removed

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -39,6 +39,10 @@
 #include <QSize>
 #include <QTime>
 #include <QTimerEvent>
+#include <QThread>
+#include <QStorageInfo>
```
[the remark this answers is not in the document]
etaishev: Removed

src/plugins/mer/mervirtualmachinesettingswidget.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+MerVirtualMachineSettingsWidget::~MerVirtualMachineSettingsWidget()
+{
+    delete ui;
+}
+
+void MerVirtualMachineSettingsWidget::setMemorySizeMb(int sizeMb)
+{
+    if (ui->memorySpinBox->maximum() < sizeMb)
+        ui->memorySpinBox->setRange(4, sizeMb);
```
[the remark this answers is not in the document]
etaishev: I set minimum memory size like in VirtualBox gui

src/plugins/mer/mervirtualmachinesettingswidget.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+void MerVirtualMachineSettingsWidget::initGui()
+{
+    auto handler = [this](int sizeMb) {
+        int maxMemorySizeMb = ui->memorySpinBox->value() > sizeMb ? ui->memorySpinBox->value() : sizeMb;
+        ui->memorySpinBox->setRange(4, maxMemorySizeMb);
+    };
+
+    ui->memorySpinBox->setRange(4, DEFAULT_MAX_MEMORY_SIZE_MB);
+    MerVirtualBoxManager::getHostTotalMemorySizeMb(this, handler);
```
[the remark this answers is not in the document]
etaishev: Fixed

src/plugins/mer/mervirtualmachinesettingswidget.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    const int maxCpuCount = MerVirtualBoxManager::getHostTotalCpuCount();
+    ui->cpuCountSpinBox->setRange(1, maxCpuCount);
+
+    connect(ui->memorySpinBox, SIGNAL(valueChanged(int)),
+            this, SIGNAL(memorySizeChanged(int)));
+    connect(ui->cpuCountSpinBox, SIGNAL(valueChanged(int)),
+            this, SIGNAL(cpuCountChanged(int)));
```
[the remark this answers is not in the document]
etaishev: Fixed

src/plugins/mer/meroptionswidget.h
```diff
@@ -92,6 +95,9 @@ private slots:
     QMap<MerSdk*, QString> m_wwwProxy;
     QMap<MerSdk*, QString> m_wwwProxyServers;
     QMap<MerSdk*, QString> m_wwwProxyExcludes;
+    QMap<MerSdk*, int> m_vdiCapacity;
+    QMap<MerSdk*, int> m_memorySize;
```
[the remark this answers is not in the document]
etaishev: Fixed. Add Mb suffix in other palaces

src/plugins/mer/meremulatordevice.cpp
etaishev: Add handling error

src/plugins/mer/meremulatordevice.cpp
```diff
@@ -807,43 +809,76 @@ void MerEmulatorDeviceManager::onDeviceListReplaced()
...
+        Utils::PortList nowQmlLivePorts = merEmulator->qmlLivePorts();
+        if (nowQmlLivePorts.toString() != oldQmlLivePortsCache.value(merEmulator->id()).toString()) {
+            Utils::PortList savedPorts = MerVirtualBoxManager::updateEmulatorQmlLivePorts(merEmulator->virtualMachine(),
                     merEmulator->qmlLivePortsList());
-        m_deviceQmlLivePortsCache.insert(merEmulator->id(), nowQmlLivePorts);
 
-        const int nowMemorySize = merEmulator->memorySizeMb();
-        if (nowMemorySize != oldMemorySizeCache.value(merEmulator->id()))
-            MerVirtualBoxManager::setMemorySizeMb(merEmulator->virtualMachine(), nowMemorySize);
+            if (savedPorts.toString() != nowQmlLivePorts.toString()) {
```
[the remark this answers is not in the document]
etaishev: Add lambda to check PortList. Yes, its work as you say

src/plugins/mer/mervirtualmachinesettingswidget.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "ui_mervirtualmachinesettingswidget.h"
+#include <QFormLayout>
+
+namespace Mer {
+namespace Internal {
+
+namespace {
+const int DEFAULT_MAX_MEMORY_SIZE_MB = 3584;
+const int MAX_VDI_SIZE_INCREMENT_GB = 10;
+const int MIN_VIRTUALBOX_MEMORY_SIZE = 4;
```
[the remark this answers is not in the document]
etaishev: What value should I set? 512 Mb?

src/plugins/mer/meremulatordevice.cpp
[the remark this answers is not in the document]
martyone: OK, not ideal (on failure it will replace ranges with plain enumerations) but let it be like this :) Just please remove the extra braces around one of the `if` blocks

src/plugins/mer/mervirtualmachinesettingswidget.cpp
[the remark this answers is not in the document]
martyone: Installer sets 1024 MB IIRC, 512 might work as well. Could you please verify that?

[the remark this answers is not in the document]
etaishev: Yes, emulator started with 512 and run apps from wizards.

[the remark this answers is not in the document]
martyone: Worth to consider something more challenging?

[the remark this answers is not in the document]
etaishev: Apps from examples working ok, show web page in Web View example. I think we should leave some space for experiment for user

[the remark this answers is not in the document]
martyone: ok :)

martyone:
I think it looks good now, I will test it and integrate later after the upcoming release. Thanks!

martyone:
Rebase onto the latest master please

etaishev:
Rebased

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -77,11 +79,22 @@ const char SDK_WWW_NATPF_RULE_NAME[] = "guestwww";
...
+const char METRICS[] = "metrics";
+const char COLLECT[] = "collect";
+const char TOTAL_RAM[] = "RAM/Usage/Total";
 
 namespace Mer {
 namespace Internal {
 
-static VirtualMachineInfo virtualMachineInfoFromOutput(const QString &output);
+static VirtualMachineInfo virtualMachineInfoFromOutput(const QString &output, bool fillVdiInfo);
+static void fetchVdCapacity(VirtualMachineInfo *virtualMachineInfo);
```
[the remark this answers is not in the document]
etaishev: Yes, renamed

martyone:
Please rebase again, time to merge it :)

etaishev:
Rebased

PR #147, opened 2018-10-26
Mer: Allow to set AllowCreateMissing QmlLive workspace option

martyone:


martyone:
Note that all but the very last commit belong to other PR

PR #148, opened 2018-11-01
IDE-16 Use correct contextual help

etaishev:
Elements with same name (for example Sailfish.Silica.Button and QtQuick.Controls.Button) have a same helpId and
 QtCreator help plugin first of all select Qt link.
This patch resolve the issue by looking for a module name of element and select best, then return url as helpId.

### DIFF
--- a/src/plugins/qmljseditor/qmljshoverhandler.cpp
+++ b/src/plugins/qmljseditor/qmljshoverhandler.cpp
@@ -179,17 +179,49 @@ bool QmlJSHoverHandler::setQmlTypeHelp(const ScopeChain &scopeChain, const Docum
         return false;
     } while (0);
 
-    // Check if the module name contains a major version.
+    // Check if the module name contains a major version and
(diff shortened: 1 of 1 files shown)


martyone:
This touches just upstream code. Could you please get this accepted by upstream first?

etaishev:
Ok. Now has some tasks in upstream which fix similar problems with help, but Qt Gerrit Reviews unavailable. I'll wait for them first.

etaishev:
Well in last QtCreator this bug didn't fixed.  The same as our bug was closed and marked as duplicate([QTCREATORBUG-17657](https://bugreports.qt.io/browse/QTCREATORBUG-17657)). But the [QTCREATORBUG-18527](https://bugreports.qt.io/browse/QTCREATORBUG-18527) isn't same and fix of it don't work for us.

martyone:
Open your own upstream bug then? Submitting your patch for review ideally.

PR #156, opened 2018-12-04
Add support for updating the translations without rebuilding the whole project

vigejolla:


src/plugins/mer/merplugin.cpp
```diff
@@ -168,6 +174,57 @@ bool MerPlugin::initialize(const QStringList &arguments, QString *errorString)
...
+    {
+        Kit* kit = SessionManager::startupProject()->activeTarget()->kit();
+        if (MerSdkManager::isMerKit(kit))
+        {
+            MerSdk* sdk = MerSdkKitInformation::sdk(kit);
+            return sdk->userName();
+        }
+        return QString();
+    });
+
```
martyone: These are already available via "%{CurrentBuild:Env:NAME}", so in the external tool definition one can use e.g. MER_SSH_PORT="%{CurrentBuild:Env:MER_SSH_PORT}" to add the environment variable.

src/plugins/mer/merplugin.cpp
```diff
@@ -168,6 +174,57 @@ bool MerPlugin::initialize(const QStringList &arguments, QString *errorString)
             MerQmlLiveBenchManager::startBench);
     menu->addAction(startQmlLiveBenchCommand);
 
+    Utils::MacroExpander *expander = Utils::globalMacroExpander();
+    expander->registerVariable("mer:merssh", tr("Location of merssh"),
+                               []()
+    {
+        return QString(Core::ICore::libexecPath() + QLatin1String("/merssh") + QStringLiteral(QTC_HOST_EXE_SUFFIX));
+    });
```
martyone: Add MER_WRAPPER_LUPDATE to wrapperScripts in mertarget.cpp, then you can specify the executable as %{CurrentBuild:Env:MER_SSH_SDK_TOOLS}/lupdate

vigejolla: That still feels like just a more complicated way of doing it. Adds complexity without added benefit.

vigejolla: Looking at the scripts (qmake etc), their only function is to add MER_SSH_TARGET_NAME and MER_SSH_SDK_TOOLS to envinroment before calling merssh. Neither of those variables are needed in this case.

martyone: There is a pattern. The pattern is used consistently in 100% cases even if there are cases when the pattern adds no functional value. The value comes from using a pattern consistently.

Moreover if I am not wrong in this particular case the pattern enables adding just one line of code instead of six lines. But even if this was not true, the formerly said applies - there is a pattern, so let's follow it.

src/tools/merssh/lupdatecommand.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+QString LUpdateCommand::name() const
+{
+    return QLatin1String("lupdate");
+}
+
+int LUpdateCommand::execute()
+{
+    QString command = QLatin1String("sb2 ") +
+                      arguments().join(QLatin1Char(' '));
```
martyone: The "-t <target-name>" is missing here.

vigejolla: Adding - and I recognize my defeat on using the script instead of calling merssh directly. After this change the script actually servers a purpose: setting the target environment variable

src/plugins/mer/merplugin.cpp
vigejolla: Now using the script

src/share/qtcreator/externaltools/lupdate.xml
```diff
@@ -31,10 +31,11 @@
     <displayname>Update Translations (lupdate)</displayname>
     <category>Linguist</category>
     <order>1</order>
-    <executable>
-        <path>%{CurrentProject:QT_INSTALL_BINS}/lupdate</path>
+    <executable output="showinpane" error="showinpane" modifiesdocument="no">
```
martyone: Shouldn't modifiesdocument be yes? Isn't showinpane the default for output/error, i.e., can we omit it as in the original code?

vigejolla: showinpane removed, modifiesdocument changed to yes.

src/share/qtcreator/externaltools/lupdate.xml
```diff
@@ -31,10 +31,11 @@
     <displayname>Update Translations (lupdate)</displayname>
     <category>Linguist</category>
     <order>1</order>
-    <executable>
-        <path>%{CurrentProject:QT_INSTALL_BINS}/lupdate</path>
+    <executable modifiesdocument="yes">
```
martyone: It was suspicious to me that in the original code modifiesdocument is not set, so I checked it and it turned out that the name modifiesdocument is not very self-explanatory - what it means is "modifies current document", so we should leave the default "no" here. https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/plugins/coreplugin/externaltool.cpp#L392

vigejolla: It is theoretically possible that it modifies the current document - if one of the language files happen to be opened, which I guess is pretty much never. If you are happy with the default, I can just remove it.

PR #162, opened 2018-12-21
Jb39150

jlehtoranta:
Add support for build engine proxy configuration

### DIFF
--- a/src/plugins/mer/mersdkdetailswidget.cpp
+++ b/src/plugins/mer/mersdkdetailswidget.cpp
@@ -35,6 +35,7 @@
 #include <QMessageBox>
 
 using namespace Utils;
+using namespace Mer::Constants;
 
 namespace Mer {
 namespace Internal {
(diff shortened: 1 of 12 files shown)


src/plugins/mer/mersdkdetailswidget.cpp
```diff
@@ -208,5 +234,142 @@ void MerSdkDetailsWidget::onPathChooserEditingFinished()
...
+
+    if (servers == m_wwwProxyServers) {
+        return;
+    }
+
+    if (m_wwwProxy == MER_SDK_PROXY_AUTOMATIC) {
+        reg.setPattern("^(\\S*)(.*)$");
+        servers.replace(reg, "\\1");
+        reg.setPattern("[,;]");
+        servers.replace(reg, "");
```
martyone: I do understand the regexps, but I do not get the purpose :) Could you please add a comment?

jlehtoranta:
PR cleaned up + decided to keep the hide/show behavior of the proxy options.

src/plugins/mer/mersdkdetailswidget.cpp
```diff
@@ -208,5 +233,131 @@ void MerSdkDetailsWidget::onPathChooserEditingFinished()
...
+    QString proxies = s;
+    QRegularExpression reg;
+
+    int cpos = m_ui->wwwProxyServersLine->cursorPosition();
+    int l = proxies.length();
+
+    reg.setPattern("[,;\\s]");
+    if (m_wwwProxy == MER_SDK_PROXY_AUTOMATIC) {
+        // Allow to input a single item with no spaces, commas or semicolons
+        proxies.replace(reg, "");
```
martyone: Wouldn't this break if the URL contains comma or semicolon? Sometimes these are used in tuple-like parameter values. Maybe cleaning whitespaces would be sufficient.

jlehtoranta: Fixed by strong quoting in wwwproxycommand.cpp. All characters are now allowed for proxy autoconfiguration URL. In manual mode entries are separated by single spaces, so all characters except whitespaces are allowed for each entry.

PR #165, opened 2019-01-09
Mer: Allow to access and update emulator state from other Qt Creator plugins

yarfruct:
We are implementing a Qt Creator plugin for sensors emulation. To do so we need to use the "mer" Qt Creator plugin for:

1. Monitoring the Sailfish OS Emulator virtual machine state (on or off).
2. Setting up port forwarding between the host machine and Sailfish OS Emulator.

When implementing the sensor emulation plugin, we faced the following issues and fixed them in this pull request:

1. The "mer" plugin does not allow to use itself as a dependency.
2. It is not possible to use the "mer" plugin classes and methods inside other Qt Creator plugins.
3. The "mer" plugin have no methods to set up port forwarding for an arbitrary port.

**First issue:** Our plugin cannot be loaded when we describe the "mer" plugin as a dependency.
**Reason of the first issue:** The "mer" plugin `Name` identifier and its shared library name are different. During Qt Creator plugin loading it checks dependencies and compares their plugin `Name` identifiers with shared libraries names. The "mer" plugin have "SailfishOS" as the plugin name and "Mer" as name of the shared library and this is a reason why it cannot be initialized as a dependency of our plugin and why our plugin cannot be loaded.
**Solution of the first issue:** Change the "mer" plugin shared library name from "Mer" to "SailfishOS".

**Second issue:** We have to use the emulator connection for monitoring the emulator VM state. To get instance of the connection it is used the `MerEmulatorDevice::connection()` method. But when we use the "mer" plugin classes inside our plugin code base the build is failed on the linking stage.
**Reasons of the second issue:** 
1. The "mer" plugin classes are inside the `Mer::Internal` namespace.
2. The "mer" plugin classes does not have the `Q_DECL_EXPORT` macros in their declarations.

**Solution of the second issue:** Move the `MerConnection`, `MerEmulatorDevice` and `MerDevice` classes from the `Mer::Internal` namespace and add the `Q_DECL_EXPORT` macros to classes that will be used inside our plugin: `MerConnection` and `MerEmulatorDevice` .

**Third issue:** We have to set up port forwarding between the host machine and Sailfish OS Emulator. The "mer" Qt Creator plugin contains the `MerVirtualBoxManager` class that uses the `VBoxManage` utility for working with virtual machines. But a problem is that this class has no methods to set up port forwarding for an arbitrary port.
**Solution of the third issue:** 
1. Add methods to update and delete a port forwarding for ports to the `MerVirtualBoxManager`.
2. Add a method to set up an arbitrary port to the `MerEmulatorDevice` that is used  `MerVirtualBoxManager`.
3. Add a method to the `MerEmulatorDevice` for checking whether the port forwarding rule is already set or not.


src/plugins/mer/meremulatordevice.h
```diff
@@ -114,6 +114,9 @@ class Q_DECL_EXPORT MerEmulatorDevice : public MerDevice
 
     MerConnection *connection() const;
 
+    void setPortForwardingRule(const QString &ruleName, quint16 hostPort, quint16 emulatorVmPort) const;
+    bool isPortForwardingSet(quint16 hostPort) const;
```
martyone: I would prefer the following:

```
bool addPortForwarding(const QString &ruleName, quint16 hostPort, quint16 emulatorVmPort);
bool removePortForwarding(const QString &ruleName);
bool hasPortForwarding(quint16 hostPort, QString *ruleName = nullptr) const;
```

src/plugins/mer/meremulatordevice.h
```diff
@@ -115,6 +114,10 @@ class MerEmulatorDevice : public MerDevice
 
     MerConnection *connection() const;
 
+    void addPortForwarding(const QString &ruleName, quint16 hostPort, quint16 emulatorVmPort) const;
+    bool removePortForwarding(const QString &ruleName);
+    bool hasPortForwarding(quint16 hostPort) const;
```
martyone: Why did I care to explicitly state how exactly I would prefer it to look like? :) - Please fix according to https://github.com/sailfishos/sailfish-qtcreator/pull/165#discussion_r248649356

yarfruct: Unfortunately, it is impossible to get information about the configured port forwarding rules by their names in the current architecture because the `VirtualMachineInfo` class contains only information about port numbers, not about rules names.

Do you propose to store the port forwarding rules names inside `VirtualMachineInfo` and use this information in the `hasPortForwarding()` method implementation?

martyone: Yes I would do so. The user of the API should be able to verify that a port is forwarded according to their needs - by checking the ruleName.

martyone: The `ruleName` in the API I suggested was an (optional) out-type parameter, so that you can determine the state with single call:

```
QString ruleName;
if (emulator->hasPortForwarding(1234, &ruleName) {
    if (ruleName == "rule1") {
        ... // OK, already set up
    } else {
        ... // Err, conflict
    }
} else {
    ... // Free, set it up
}
```

src/plugins/mer/mervirtualboxmanager.cpp
```diff
@@ -601,15 +650,18 @@ VirtualMachineInfo virtualMachineInfoFromOutput(const QString &output)
     while ((pos = rexp.indexIn(output, pos)) != -1) {
         pos += rexp.matchedLength();
         if (rexp.cap(0).startsWith(QLatin1String("Forwarding"))) {
+            QString ruleName = rexp.cap(1);
```
martyone: I think we should either not introduce `ruleName` or replace all occurences of  `rexp.cap(1)` with `ruleName`

src/plugins/mer/mervirtualboxmanager.h
```diff
@@ -48,8 +48,9 @@ class VirtualMachineInfo
     QString sharedSsh;
     quint16 sshPort;
     quint16 wwwPort;
-    QList<quint16> freePorts;
-    QList<quint16> qmlLivePorts;
+    QHash<QString, quint16> freePorts;
+    QHash<QString, quint16> qmlLivePorts;
+    QHash<QString, quint16> otherPorts;
```
martyone: We should use QMap instead - order matters

src/plugins/mer/meremulatordevice.cpp
```diff
@@ -454,6 +454,41 @@ MerConnection *MerEmulatorDevice::connection() const
...
+}
+
+bool MerEmulatorDevice::hasPortForwarding(quint16 hostPort, QString *ruleName) const
+{
+    QList<QHash<QString, quint16>> rules({MerVirtualBoxManager::fetchVirtualMachineInfo(
+                                                m_connection->virtualMachine()).otherPorts,
+                                            MerVirtualBoxManager::fetchVirtualMachineInfo(
+                                                m_connection->virtualMachine()).qmlLivePorts,
+                                            MerVirtualBoxManager::fetchVirtualMachineInfo(
+                                                m_connection->virtualMachine()).freePorts});
```
martyone: We should fetch just once. I would also move the implementation under MerVirtualBoxManager.

src/plugins/mer/meremulatordevice.cpp
```diff
@@ -454,6 +454,41 @@ MerConnection *MerEmulatorDevice::connection() const
...
+        if (rules[i].values().contains(hostPort)) {
+            if (ruleName)
+                *ruleName = rules[i].key(hostPort);
+            return true;
+        }
+    }
+    return false;
+}
+
+void MerEmulatorDevice::runVirtualMachine() const
```
martyone: We shouldn't add this. device->connection()->connectTo() should be used instead.

yarfruct:
@martyone All comments are resolved and pull request is updated.

martyone:
There is one issue I didn't realize before - there are WIP pull requests #146 and #168 (the latter on top of the former) that refactorizes MerVirtualBoxManager and have more changes than your commit has, so I would prefer you to further rebase you PR on top of #168, i.e., on top of branch jb44217.

Also please check https://chris.beams.io/posts/git-commit/#seven-rules again.

yarfruct:
Sorry, but I can't make a rebase right now because an assembly of the `mer` plugin is failed.

Call `fetchVirtualMachineInfo(vmName, true)` on [this line](https://github.com/sailfishos/sailfish-qtcreator/pull/168/files#diff-44c14b8d61c9594fcffd8cd11b375bc5R556) takes `boolean` value as second argument but [the signature of this method](https://github.com/sailfishos/sailfish-qtcreator/pull/168/files#diff-9394f05edd7f6142dee6046a2957c186R86) declares the argument as `ExtraInfos` instead of `boolean`. So the plugin cannot be built.

When this issue will be fixed, I'll try to rebase. Or you can notify me here when WIP pull request will be accepted and I'll make a rebase from `master` again.

martyone:
> Call `fetchVirtualMachineInfo(vmName, true)` on [this line](https://github.com/sailfishos/sailfish-qtcreator/pull/168/files#diff-44c14b8d61c9594fcffd8cd11b375bc5R556) takes `boolean` value as second argument but [the signature of this method](https://github.com/sailfishos/sailfish-qtcreator/pull/168/files#diff-9394f05edd7f6142dee6046a2957c186R86) declares the argument as `ExtraInfos` instead of `boolean`. So the plugin cannot be built.

Fixed

yarfruct:
@martyone I did `git rebase upstream/jb44217`, resolved conflicts and made a push. Please, check resulting commit history.

martyone:
You need to rebase just the commits in your branch, not commits from master that are missing from jb44217. Ensure that there are just your commits on top of jb44217.

yarfruct:
@martyone 

> You need to rebase just the commits in your branch, not commits from master that are missing from jb44217. Ensure that there are just your commits on top of jb44217.

Done.

martyone:
Great, now please rebase again - I just rebased jb44217 so that we will get rid of the conflicts.

And please try to apply https://chris.beams.io/posts/git-commit/#seven-rules finally - I kindly suggested you already three times.

yarfruct:
@martyone 

> And please try to apply https://chris.beams.io/posts/git-commit/#seven-rules finally - I kindly suggested you already three times.

Please, write what is actually wrong with my commit messages.

martyone:
At least 6. Wrap the body at 72 characters

yarfruct:
@martyone Rebase is done and commit bodies are removed.

martyone:
Quite radical solution but well, the subjects are descriptive enough :)

How about renaming "Mer: Move the MerConnection, MerEmulatorDevice and MerDevice classes from the Internal namespace" as "Mer: Enable using MerEmulatorDevice from other plugins"?

yarfruct:
@martyone 

> How about renaming "Mer: Move the MerConnection, MerEmulatorDevice and MerDevice classes from the Internal namespace" as "Mer: Enable using MerEmulatorDevice from other plugins"?

Done.

martyone:
I changed the target branch back to master, as this is where it is going to be merged - yes, github is not very helpful here, where one PR is stacked on top of another one.

I will try to merge the underlying PRs and after that this one ASAP, likely during next week.

PR #170, opened 2019-02-12
Mer: Use kits added with sdktool instead of creating new ones

vigejolla:
Sometimes non-mer kits were visible in the configuration wizard, e.g.
"Desktop" kit. These could not be used for Sailfish OS projects, so it's
better to deregister them.

### DIFF
--- a/src/plugins/mer/mersdk.cpp
+++ b/src/plugins/mer/mersdk.cpp
@@ -529,17 +529,25 @@ bool MerSdk::addTarget(const MerTarget &target)
     QScopedPointer<MerQtVersion> version(target.createQtVersion());
     if (version.isNull())
         return false;
-    std::unique_ptr<Kit> kit = target.createKit();
-    if (!kit)
+
+    Kit* kitptr = target.kit();
+    std::unique_ptr<Kit> kit(nullptr);
+    if (!kitptr) {
+        kit = std::make_unique<Kit>();
+        kitptr = kit.get();
+    }
+
+    if (!target.finalizeKitCreation(kitptr))
         return false;
 
     ToolChainManager::registerToolChain(toolchain.data());
     ToolChainManager::registerToolChain(toolchain_c.data());
     QtVersionManager::addVersion(version.data());
-    QtKitInformation::setQtVersion(kit.get(), version.data());
-    ToolChainKitInformation::setToolChain(kit.get(), toolchain.data());
-    ToolChainKitInformation::setToolChain(kit.get(), toolchain_c.data());
-    KitManager::registerKit(std::move(kit));
+    QtKitInformation::setQtVersion(kitptr, version.data());
+    ToolChainKitInformation::setToolChain(kitptr, toolchain.data());
+    ToolChainKitInformation::setToolChain(kitptr, toolchain_c.data());
+    if (kit)
+        KitManager::registerKit(std::move(kit));
     toolchain.take();
     toolchain_c.take();
     version.take();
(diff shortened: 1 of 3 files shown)


martyone:
This would be incompatible with the aim to get our plugin adopted by upstream. I would prefer fixing it by filtering available kits instead.

vigejolla:
> This would be incompatible with the aim to get our plugin adopted by upstream. I would prefer fixing it by filtering available kits instead.

That would require changes outside mer plugin, which in turn will make all future upgrades more difficult.

martyone:
That's OK. This might be also a type of change that could be quickly accepted by upstream, so better to submit change to upstream gerrit immediately.

Kaffeine:
IIRC two years ago I managed to use SFOS QtC to build a project with both desktop and SFOS kits.
IMO it is important to avoid stuff that prevents other kits from working properly.

vigejolla:
> IIRC two years ago I managed to use SFOS QtC to build a project with both desktop and SFOS kits.
> IMO it is important to avoid stuff that prevents other kits from working properly.

I think that is a valid point. Once I get the filtering to work, I will add an option somewhere where it's possible to disable the filter.

vigejolla:
The current implementation filters non-mer kits out of the list, if there are any mer kits. If there are no mer kits, the normal filtering is used. I'm still going to add an option to make this configurable.

vigejolla:
> Replacing the internally set `requiredKitPredicate` seems like a bad idea. It is not only used during initial configuration but also later when enabling more kits I think. 

No it's not. Adding non-mer kits can be done later just fine, requiredKitPredicate only affects the initial project configuration wizard.

> But maybe the whole approach with filtering kits is a wrong idea, trying to fix a symptom instead of the root cause. Maybe we should instead add an option to disable autodetection, make it on-demand only, and make that the default in our installation. If the user wants to use tools not provided by the SDK, they can use autodetection explicitly or add them manually. The need to choose the right kit during project configuration in that case feels naturally.

What would that accomplish, in addition to what my patch does?

src/plugins/mer/mertarget.cpp
```diff
@@ -253,6 +253,39 @@ void MerTarget::ensureDebuggerIsSet(Kit *k) const
     }
 }
 
+void MerTarget::ensureToolChainIsSet(Kit *k) const
+{
+    if (ToolChainKitInformation::toolChains(k).count() > 0) {
```
martyone: nitpick: !isEmpty() and without braces :)

src/plugins/mer/mersdk.cpp
```diff
@@ -529,17 +529,26 @@ bool MerSdk::addTarget(const MerTarget &target)
     QScopedPointer<MerQtVersion> version(target.createQtVersion());
     if (version.isNull())
         return false;
-    std::unique_ptr<Kit> kit = target.createKit();
-    if (!kit)
+    Kit* kitptr = target.kit();
+    std::unique_ptr<Kit> kit(nullptr);
+    if (!kitptr) {
+        kit = target.createKit();
```
martyone: It took me a while to spot the trick again now that I returned to it after some time and I would have one suggestion: I would remove MerTarget::createKit, called make_unique<Kit>() directly here, move the sysroot check to finalizeKitCreation, let it return bool and check that here.

vigejolla: Thanks. This code is really ugly - let's see if it turns out any prettier with your suggestions.

vigejolla: Done like you suggested. It's a little bit prettier now. The annoying pointer juggling is still there, but I'm not sure if it's possible to completely get rid of it.

martyone:
Except the extra braces around one 'if' block it LGTM ;)

vigejolla:
> Except the extra braces around one 'if' block it LGTM ;)

Removed the extra braces

PR #172, opened 2019-02-18
Jb44818

martyone:


martyone:
Note that the commits by Sergey belong to PR #141

src/plugins/mer/meremulatordevicewidget.cpp
```diff
@@ -137,7 +139,7 @@ void MerEmulatorDeviceWidget::userNameEditingFinished()
         SshConnectionParameters sshParams = device->sshParameters();
         const QString& user = m_ui->userLineEdit->text();
         //TODO fix me:
-        const QString privKey = device->sharedConfigPath() +
+        const QString privKey = MerSdkManager::sdks().first()->sharedConfigPath() +
```
Kaffeine: `sdks()` returns non-const `QList` so here is a copy+detach.
`MerSdkManager::sdks().constFirst()` might be a bit faster but of course it is up to you to decide in this trade of readability vs a few CPU cycles optimization.

src/plugins/mer/mersdkmanager.cpp
```diff
@@ -625,7 +625,16 @@ bool MerSdkManager::generateSshKey(const QString &privKeyPath, QString &error)
 // this method updates the Mer devices.xml, nothing else
 void MerSdkManager::updateDevices()
 {
+    if (m_sdks.count() == 0)
```
Kaffeine: Consider `isEmpty()` (according to Qt convention for better readability)

src/plugins/mer/merrunconfigurationaspect.cpp
```diff
@@ -242,9 +255,20 @@ void MerRunConfigurationAspect::applyTo(ProjectExplorer::Runnable *r) const
                                    QLatin1String(QMLLIVE_SAILFISH_PRELOAD),
                                    QLatin1String(LIST_SEP));
 
-        if (qmlLiveIpcPort() > 0) {
+        int qmlLiveIpcPort = this->qmlLiveIpcPort();
```
Kaffeine: Maybe use `port`, `adjustedPort` or `adjustedQmlLiveIpcPort` instead of a method shadowing?

src/plugins/mer/meremulatordevicewidget.cpp
martyone: Hm, it never occurred to me that `first()` is different from `at(0)`... that's really ugly. Even uglier it is with e.g. `QMap` because there is no `QMap::constFirst`, just `constBegin`. And what is the worst is that you cannot easily tell if it would actually detach without knowing or guessing the implementation of the function on the left side. Like in this particular case - it would actually not detach I think, because `sdks()` returns a temporary. So I will leave as is. Good to know - thanks.

src/plugins/mer/merrunconfigurationaspect.cpp
martyone: I thought hardly what could be a more appropriate name.. I think we should rather keep using the same name.

src/plugins/mer/mersdkmanager.cpp
martyone: Fixed

src/plugins/mer/merrunconfigurationaspect.cpp
```diff
@@ -160,12 +161,24 @@ class MerRunConfigWidget : public QWidget
 
         // Init port QSpinBox
 
-        m_qmlLiveDetailsUi->port->setValue(m_aspect->qmlLiveIpcPort());
+        auto updatePort = [this] {
+            bool useDefault = m_aspect->qmlLiveIpcPort() <= 0;
```
vigejolla: Is using 0 or negative port for meaning "default" documented somewhere?

vigejolla: I guess I could state more explicitly why I'm asking: I'm somewhat worried because later on you are using a specific value (-1) to mean "default". What if some day someone decides that it's a great idea to use -2 to mean something else, like an error condition. It would be great to have some sort of assert here.

martyone: It's documented by exactly this line of code :) I am using -1 because it attracts more attention than 0. If we need to handle more cases in future, it will be simply rewritten - it is limited to the scope of this single file/module.

vigejolla: You are assuming that the next person who touches the code will notice this line of code here. Which is not a good assumption to make imho.

martyone: Look around. What you see is not an airplane firmware. It is a piece of trivial code serving a trivial use case. The variable in question is touched on a handful of places all in single file, easy to spot in a moment. I am not going to write comments like this or "this is a constructor, it initializes this class".

vigejolla: Now you misunderstood me. I wasn't asking for a comment, I was asking for something like QTC_ASSERT(m_aspect->qmlLiveIpcPort() == -1). Something to make it easy to spot when an error is made later on. I'm just asking to make the code easier for future developers to understand. This whole QtC code is far from trivial or easy to understand (which is not your fault), but there is no reason for us to be as so arrogant that we can't take the time to think about how to make the lives of future developers a bit easier. In my eyes complex and hard to understand code is bad code. Even if it works. Even if there are no known bugs. If there are undocumented pieces of required knowledge (like, -1 means default), it is definitely bad code.

vigejolla: I think the basic difference between my thinking and your thinking is that if a person A does not understand the code person B has written, I think it is because person B wrote bad code. You think it is because person A is stupid.

martyone: I think you misunderstood me as well :) Please check e.g. this code https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/plugins/debugger/debuggerengine.cpp . I will be honest - I do not understand it (yet). And I am certain it is not because it is written badly. Now compare it with our trivial case.

I also used to like Donald Knuth's idea about Literate programming, but then I had to deal with some real world code (and it's clear we are not working on an airplane firmware as I wrote) so I had to change my mind.

Regarding your your suggestion with assert, where would you put such an assertion? Assertion expresses and invariant that a block of code relies on. But there is no such requirement in the code I think, because while trying to balance between code correctness and robustness while maintaining the cost of coding I chose to follow the famous "Be generous in what you accept, rigorous in what you emit", and decided to accept any number less than 1 as the default while "emitting" -1.

vigejolla: Actaully, I don't think debuggerengine.cpp is that bad - you could have found better (or worse?) examples from QtC code. 

I also think that Knuth goes a bit too far with the Literate programming idea. But he has the right intention there. My perspective comes from the fact that I've had to work with quite a few developers - and I've found that they do not all think alike. Therefore I have found that writing code that is difficult to break saves from headaches later on. Even if we are not writing airplane software here.

The assert should be where you have assumed something. In this case you are assuming, that if qmlLiveIpcPort() < 0, it must be -1. And you are even sometimes using 0 to mean "use default", and -1 at other times, which increases confusion.

I myself like the "fail fast" approach more than "be generous in what you accept".

martyone: You wanted to say if qmlLiveIpcPort() < 1 :)

Fail fast works well for code 100% covered with tests. In other cases robustness is more desired. This is supported by the fact that QTC_ASSERT exists as a replacement for QT_ASSERT.

OK, let me enjoy some bikeshedding and fix it a proper way..

vigejolla: I like the new version :) Even though having a separate class for port number is a bit... enterprisey. But it was there already, so why not use it!

vigejolla: I think you just proved my original point ;-)

martyone: Do you mean the point you disproved yourself with "I wasn't asking for a comment"? :D

vigejolla: Eh.. no. Like I wrote there, I wasn't asking for a comment. Even though it might have seen like I was - I'm not that great with words. I really wanted you to write the code so that it expresses the idea directly - or if that was not possible, document the idea.

PR #173, opened 2019-02-21
Mer: Typo fixed

ghost:


martyone:
Thank you. The lower case use for "Sailfish OS build engine" is correct - it is used where it means a build engine in general, not the vanilla "Sailfish OS Build Engine" specifically. Similar applies to "Sailfish OS emulator" vs "Sailfish OS Emulator". Please use imperative mood in commit message subject. Refer e.g. to https://chris.beams.io/posts/git-commit/#seven-rules.

ids-imaging:
Thank you very much for the explanation.

martyone:
You are welcome. Please also use git-rebase to discard unneeded commits - in this case just the very last commit is needed (you will need to pass --force to git-push after that).

martyone:
I would prefer excluding the cosmetic changes. It is likely to conflict with other PRs in review and unnecessary pollutes the history. Ugly code is being fixed continuously when it needs to be touched.

PR #174, opened 2019-02-27
Mer: Use -Wall -Werror for GCC builds

martyone:


src/plugins/mer/meremulatordevice.cpp
```diff
@@ -154,6 +154,8 @@ private slots:
             setValue(0);
             setLabelText(tr("Error occured"));
             setCancelButtonText(tr("Close"));
+            close();
```
vigejolla: Could have just used [[fallthrough]]

martyone: Or rather Q_FALLTHROUGH, but I found it more clear not to do so in this case.

PR #176, opened 2019-03-05
Provide dumps of GCC macros and include paths

martyone:


src/plugins/mer/mertarget.cpp
```diff
@@ -148,7 +162,8 @@ QVariantMap MerTarget::toMap() const
 
 bool MerTarget::isValid() const
 {
-    return m_sdk && !m_name.isEmpty() && !m_qmakeQuery.isEmpty() && !m_gccMachineDump.isEmpty()
+    return !m_name.isEmpty() && !m_qmakeQuery.isEmpty() && !m_gccMachineDump.isEmpty()
```
vigejolla: I don't like the fact that you dropped m_sdk check here. Sure, it should never be null, but I would like to make sure, as programmers are known to make mistakes. I would consider even adding an assert in constructor.

src/plugins/mer/mertoolchain.cpp
```diff
@@ -157,18 +157,17 @@ QList<Task> MerToolChain::validateKit(const Kit *kit) const
...
-    if (m_headerPathsOnHost.isEmpty()) {
-        m_headerPathsOnHost.append(HeaderPath(sysRoot.toString() + QLatin1String("/usr/local/include"),
-                                             HeaderPathType::System));
-        m_headerPathsOnHost.append(HeaderPath(sysRoot.toString() + QLatin1String("/usr/include"),
-                                             HeaderPathType::System));
-    }
-    return m_headerPathsOnHost;
-}
-
+//ProjectExplorer::HeaderPaths MerToolChain::builtInHeaderPaths(const QStringList &cxxflags, const FileName &sysRoot) const
```
vigejolla: Please remove commented code before removing *WIP* :)

src/plugins/mer/mertarget.cpp
martyone: Me neither - it's a mistake I did while rebasing, thanks :)

src/plugins/mer/mertoolchain.cpp
martyone: Yup, I just need to recollect why I commented it out :)

PR #177, opened 2019-03-06
Mer: Disable support for job count parameter in toolchain

vigejolla:
Jb45044


vigejolla:
Naturally I tested it before pushing :)

PR #182, opened 2019-04-09
Mer: Don't use MerEmulatorModeDialog::m_dialog after calling exec

vigejolla:
If m_dialog parent is deleted while exec is running, also m_dialog is
deleted. Accessing m_dialog after exec in these cases may crash.

### DIFF
--- a/src/plugins/mer/meremulatormodedialog.cpp
+++ b/src/plugins/mer/meremulatormodedialog.cpp
@@ -57,7 +57,6 @@ using namespace Mer::Constants;
 MerEmulatorModeDialog::MerEmulatorModeDialog(QObject *parent)
     : QObject(parent),
       m_action(new QAction(this)),
-      m_dialog(0),
       m_ui(0),
       m_kit(0)
 {
@@ -81,7 +80,6 @@ MerEmulatorModeDialog::MerEmulatorModeDialog(QObject *parent)
 MerEmulatorModeDialog::MerEmulatorModeDialog(const MerEmulatorDevice::Ptr &emulator, QObject *parent)
     : QObject(parent),
       m_action(new QAction(tr(MER_EMULATOR_MODE_ACTION_NAME), this)),
-      m_dialog(nullptr),
       m_ui(nullptr),
       m_kit(nullptr)
 {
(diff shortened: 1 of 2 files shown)


src/plugins/mer/meremulatormodedialog.cpp
```diff
@@ -190,6 +190,9 @@ bool MerEmulatorModeDialog::execDialog()
     QTC_ASSERT(m_ui == 0, return false); // possible, but little issue..
 
     m_dialog = new QDialog(ICore::dialogParent());
+    connect (m_dialog, &QDialog::destroyed, [=] (QObject*) {
+        m_dialog = nullptr;
+    });
```
martyone: I would prefer storing m_dialog using QPointer instead

src/plugins/mer/meremulatormodedialog.cpp
```diff
@@ -222,7 +225,8 @@ bool MerEmulatorModeDialog::execDialog()
         m_ui->restartEmulatorCheckBox->setEnabled(false);
     }
 
-    if (m_dialog->exec() != QDialog::Accepted) {
+    int dialogResult = m_dialog->exec();
+    if (dialogResult != QDialog::Accepted) {
```
martyone: You can fix the style while touching this :)

src/plugins/mer/meremulatormodedialog.cpp
```diff
@@ -222,7 +225,8 @@ bool MerEmulatorModeDialog::execDialog()
         m_ui->restartEmulatorCheckBox->setEnabled(false);
     }
 
-    if (m_dialog->exec() != QDialog::Accepted) {
+    int dialogResult = m_dialog->exec();
```
martyone: Since we do not use it for anything else later, I would consider just moving the original line

`const bool result = m_dialog->result() == QDialog::Accepted;`

here.

src/plugins/mer/meremulatormodedialog.cpp
vigejolla: Good idea, will do

src/plugins/mer/meremulatormodedialog.cpp
vigejolla: Sure

src/plugins/mer/meremulatormodedialog.cpp
```diff
@@ -222,9 +220,9 @@ bool MerEmulatorModeDialog::execDialog()
         m_ui->restartEmulatorCheckBox->setEnabled(false);
     }
 
-    if (m_dialog->exec() != QDialog::Accepted) {
+    bool result = (m_dialog->exec() == QDialog::Accepted);
```
martyone: Possibly without parentheses and with const? :)

vigejolla: Parentheses improve readability so I'm not willing to remove them.

PR #189, opened 2019-04-26
Jb45644

martyone:


martyone:
No, there is no underlying problem. It is just because QProcess::terminate is implemented in a way more suitable for GUI programs - check https://doc.qt.io/qt-5/qprocess.html#terminate

vigejolla:
> No, there is no underlying problem. It is just because QProcess::terminate is implemented in a way more suitable for GUI programs - check https://doc.qt.io/qt-5/qprocess.html#terminate

Ok. I'm not in the office today and I don't have a windows machine here to test with, so I'm just going to have to take your word on it :)

PR #192, opened 2019-05-17
Implement base plugin for adding templates

etaishev:


martyone:
Currently the whole codebase for Sailfish SDK integration on top of vanilla Qt Creator is about 25k lines of C++ code. You want to add about 10k lines of C++ code (counted for the original PR #190). I would like to see most of these 10k lines contributed to upstream code, extending the current file/project wizards facilities in vanilla Qt Creator, and only a thin layer on top of that left on our side, mostly in form of data.

etaishev:
Like what part? Almost all of the projects associated with the addition of specific parts in spec/yaml. Upstream will need to make these solutions cross-platform

martyone:
I believe it just needs a carefully designed abstraction. Qt Creator comes with a mature API for wizards. If it lacks some functionality, we should try to improve it first and only resort to own implementation if it proves impossible to abstract.

etaishev:
I will try to rethink it. But most of the code is pages and editors

martyone:
Regarding wizards, consider e.g. the commit to add CppClassWizardFactory https://github.com/sailfishos/sailfish-qtcreator/pull/190/commits/5fabe55e42e7e00e42d6857e80812e82aac85c4d. Looking at the UIs I am confident ProjectExplorer::JsonWizardFactory provides good enough expressiveness to support our use case. Compare e.g. with https://github.com/sailfishos/sailfish-qtcreator/blob/master/share/qtcreator/templates/wizards/projects/qtquickapplication/scroll/wizard.json#L37. Fully custom pages are supported as well if some really specialized UI is required, through JsonWizardFactory::registerPageFactory like here https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/plugins/vcsbase/vcsplugin.cpp#L100.

Regarding editors, desktop file format is not totally Sailfish specific so there is a big potential to get most of the code accepted upstream and express Sailfish specific bits as a configuration layer.

For spectacle editor, currently the implementation is based on the wizard pages. Given that this way of editing is more suitable for beginners than average users, I would consider not implementing it as an editor and instead add an Core::InfoBar with a button that would enable launchich the wizard again on an existing spectacle document. With that, it should be again possible to implement it with JsonWizardFactory as a base and get necessary extensions upstream.

etaishev:
I saw code snippet wizards and main problem to rewrite it on JsonWizard will be that need post generator which will be add dependences in pro, spec, yaml files.

martyone:
Could you please expand more on this?

Back to what I wrote last time, I noticed that my suggestion "[...] instead add an Core::InfoBar with a button that would enable launching the wizard again on an existing spectacle document." would not be a good solution either. The main use case for the spectacle editor is to add dependencies conveniently - package description, version and other fields can be edited conveniently enough with the text editor. And adding dependencies is already supported by Qt Creator with some UI https://doc.qt.io/qtcreator/creator-project-qmake-libraries.html, so we should adapt that instead.

etaishev:
For example, in wizard adding desktop file need change pro file for add icons, or adding dependences in yaml file need export in pro file. 
In example **Adding Libraries to Projects** used custom #QmakeManager##  for adding snippet in pro file.
If we create own generator or do the same as **Adding Libraries to Projects** we need change upstream code, and if upstream it will take a long time

martyone:
Yes, just create an abstraction for the post generator in upstream code. Adding dependencies to .pro files, registering icons, .... these are all general purpose tasks that can definitely be supported by upstream code.

PR #193, opened 2019-05-23
Merge upstream v4.9.1

martyone:


src/plugins/mer/meremulatordevice.cpp
```diff
@@ -197,6 +197,7 @@ MerEmulatorDevice::MerEmulatorDevice(Core::Id id)
     , m_cpuCount(0)
     , m_vdiCapacityMb(0)
 {
+    init();
```
vigejolla: The upstream moved away from having separate init() in commit 54385ad67be075ac14d0edae5c451ed3d925edc8, maybe we should follow their example?

src/plugins/mer/merconnection.cpp
```diff
@@ -80,22 +80,20 @@ class MerConnectionRemoteShutdownProcess : public QObject
     MerConnectionRemoteShutdownProcess(QObject *parent)
         : QObject(parent)
         , m_runner(0)
-        , m_connectionError(SshNoError)
+        , m_connectionErrorOccured()
```
vigejolla: Please initialize your member variables

src/plugins/mer/meremulatordevice.cpp
martyone: Added commit 98780c6cab3880c179c8f45d14b1c6733d1b9b99 but decided to left init() there to not break history (it is not immediately adjacent to the constructor)

src/plugins/mer/merconnection.cpp
martyone: Fixed (typo)

martyone:
Newly added commits aafe420f2a1f326d7cd71cbc3fcd72586ccc009e and a9e0a6648a7a70ac26b74113626d4289feb04935 fix adding emulator manually.

vigejolla:
LGTM so far

PR #194, opened 2019-05-28
Mer: Not initialize factory snapshot when no snapshots exist

vigejolla:
MerEmualtorVMPage::handleEmulatorVmChanged sets the first snapshot
as factory snapshot. If there are no snapshots, this would cause
Qt Creator to crash.


martyone:
lgtm. I would prefer merging it only after 4.9 is merged.

PR #195, opened 2019-05-31
Jb41631

martyone:


src/plugins/mer/mertarget.cpp
```diff
@@ -426,6 +428,52 @@ bool MerTarget::createCacheFile(const QString &fileName, const QString &content)
     file.close();
     return ok;
 }
+
+bool MerTarget::createPkgConfigWrapper(const QString &targetPath) const
+{
+    const QFile::Permissions rwrw = QFile::ReadOwner|QFile::ReadUser|QFile::ReadGroup
```
vigejolla: This variable is only used once, on the next line as a base for rwxrwx, which is again used only once, on line 472. I don't really see any additional value in having either of the variables.

martyone: It is a copy of code that exists here https://github.com/sailfishos/sailfish-qtcreator/blob/4aec435c1b7201423ec2dcf0dad793ba18b38f21/src/plugins/mer/mertarget.cpp#L342 - I didn't want to touch the original location so I chose to rather duplicate it.

PR #196, opened 2019-05-31
Jb45932

martyone:


src/tools/sdktool/addmersdkoperation.cpp
```diff
@@ -326,6 +326,7 @@ int AddMerSdkOperation::execute() const
     if (map.isEmpty())
         map = initializeSdks(m_version);
 
+    map.insert(QLatin1String(Mer::Constants::MER_SDK_FILE_VERSION_KEY), version);
```
vigejolla: version seems to be undefined here

martyone: hm, it seems I tried building wrong subdir... fixed, thanks!

PR #200, opened 2019-06-18
QMakePM: Disable error dialog when deleting a file with wildcard

vigejolla:
When a file is removed from project, an error dialog was displayed if
the project file is not changed. If the file had been included in the
project because of a wildcard, this logic fails.

### DIFF
--- a/src/plugins/qmakeprojectmanager/qmakeparsernodes.cpp
+++ b/src/plugins/qmakeprojectmanager/qmakeparsernodes.cpp
@@ -584,6 +584,7 @@ bool QmakePriFile::addFiles(const QStringList &filePaths, QStringList *notAdded)
 bool QmakePriFile::removeFiles(const QStringList &filePaths,
                               QStringList *notRemoved)
 {
+    QStringList notChanged;
     QStringList failedFiles;
     using TypeFileMap = QMap<QString, QStringList>;
     // Split into lists by file type and bulk-add them.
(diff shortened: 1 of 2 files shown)


vigejolla:
I will upstream this, but I want some sanity check from you first :)

src/plugins/qmakeprojectmanager/qmakeparsernodes.h
```diff
@@ -234,6 +234,7 @@ class QMAKEPROJECTMANAGER_EXPORT QmakePriFile
     QSet<Utils::FileName> m_recursiveEnumerateFiles; // FIXME: Remove this?!
     QSet<QString> m_watchedFolders;
     bool m_includedInExactParse = true;
+    QMap<QString, QStringList> m_wildcardDirectoryContents;
```
martyone: It looks so lonely without the m_wildcardWatcher here - shouldn't we move it as well? And the associated code?

src/plugins/qmakeprojectmanager/qmakeparsernodes.cpp
```diff
@@ -594,7 +595,13 @@ bool QmakePriFile::removeFiles(const QStringList &filePaths,
     }
     foreach (const QString &type, typeFileMap.keys()) {
         const QStringList typeFiles = typeFileMap.value(type);
-        changeFiles(type, typeFiles, &failedFiles, RemoveFromProFile);
+        changeFiles(type, typeFiles, &notChanged, RemoveFromProFile);
+        foreach (const QString& file, notChanged) {
```
martyone: Space before ampersand

src/plugins/qmakeprojectmanager/qmakeparsernodes.h
vigejolla: That is a good question. I'll have to think about it.

vigejolla: Thinking about it I came to the conclusion that moving this to QmakePriFile was a wrong move. The results from expanding the wildcards are stored in the pro file level. So when a file removal is attempted in a pri file, the wildcards should be checked from the pro file.

PR #211, opened 2019-07-22
Jb46579

martyone:


martyone:
@sergey-levin please test if this works for you

src/libs/sfdk/sdk_p.h
```diff
@@ -82,6 +87,7 @@ class SdkPrivate : public QObject
     std::unique_ptr<BuildEngineManager> buildEngineManager;
     std::unique_ptr<EmulatorManager> emulatorManager;
     std::unique_ptr<DeviceManager> deviceManager;
+    QString customVBoxManagePath_;
```
vigejolla: What's with the underscores in member variable names? What's wrong with the traditional m_ prefix?

martyone: By convention, m_ prefixes are not used in private classes and at the same time getters are not named with "get" prefix, which sometimes leads to a situation like this. Underscore suffix on either name fixes it.

vigejolla: Wouldn't it then make sense to name all member variables consistently with the underscore?

martyone: nope, it's just to deal with the exceptional cases. It's less disturbing than using "artificial names". In general, whenever you see an underscore suffix in a name, you know immediately it represents just another "view" of an entity that differs just slightly - e.g. by the data type, state, read/write access - but otherwise meaning the same. Also the analogy with the use of apostrophe in mathematics helps to get more natural feeling about it. Pity C++ doesn't accept apostrophes in names (yet) :( https://stackoverflow.com/questions/5673916/apostrophe-in-identifiers-in-haskell

vigejolla: I'm sure c++23 will fix that with the possibility to use a simple std::variable<Myclass::member>(myvariable') notation ;)

martyone: you pessimist :)

PR #214, opened 2019-07-26
Mer: Add russian translation for Sailfish OS plugin

etaishev:


martyone:
There are translations added for messages that are not found in this tree. These should be excluded.

etaishev:
Remove it

martyone:
@etaishev could you please rebase it onto current master?

etaishev:
Rebased

PR #215, opened 2019-07-26
Add show global modal progress bar for apply vm changes

etaishev:
The main bug: Add some changes for vm, press **Apply**, immediately press **Ok** -> Qt Creator show error that can't lock down vm and crash.

### DIFF
--- a/src/plugins/mer/meroptionspage.cpp
+++ b/src/plugins/mer/meroptionspage.cpp
@@ -66,7 +66,8 @@ void MerOptionsPage::apply()
 
 void MerOptionsPage::finish()
 {
-    delete m_widget;
+    if(m_widget)
+        m_widget->close();
 }
 
 bool MerOptionsPage::matches(const QString &key) const
--- a/src/plugins/mer/meroptionswidget.cpp
+++ b/src/plugins/mer/meroptionswidget.cpp
@@ -151,8 +151,13 @@ void MerOptionsWidget::store()
 
     disconnect(m_sdksUpdatedConnection);
 
+    QString errors;
+
     QList<MerSdk*> lockedDownSdks;
     ok &= lockDownConnectionsOrCancelChangesThatNeedIt(&lockedDownSdks);
+    if (!ok)
+        errors.append("\nFailed lock down VM connections");
+
 
     foreach (MerSdk *sdk, sdks) {
         progress.setLabelText(tr("Applying virtual machine settings: '%1'").arg(sdk->virtualMachineName()));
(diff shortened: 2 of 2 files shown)


src/plugins/mer/meroptionswidget.cpp
```diff
@@ -135,14 +135,15 @@ QString MerOptionsWidget::searchKeyWordMatchString() const
...
-    QProgressDialog progress(this);
+    QProgressDialog progress(ICore::mainWindow());
     progress.setWindowModality(Qt::WindowModal);
     QPushButton *cancelButton = new QPushButton(tr("Abort"), &progress);
     cancelButton->setDisabled(true);
     progress.setCancelButton(cancelButton);
     progress.setMinimumDuration(2000);
     progress.setMinimum(0);
     progress.setMaximum(0);
+    progress.show();
```
martyone: It is not desired to call show here. Instead, setting minimum duration orders for showing the dialog only if the operation takes more time than anticipated.

src/plugins/mer/meroptionswidget.cpp
```diff
@@ -151,8 +151,13 @@ void MerOptionsWidget::store()
 
     disconnect(m_sdksUpdatedConnection);
 
+    QString errors;
+
     QList<MerSdk*> lockedDownSdks;
     ok &= lockDownConnectionsOrCancelChangesThatNeedIt(&lockedDownSdks);
+    if (!ok)
+        errors.append("\nFailed lock down VM connections");
```
martyone: All messages should be internationalized. In case of those that correspond to form values I would take the string from the label. After that using a list and join('\n') later will become more convenient.

src/plugins/mer/meroptionspage.cpp
```diff
@@ -66,7 +66,8 @@ void MerOptionsPage::apply()
 
 void MerOptionsPage::finish()
 {
-    delete m_widget;
+    if(m_widget)
+        m_widget->close();
```
martyone: I checked implementation of other options pages, most of them do delete m_widget in their finish(). To me it is a sign this is not fixing the actual source of the bug. Could you please recheck?

etaishev: Hm. Other way to fix problem use `loop->exec(QEventLoop::ExcludeUserInputEvents);` in MerConnection::vmPollState(Synchronous) to really by a synchronous func.

martyone: OK, now I see where the issues comes from. There are changes in progress that are likely to influence this - let's wait with this PR and recheck later after that changes are completed.

etaishev:
No longer relevant, fixed with last update master. Close it

PR #217, opened 2019-08-08
Jb45548

martyone:


martyone:
> LGTM, I guess. Someone should write a tool for generating proper human readable diffs for these multiline json strings...

Yeah, it's terrible. `git diff --color-words` helps a bit.

PR #219, opened 2019-08-29
jb46741

jlehtoranta:
Force the left alignment on MacOS. The items are otherwise sometimes centered, if there's space for it.

### DIFF
--- a/src/plugins/mer/mersdkdetailswidget.ui
+++ b/src/plugins/mer/mersdkdetailswidget.ui
@@ -53,6 +53,9 @@
           <string>Virtual machine</string>
          </property>
          <layout class="QFormLayout" name="formLayout">
+          <property name="formAlignment">
+           <set>Qt::AlignLeading|Qt::AlignLeft|Qt::AlignTop</set>
+          </property>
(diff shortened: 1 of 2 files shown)


src/plugins/mer/mersdkdetailswidget.ui
```diff
@@ -53,6 +53,9 @@
           <string>Virtual machine</string>
          </property>
          <layout class="QFormLayout" name="formLayout">
+          <property name="formAlignment">
+           <set>Qt::AlignLeading|Qt::AlignLeft|Qt::AlignTop</set>
```
vigejolla: At least according to Qt documentation, AlignLeading and AlignLeft are synonyms

jlehtoranta: I noticed that too, but the same combination is used also elsewhere in this file. Should I remove it from the other lines too?

vigejolla: I have no strong opinions on whether you should remove it from the other lines or not. But if you do remove it, you should do it in a separate commit.

PR #220, opened 2019-09-11
Fix save settings on first start on Windows

etaishev:
Fix problem that in Windows the "settings" folder is installed on the disk from which the installer was started

Probably fix Jb45141

### DIFF
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -64,6 +64,10 @@
 #include <vector>
 #include <iterator>
 
+#ifdef Q_OS_WIN
+#include <windows.h>
+#endif
+
 #ifdef ENABLE_QT_BREAKPAD
 #include <qtsystemexceptionhandler.h>
 #endif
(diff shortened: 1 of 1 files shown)


src/app/main.cpp
```diff
@@ -64,6 +64,11 @@
 #include <vector>
 #include <iterator>
 
+#ifdef Q_OS_WIN
+#include <windows.h>
+#endif
+
+
```
vigejolla: remove extra empty line

src/app/main.cpp
```diff
@@ -445,8 +450,14 @@ int main(int argc, char **argv)
 #ifdef Q_OS_WIN
     else {
         // set the windows settings userdir to the install dir
-        QDir rootDir = QApplication::applicationDirPath();
+        // do not use QApplication::applicationDirPath() befor create QApplication, becouse
+        // if installer started from other drive (D:\, E:\ etc.) it return the launch dir, not the app dir
```
vigejolla: I think it would be better to not write about the installer here - it should be sufficient to say that it returns the launch dir instead of app dir.

etaishev:
Update on current master

PR #221, opened 2019-10-10
Jb45560

martyone:


src/libs/sfdk/asynchronous.h
```diff
@@ -0,0 +1,57 @@
...
+template<typename Fn, typename... InArgs, typename... OutArgs>
+void execAsynchronous(std::tuple<OutArgs&...> outArgs, const Fn &fn, InArgs&&... inArgs)
+{
+	QEventLoop loop;
+	auto whenDone = [&loop, &outArgs](OutArgs...as) {
+		loop.quit();
+		outArgs = std::tie(as...);
+	};
+	fn(std::forward<InArgs>(inArgs)..., &loop, whenDone);
+	loop.exec();
```
vigejolla: If fn() happens to be synchronous (i.e. calls whenDone before returning), loop.exec() blocks forever. I would consider setting a flag in whenDone, and not calling exec() if whenDone has been already called. This way the caller would not have to care whether the implementation is asynchronous or not.

src/plugins/mer/merabstractvmstartstep.h
```diff
@@ -1,6 +1,6 @@
 /****************************************************************************
 **
-** Copyright (C) 2014 Jolla Ltd.
+** Copyright (C) 2014-2015,2019 Jolla Ltd.
```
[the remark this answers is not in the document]
martyone: Check https://www.gnu.org/licenses/gpl-howto.html

> For software with several releases over multiple years, it's okay to use a range (“2008-2010”) instead of listing individual years (“2008, 2009, 2010”) if and only if every year in the range, inclusive, really is a “copyrightable” year that would be listed individually; and you make an explicit statement in your documentation about this usage.

src/libs/sfdk/asynchronous.h
martyone: It is designed so that it is guaranteed that any two asynchronous calls would complete in the same order in which they were initiated. It makes things easier a lot on the caller side. For that to work, every asynchronous function must be really asynchronous. If you need to fail an asynchronous function immediately, use something like `commandQueue()->enqueueCheckPoint(context, std::bind(functor, false))` to achieve proper serialization.

src/plugins/mer/merabstractvmstartstep.h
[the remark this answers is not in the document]
vigejolla: That howto talks about releases. I don't think they mean that each and every file has to have modifications during the year. But sure, using individual years is definitely correct.

src/libs/sfdk/sfdkconstants.h
```diff
@@ -0,0 +1,32 @@
...
+
+#pragma once
+
+namespace Sfdk {
+namespace Constants {
+
+const int DEFAULT_QML_LIVE_PORT = 10234;
+const int MAX_QML_LIVE_PORTS = 10;
+
+} // namespace Log
```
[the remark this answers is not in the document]
martyone: Fixed

src/libs/sfdk/asynchronous.h
vigejolla: There was no mention about the order preservation, so I didn't know that. This could of course be implemented so that it would do also that even with synchronous functions, but that would require a bit more code here.

My issue is not with failing function immediately, more that when the function implementation is more or less no-op, making it asynchronous feels like a lot of wasted lines of code.

src/libs/sfdk/virtualmachine.cpp
```diff
@@ -0,0 +1,798 @@
...
+            d->virtualMachineInfo.vdiCapacityMb = vdiCapacityMb;
+            VirtualMachineInfoCache::insert(uri(), d->virtualMachineInfo);
+            emit vdiCapacityMbChanged(vdiCapacityMb);
+        }
+        if (context_)
+            functor(ok);
+    });
+}
+
+bool VirtualMachine::hasPortForwarding(quint16 hostPort, QString *ruleName) const
```
vigejolla: I'm not sure the name of this method is the best possible. I get the feeling that the main purpose of the method is to get the name of the rule instead of just checking the existence of the rule.

src/libs/sfdk/virtualmachine.h
```diff
@@ -0,0 +1,211 @@
...
+    void addPortForwarding(const QString &ruleName, const QString &protocol,
+            quint16 hostPort, quint16 emulatorVmPort, const QObject *context,
+            const Functor<bool> &functor);
+    void removePortForwarding(const QString &ruleName, const QObject *context,
+            const Functor<bool> &functor);
+
+    QStringList snapshots() const;
+
+    void refreshConfiguration(const QObject *context, const Functor<bool> &functor);
+    // FIXME API style?
```
vigejolla: Suspicious comment. Is this something you intended to fix before merging?

src/libs/sfdk/virtualmachine.cpp
martyone: The main purpose is really to check if a port is used or free to use. The optional out-argument ruleName exists to allow verification (in the former case) that the rule is the expected one.

vigejolla: Now I just noticed that the method is not even called anywhere. Is it even needed at all?

src/libs/sfdk/virtualmachine.h
martyone: Not planned to be fixed anytime soon. The same applies to all the other TODO and FIXME comments left there. They highlight the ideas for further development with no schedule set.

src/libs/sfdk/virtualmachine.cpp
martyone: It is used by the sensors emulation plugin developed by FRUCT https://github.com/sailfishos/sailfish-qtcreator/pull/165/files#diff-af529b4089f9cb42afa0b6dca536bbf2R146

vigejolla: Ah, ok then. I would have done it the other way around, by including the additional optional check in hasPortForwarding() instead of having the caller do it. But I guess that ship has already sailed :-/

martyone: What you prefer would not allow to query the rule name :)

vigejolla: > What you prefer would not allow to query the rule name :)

That was exactly why I started asking about the method name in the first place. If I want to query the rule name, "hasPortForwarding()" is not the first method I will look at. I would first try to find if there is a method called "portForwadingRuleName()" or something like that. But then you wrote that it's not actually used for that but instead for checking that the name is what expected, i.e. the name is already known. In which case the caller is not interested in querying the rule name but verifying that it really is what it should be. So which is it? Is it used for querying the rule name, or checking that the rule name matches? In the first case the method name is not the best possible, and in the latter case the implementation could be done here, instead of everywhere where this method is called.

martyone:
```
QString ruleName;
if (vm->hasPortForwarding(PORT, &ruleName)) {
     if (ruleName != RULE_NAME)
           throw new StolenPortError(ruleName);
} else {
   vm->addPortForwarding(RULE_NAME, ...);
}
```

martyone: Nitpick: in the original implementation, it saved one VBoxManage invocation, hence the design :)

src/libs/sfdk/sdk.cpp
```diff
@@ -53,4 +90,59 @@ SdkPrivate::SdkPrivate(Sdk *q)
 
 SdkPrivate::~SdkPrivate() = default;
 
+Utils::FileName SdkPrivate::settingsFile(SettingsScope scope, const QString &basename)
+{
+    const QString prefix = scope == SessionScope
+        ? QString::fromLatin1(Constants::LIB_ID) + '-'
+        : QString();
```
vigejolla: Even though this is syntactically valid, and probably does what you intend it to do, please don't write code like this. For beginner programmers like me, code like this is really difficult to read. There is really no need to put assignments and comparisons to a single line. ?: operator is difficult enough to parse as it is.

src/libs/sfdk/sdk.cpp
```diff
@@ -53,4 +90,59 @@ SdkPrivate::SdkPrivate(Sdk *q)
...
+        : QString::fromLatin1(Constants::LIB_ID);
+
+    QSettings settings(QSettings::IniFormat, qscope, QCoreApplication::organizationName(),
+            applicationName);
+
+    // See ICore::userResourcePath()
+    QTC_CHECK(settings.fileName().endsWith(".ini"));
+    const auto iniInfo = QFileInfo(settings.fileName());
+    const QString resourceDir = iniInfo.completeBaseName().toLower();
+    *location = FileName::fromString(iniInfo.path() + '/' + resourceDir);
```
vigejolla: I have a feeling this could use some comments. I have no doubt the code above works, and I also believe there are reasons for doing the things that are done, but for the future coder who has to fix bugs here, it might make things faster if here were some explanations. Why are we using QSettings to generate a filename, which we don't even use (but use as a directory name instead). Why are we converting the filename to lower case?

src/libs/sfdk/sdk.cpp
martyone: You mean you would prefer

```
bool isSessionScope;
if (scope == SessionScope)
    isSessionScope = true;
else
    isSessionScope = false;

```
over

```
bool isSessionScope = scope == SessionScope;
```

or what?

Coding conventions exist to make code easier to read. When coding conventions are followed, you do not need to read every character of a code to understand its structure. This is the first knowledge a beginner programmer should adopt. Like breaking longer or otherwise complex lines with ternary operator involved, which is also a special case for the "one statement on a line" rule.

```
coreplugin/dialogs/addtovcsdialog.cpp-    QString addTo = files.size() == 1
coreplugin/dialogs/addtovcsdialog.cpp:            ? tr("Add the file to version control (%1)").arg(vcsDisplayName)
coreplugin/dialogs/addtovcsdialog.cpp-            : tr("Add the files to version control (%1)").arg(vcsDisplayName);
--
coreplugin/dialogs/externaltoolconfig.cpp-    const QString placeholderText = Utils::HostOsInfo::isWindowsHost()
coreplugin/dialogs/externaltoolconfig.cpp:            ? tr("PATH=C:\\dev\\bin;${PATH}")
coreplugin/dialogs/externaltoolconfig.cpp-            : tr("PATH=/opt/bin:${PATH}");
--
coreplugin/dialogs/saveitemsdialog.cpp-    const QDialogButtonBox::ButtonRole discardButtonRole = Utils::HostOsInfo::isMacHost()
coreplugin/dialogs/saveitemsdialog.cpp:            ? QDialogButtonBox::ResetRole : QDialogButtonBox::DestructiveRole;
coreplugin/dialogs/saveitemsdialog.cpp-
--
coreplugin/dialogs/shortcutsettings.cpp-    shortcutLabel->setToolTip(Utils::HostOsInfo::isMacHost()
coreplugin/dialogs/shortcutsettings.cpp:           ? QLatin1String("<html><body>")
coreplugin/dialogs/shortcutsettings.cpp-             + tr("Use \"Cmd\", \"Opt\", \"Ctrl\", and \"Shift\" for modifier keys. "
--
coreplugin/dialogs/shortcutsettings.cpp-    item->m_item->setForeground(2, hasCollision
coreplugin/dialogs/shortcutsettings.cpp:                                ? Utils::creatorTheme()->color(Utils::Theme::TextColorError)
coreplugin/dialogs/shortcutsettings.cpp-                                : commandList()->palette().window());
--
coreplugin/editormanager/openeditorswindow.cpp-    item->setIcon(0, !entry->fileName().isEmpty() && entry->document->isFileReadOnly()
coreplugin/editormanager/openeditorswindow.cpp:                  ? DocumentModel::lockedIcon() : m_emptyIcon);
coreplugin/editormanager/openeditorswindow.cpp-    item->setText(0, title);
--
coreplugin/editormanager/editormanager.cpp-        revertToSavedAction->setText(document->isModified()
coreplugin/editormanager/editormanager.cpp:                                     ? tr("Revert %1 to Saved").arg(quotedName)
coreplugin/editormanager/editormanager.cpp-                                     : tr("Reload %1").arg(quotedName));
--
coreplugin/editormanager/editormanager.cpp-    const QString windowTitleAddition = d->m_titleAdditionHandler
coreplugin/editormanager/editormanager.cpp:            ? d->m_titleAdditionHandler(filePath)
coreplugin/editormanager/editormanager.cpp-            : QString();
--
coreplugin/editormanager/editormanager.cpp-    const QString windowTitleVcsTopic = d->m_titleVcsTopicHandler
coreplugin/editormanager/editormanager.cpp:           ? d->m_titleVcsTopicHandler(filePath)
coreplugin/editormanager/editormanager.cpp-           : QString();
--
coreplugin/editormanager/editormanager.cpp-    const QString sessionTitle = d->m_sessionTitleHandler
coreplugin/editormanager/editormanager.cpp:           ? d->m_sessionTitleHandler(filePath)
coreplugin/editormanager/editormanager.cpp-           : QString();
--
coreplugin/editormanager/editormanager.cpp-    d->m_closeCurrentEditorContextAction->setText(entry
coreplugin/editormanager/editormanager.cpp:                                                    ? tr("Close \"%1\"").arg(quotedDisplayName)
coreplugin/editormanager/editormanager.cpp-                                                    : tr("Close Editor"));
coreplugin/editormanager/editormanager.cpp-    d->m_closeOtherDocumentsContextAction->setText(entry
coreplugin/editormanager/editormanager.cpp:                                                   ? tr("Close All Except \"%1\"").arg(quotedDisplayName)
coreplugin/editormanager/editormanager.cpp-                                                   : tr("Close Other Editors"));
--
coreplugin/find/basetextfind.cpp-            candidate.movePosition(options & QTextDocument::FindBackward
coreplugin/find/basetextfind.cpp:                                   ? QTextCursor::PreviousCharacter
coreplugin/find/basetextfind.cpp-                                   : QTextCursor::NextCharacter);
--
coreplugin/find/highlightscrollbarcontroller.cpp-    const double backgroundRatio = sizeDocInvisible
coreplugin/find/highlightscrollbarcontroller.cpp:            ? ((double)scrollBarBackgroundHeight / sizeDocInvisible) : 0;
coreplugin/find/highlightscrollbarcontroller.cpp-
--
coreplugin/find/highlightscrollbarcontroller.cpp-    const double handleRatio = sizeDocVisible
coreplugin/find/highlightscrollbarcontroller.cpp:            ? ((double)handleRect.height() / sizeDocVisible) : 0;
coreplugin/find/highlightscrollbarcontroller.cpp-
--
coreplugin/find/searchresulttreeitemdelegate.cpp-    QPalette::ColorGroup cg = option.state & QStyle::State_Enabled
coreplugin/find/searchresulttreeitemdelegate.cpp:                              ? QPalette::Normal : QPalette::Disabled;
coreplugin/find/searchresulttreeitemdelegate.cpp-    if (cg == QPalette::Normal && !(option.state & QStyle::State_Active))
--
coreplugin/find/searchresultwindow.cpp-    color.highlightBackground = highlightBackgroundColor.isValid()
coreplugin/find/searchresultwindow.cpp:            ? highlightBackgroundColor
coreplugin/find/searchresultwindow.cpp-            : textBackgroundColor;
coreplugin/find/searchresultwindow.cpp-    color.highlightForeground = highlightForegroundColor.isValid()
coreplugin/find/searchresultwindow.cpp:            ? highlightForegroundColor
coreplugin/find/searchresultwindow.cpp-            : textForegroundColor;
--
coreplugin/locator/locatorwidget.cpp-        const int highlightColumn = entry.highlightInfo.dataType == LocatorFilterEntry::HighlightInfo::DisplayName
coreplugin/locator/locatorwidget.cpp:                                                                 ? DisplayNameColumn
coreplugin/locator/locatorwidget.cpp-                                                                 : ExtraInfoColumn;
--
coreplugin/documentmanager.cpp-            fileDialogPath = defaultPath + (suggestedName.isEmpty()
coreplugin/documentmanager.cpp:                    ? QString()
coreplugin/documentmanager.cpp-                    : '/' + suggestedName);
--
coreplugin/editortoolbar.cpp-    d->m_editorList->setToolTip(document->filePath().isEmpty()
coreplugin/editortoolbar.cpp:                                ? document->displayName()
coreplugin/editortoolbar.cpp-                                : document->filePath().toUserOutput());
--
coreplugin/fancyactionbar.cpp-    const QIcon::Mode iconMode = isEnabled()
coreplugin/fancyactionbar.cpp:                                     ? ((isDown() || isChecked()) ? QIcon::Active : QIcon::Normal)
coreplugin/fancyactionbar.cpp-                                     : QIcon::Disabled;
--
coreplugin/fancyactionbar.cpp-                                      // state as they would do on the mode buttons.
coreplugin/fancyactionbar.cpp:                                      ? Theme::FancyTabWidgetEnabledUnselectedTextColor
coreplugin/fancyactionbar.cpp-                                      : Theme::FancyTabWidgetDisabledUnselectedTextColor));
--
coreplugin/iwizardfactory.cpp-            path = DocumentManager::useProjectsDirectory()
coreplugin/iwizardfactory.cpp:                       ? DocumentManager::projectsDirectory().toString()
coreplugin/iwizardfactory.cpp-                       : DocumentManager::fileDialogLastVisitedDirectory();
--
coreplugin/manhattanstyle.cpp-            const QColor color = creatorTheme()->color(enabled
coreplugin/manhattanstyle.cpp:                                                       ? Theme::MenuItemTextColorNormal
coreplugin/manhattanstyle.cpp-                                                       : Theme::MenuItemTextColorDisabled);
--
coreplugin/manhattanstyle.cpp-            pal.setBrush(QPalette::ButtonText, dis
coreplugin/manhattanstyle.cpp:                ? creatorTheme()->color(Theme::MenuBarItemTextColorDisabled)
coreplugin/manhattanstyle.cpp-                : creatorTheme()->color(Theme::MenuBarItemTextColorNormal));
--
coreplugin/manhattanstyle.cpp-                pal.setBrush(QPalette::Text, creatorTheme()->color(dis
coreplugin/manhattanstyle.cpp:                                                                   ? Theme::IconsDisabledColor
coreplugin/manhattanstyle.cpp-                                                                   : Theme::PanelTextColorLight));
--
coreplugin/manhattanstyle.cpp-                painter->setPen((option->state & State_Enabled)
coreplugin/manhattanstyle.cpp:                                  ? option->palette.color(QPalette::Foreground)
coreplugin/manhattanstyle.cpp-                                  : creatorTheme()->color(Theme::IconsDisabledColor));
--
coreplugin/manhattanstyle.cpp-                    const QColor hightLight = creatorTheme()->flag(Theme::FlatToolBars)
coreplugin/manhattanstyle.cpp:                            ? creatorTheme()->color(Theme::FancyToolBarSeparatorColor)
coreplugin/manhattanstyle.cpp-                            : StyleHelper::sidebarHighlight();
coreplugin/manhattanstyle.cpp-                    const QColor borderColor = drawLightColored
coreplugin/manhattanstyle.cpp:                            ? QColor(255, 255, 255, 180) : hightLight;
coreplugin/manhattanstyle.cpp-                    if (widget && widget->property("topBorder").toBool()) {
--
coreplugin/navigationwidget.cpp-    const char *trToolTip = d->m_side == Side::Left
coreplugin/navigationwidget.cpp:                                ? (isShown() ? Constants::TR_HIDE_LEFT_SIDEBAR : Constants::TR_SHOW_LEFT_SIDEBAR)
coreplugin/navigationwidget.cpp-                                : (isShown() ? Constants::TR_HIDE_RIGHT_SIDEBAR : Constants::TR_SHOW_RIGHT_SIDEBAR);
--
coreplugin/navigationwidget.cpp-    return side == Side::Left
coreplugin/navigationwidget.cpp:            ? Utils::Icons::CLOSE_SPLIT_LEFT.icon()
coreplugin/navigationwidget.cpp-            : Utils::Icons::CLOSE_SPLIT_RIGHT.icon();
--
coreplugin/outputpane.cpp-    int minimum = (d->m_splitter->orientation() == Qt::Vertical
coreplugin/outputpane.cpp:                   ? om->sizeHint().height() : om->sizeHint().width());
coreplugin/outputpane.cpp-    if (nonMaximizedSize() < minimum && !d->m_isMaximized)
--
coreplugin/outputpanemanager.cpp-        QRect r = creatorTheme()->flag(Theme::FlatToolBars)
coreplugin/outputpanemanager.cpp:                  ? rect() : rect().adjusted(numberAreaWidth(), 1, -1, -1);
coreplugin/outputpanemanager.cpp-        p.fillRect(r, c);
--
coreplugin/sidebar.cpp-    d->m_widgets.at(0)->setCloseIcon(d->m_widgets.size() == 1
coreplugin/sidebar.cpp:                                     ? Utils::Icons::CLOSE_SPLIT_LEFT.icon()
coreplugin/sidebar.cpp-                                     : Utils::Icons::CLOSE_SPLIT_TOP.icon());
--
coreplugin/sidebar.cpp-            d->m_widgets.at(0)->setCloseIcon(d->m_widgets.size() == 1
coreplugin/sidebar.cpp:                                             ? Utils::Icons::CLOSE_SPLIT_LEFT.icon()
coreplugin/sidebar.cpp-                                             : Utils::Icons::CLOSE_SPLIT_TOP.icon());
--
coreplugin/vcsmanager.cpp-    return files.size() == 1
coreplugin/vcsmanager.cpp:        ? tr("Add the file\n%1\nto version control (%2)?")
coreplugin/vcsmanager.cpp-              .arg(files.front(), vc->displayName())
--
coreplugin/vcsmanager.cpp-    return files.size() == 1
coreplugin/vcsmanager.cpp:        ? tr("Could not add the file\n%1\nto version control (%2)\n")
coreplugin/vcsmanager.cpp-              .arg(files.front(), vc->displayName())
```

src/libs/sfdk/sdk.cpp
martyone: There is a comment. Look a bit up.

src/libs/sfdk/sdk.cpp
vigejolla: > You mean you would prefer
> 
> ```
> bool isSessionScope;
> if (scope == SessionScope)
>     isSessionScope = true;
> else
>     isSessionScope = false;
> ```
> 
> over
> 
> ```
> bool isSessionScope = scope == SessionScope;
> ```
> 
> or what?

If I have to choose between those two, then yes, I would prefer the if - else version. But this I would like the most:

`bool isSessionScope = (scope == SessionScope);`

The use of parenthesis improves the code readability a *lot*. The reader does not have to wonder about operator precedence and he can think about more important issues.

And also separating the boolean initialization and the ternary operator like this:

```
bool isSessionScope = (scope == SessionScope);
QString prefix = isSessionScope ? foo : bar;
```

again makes it easier for the poor fellow who has to fix bugs in the code. I don't really like the ternary operator, it never makes the code easier to read but just the opposite. But if one really can't be bothered to write if/else blocks, at least the ternary operators could be used in a fashion which is easy to read and understand. Also, having the boolean in a separate variable makes it easier to debug the code.

Yes, I know there are plenty of examples in Qt Creator sources where the code is far from readable and understandable. It doesn't make it any better though. The guys behind Qt Creator are really smart and good programmers, but I wouldn't say that they are considerate towards fellow programmers who have to read their code. To me readable and easy to understand code is always better than code which uses neat tricks to perform stuff using the least amount of keystrokes. 

I'm also quite sure there isn't a Qt Creator specific coding convention which would require the programmer to stuff as much as possible to a single statement. I might be wrong about this though.

I hope you don't read this so that I would think that more code is always the better - quite the opposite is actually true.

Also, please don't take these comments as requirements for changing your current code. Instead I'm asking that you could perhaps consider the reader more when writing your future code.

src/libs/sfdk/sdk.cpp
vigejolla: You mean "See ICore::userResourcePath()"? 

I did look at ICore::userResourcePath() before writing my comment. It didn't really answer my questions.

vigejolla: But the implementation here is not identical to ICore::userResourcePath(). Looking at ICore::userResourcePath() does not tell me why are we using settings file name as a directory name (userResourcePath() does not do that). It also does not tell me why the filename is turned to lower case before using as a directory name, as userResourcePath does not do that either.

martyone: identical != similar, (userResourcePath() does not do that) != (code involved in the ICore::userResourcePath() call does not do that)

vigejolla: Well, userResourcePath() or any code called by userResourcePath() does not do that. Some code which calls userResourcePath() might do something similar, but maybe you should then hint the reader towards that code? Or maybe just replace the comment with "// see Qt Creator code"? It would be just as helpful...

src/libs/sfdk/buildengine.cpp
```diff
@@ -0,0 +1,1133 @@
...
+    return SdkPrivate::settingsLocation(SdkPrivate::UserScope)
+        .appendPath(Constants::BUILD_TARGET_TOOLS);
+}
+
+/*!
+ * \class BuildEngine::PrivateConstructorTag
+ * \internal
+ */
+
+struct BuildEngine::PrivateConstructorTag {};
```
vigejolla: What's the point of PrivateConstructorTag? What improvement is achieved by having to call BuildEngine(parent, PrivateConstructorTag{}) instead of just BuildEngine(parent)?

martyone: https://stackoverflow.com/questions/32351296/factory-pattern-with-private-constructors-in-c

martyone: Or better https://stackoverflow.com/questions/8147027/how-do-i-call-stdmake-shared-on-a-class-with-only-protected-or-private-const

vigejolla: That's what I thought this code tried to accomplish. 
But in our case PrivateConstructorTag::PrivateConstructorTag() is public, so anyone can just call BuildEngine(parent, PrivateConstructorTag{}). 
In your first link the tag is private (with a friend Factory class), and in the second it's protected, thus making it impossible to create it in unwanted places. 
Am I still missing something here?

martyone: yup, the declaration is public, but the definition is private. Declaring the manager class as a friend to the tag would improve this further, replacing linker error with compiler error, but I do not try to overcomplicate access control at package scope - if it has private or internal in its name, it should be clear enough it shouldn't be used from outside.

vigejolla: Sorry I keep bugging you about this, it's just because I have this feeling that there is more here than meets (my) eye. What would happen if we removed PrivateConstructorTag completely? Would some std::whatever() function somewhere generate extra BuildEngines? Or is it really just to make it more difficult (for the human programmer) to create BuildEngines (and possibly bugs)?

martyone: It is there 1) to make it _impossible_ for the outside programmer to create build engines other way than via Sdk::createBuildEngine() and 2) to make the internal programmer stop and think before they try to create a build engine, because it needs some special care from the BuildEngineManager side https://github.com/sailfishos/sailfish-qtcreator/blob/jb45560/src/libs/sfdk/buildengine.cpp#L961

src/libs/sfdk/buildengine.cpp
```diff
@@ -391,11 +391,7 @@ void BuildEnginePrivate::enableUpdates()
 
     virtualMachine->setAutoConnectEnabled(true);
 
-    VirtualBoxManager::fetchVirtualMachineInfo(virtualMachine->name(),
-            VirtualBoxManager::NoExtraInfo, q, [=](const VirtualMachineInfo &info, bool ok) {
-        QTC_ASSERT(ok, return);
-        updateVmProperties(info);
-    });
+    updateVmProperties(q, [](bool ok) { Q_UNUSED(ok); });
```
vigejolla: Q_UNUSED does not need a semicolon -> removing it would get rid of a compiler warning

martyone: Fixed. Interesting, which compiler it is? Only with that I noticed I didn't use -Wall -Werror for sfdk lib (fixed), but even after enabling that, g++-8 doesn't issue a warning here.

vigejolla: clang

src/libs/sfdk/virtualmachine.cpp
```diff
@@ -191,8 +186,16 @@ void VirtualMachine::setMemorySizeMb(int memorySizeMb, const QObject *context,
...
-    return VirtualMachinePrivate::s_availableMemmorySizeMb;
+#ifdef Q_OS_WIN
+    MEMORYSTATUSEX status;
+    status.dwLength = sizeof(status);
+    GlobalMemoryStatusEx(&status);
+    return status.ullTotalPhys >> 20;
+#else
+    long pages = sysconf(_SC_PHYS_PAGES);
+    long page_size = sysconf(_SC_PAGE_SIZE);
+    return (pages * page_size) >> 20;
```
vigejolla: More warnings from clang: implicit conversion loses integer precision: 'long' to 'int'

I guess it would like you to make the cast explicit.

martyone: I was surprised it didn't complain about these, so I thought it was so clever that >>20 made it quiet :) Unfortunately it's because -Wconversion is not included in -Wall and it is not possible to simply enable it with Qt used from build tree, as it issues a lot of warning from Qt headers. `QMAKE_CXXFLAGS += -isystem $$[QT_INSTALL_HEADERS]` does not help because when Qt is used from build tree, the headers are still in the source tree and so not covered by QT_INSTALL_HEADERS.

src/libs/sfdk/virtualmachine.h
```diff
@@ -0,0 +1,211 @@
...
+    int vdiCapacityMb() const;
+    void setVdiCapacityMb(int vdiCapacityMb, const QObject *context,
+            const Functor<bool> &functor);
+
+    bool hasPortForwarding(quint16 hostPort, QString *ruleName = nullptr) const;
+    void addPortForwarding(const QString &ruleName, const QString &protocol,
+            quint16 hostPort, quint16 emulatorVmPort, const QObject *context,
+            const Functor<bool> &functor);
+    void removePortForwarding(const QString &ruleName, const QObject *context,
+            const Functor<bool> &functor);
```
vigejolla: It seems these three methods (hasPortForwarding, addPortForwarding and removePortForwarding) are not called anywhere

martyone: Used by the sensors emulation plugin developed by FRUCT - didn't we already talk about these? :)

vigejolla: Maybe we did. Where are the sources for that plugin? Or some interface specification? Or something to help me review this. You don't really expect me to just rubber stamp this code without checking that it does what is expected from it?

martyone: https://github.com/sailfishos/sailfish-qtcreator/pull/165

vigejolla: Thanks. On a closer look I found the original conversation on the .cpp side. And you also linked the fruct in that conversation... So please accept my apologies - I was clearly lazy. And/or been looking at this PR for too long, time to do something else for a change.

src/libs/sfdk/vboxvirtualmachine.cpp
```diff
@@ -0,0 +1,1008 @@
...
+            ruleName = QLatin1String(SSH_NATPF_RULE_NAME);
+            guestPort = 22;
+            break;
+        case VirtualMachinePrivate::WwwPort:
+            ruleName = QLatin1String(WWW_NATPF_RULE_NAME);
+            guestPort = 8080;
+            break;
+    }
+    const QString rule = QString::fromLatin1(NATPF_RULE_TEMPLATE)
+        .arg(ruleName).arg(port).arg(guestPort);
```
vigejolla: Now that you added -Werror, this no longer compiles. error: ‘guestPort’ may be used uninitialized in this function.

share/qtcreator/sfdk/modules/30-building-builtin.json
```diff
@@ -15,9 +15,9 @@
     [
         {
             "name": "engine",
-            "trSynopsis": "{start|stop|status}",
+            "trSynopsis": "{start|stop|status}\nexec [<command> [<arg>...]]",
             "trBrief": "Control the build engine",
-            "trDescription": "The 'start' and 'stop' commands allow to control the build engine virtual machine conveniently. Starting the build engine explicitly is usually not required - commands that need build engine running will ensure it is running. The 'status' command prints status information."
+            "trDescription": "The 'start' and 'stop' commands allow to control the build engine virtual machine conveniently. Starting the build engine explicitly is usually not required - commands that need build engine running will ensure it is running. The 'status' command prints status information.\n\nThe 'exec' command allows executing an arbitrary <command> inside the build engine. If no <command> is specified, an interactive shell is executed."
```
vigejolla: I think it could be mentioned that path mapping is performed on the output when running the exec command. It can be a bit surprising - e.g. try "sfdk engine exec echo /home/mersdk/share"

src/libs/sfdk/vboxvirtualmachine.cpp
martyone: Fixed

share/qtcreator/sfdk/modules/30-building-builtin.json
martyone: Checked that, it would need to be mentioned under most of the other commands as well, so I think the paragraph explaining path mapping under domain help should be enough.

src/libs/sfdk/vboxvirtualmachine.cpp
```diff
@@ -0,0 +1,1008 @@
...
+{
+    Q_Q(VBoxVirtualMachine);
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    qCDebug(vms) << "Setting reserved port forwarding" << which << "for" << q->uri().toString()
+        << "to" << port;
+
+    QString ruleName;
+    int guestPort{};
```
vigejolla: Technically, this works, but how is it better than "int guestPort = 0;", which would do the same thing and would also be clear to the reader?

martyone: it's the same as "QString ruleName;", but for primitive types you need to require the default construction explicitly and at the same time "int questPort();" would be treated as forward declaration of a function questPort. That's one of the reasons why the unified initialization was introduced with C++11. If I do not need a specific value, I prefer expressing it using default initialization, explicit where it is needed.

vigejolla: While I do understand your logic, I have a minor issue with the fact that you are using the uniform initialization to perform the default initialization to zero. Yes, I do understand that you are trying to express the fact that you don't really care about the value, but at the same time you are hiding the fact that it is being initialized to zero. Sure, not a big issue for anyone who can remember all of the C++ standard by heart.

src/plugins/mer/merplugin.h
```diff
@@ -57,15 +57,6 @@ private slots:
 
 private:
     QMap<QString, Sfdk::VirtualMachine *> m_stopList;
-
-#ifdef WITH_TESTS
-    void verifyTargets(const QString &vm, QStringList expectedKits, QStringList expectedToolChains, QStringList expectedQtVersion);
-    void testMerSshOutputParsers_data();
```
vigejolla: testMerSshOutputParsers() and corresponding _data() are still there in the .cpp files, thus failing the build.

martyone: fixed, these two were removed by mistake, so added back.

PR #222, opened 2019-10-28
Jb47430

etaishev:
Move OS brand name to variable, which configure by qmake variable. By default use Sailfish OS, if variable VENDOR_NAME is set use $${VENDOR_NAME} OS

### DIFF
--- a/src/plugins/mer/merconstants.h
+++ b/src/plugins/mer/merconstants.h
@@ -29,11 +29,19 @@
 namespace Mer {
 namespace Constants {
 
+#ifdef MER_BASE_OS_NAME
+const char MER_OS_NAME[] = QT_TRANSLATE_NOOP("Mer", MER_BASE_OS_NAME " OS");
+const char MER_SDK_NAME[] = QT_TRANSLATE_NOOP("Mer", MER_BASE_OS_NAME " SDK");
+#else
+const char MER_OS_NAME[] = "Sailfish OS";
+const char MER_SDK_NAME[] = "Sailfish SDK";
+#endif
+
 const char MER_QT[] = "QmakeProjectManager.QtVersion.Mer";
 const char MER_TOOLCHAIN_ID[] = "QmakeProjectManager.ToolChain.Mer";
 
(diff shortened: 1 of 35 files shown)


martyone:
Sadly this touches code that has been changed a lot under branch jb45560 which is about to be merged soon. You will need to rebase you changes on top of that :(

etaishev:
Okay, wait it.

vigejolla:
I think we should have a better name for the qmake variable than "VENDOR_NAME". As the contents of the variable is not the name of the vendor, but the base name of the OS.

etaishev:
Update on current master

martyone:
@etaishev I only returned now to the review and I found some issues that should be fixed, so I reverted it from master. Please open a new PR with the same content, I will add my comments there.

etaishev:
@martyone   https://github.com/sailfishos/sailfish-qtcreator/pull/226

src/plugins/mer/merconstants.h
```diff
@@ -29,11 +29,19 @@
...
 namespace Constants {
 
+#ifdef MER_BASE_OS_NAME
+const char MER_OS_NAME[] = QT_TRANSLATE_NOOP("Mer", MER_BASE_OS_NAME " OS");
+const char MER_SDK_NAME[] = QT_TRANSLATE_NOOP("Mer", MER_BASE_OS_NAME " SDK");
+#else
+const char MER_OS_NAME[] = "Sailfish OS";
+const char MER_SDK_NAME[] = "Sailfish SDK";
+#endif
+
```
martyone: This seems wrong - one branch is i18n-enabled, the other is not. Is is necessary to enable i18n for these at all?

This also makes me think, do we really need to add a custom mechanism to change the branding? Wouldn't it be more convenient to use i18n for that. I mean you can leave Sailfish OS/Sailfish SDK in the code and provide translations (easy to automate that) that say something else.

martyone: Of course this comment should be posted under the new PR...

PR #223, opened 2019-10-29
Docker based build engine

vigejolla:
Please note that only the last 6 commits belong to this PR

### DIFF
--- a/src/libs/sfdk/dockervirtualmachine.cpp
+++ b/src/libs/sfdk/dockervirtualmachine.cpp
@@ -1,0 +1,610 @@
+/****************************************************************************
+**
+** Copyright (C) 2019 Open Mobile Platform LLC.
+** Contact: http://jolla.com/
+**
+** This file is part of Qt Creator.
+**
+** GNU Lesser General Public License Usage
+** Alternatively, this file may be used under the terms of the GNU Lesser
+** General Public License version 2.1 as published by the Free Software
+** Foundation and appearing in the file LICENSE.LGPL included in the
+** packaging of this file.  Please review the following information to
+** ensure the GNU Lesser General Public License version 2.1 requirements
+** will be met: http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html.
+**
+** Other Usage
+**
+** Alternatively, this file may be used in accordance with the terms and
+** conditions contained in a signed written agreement between you and
+** The Qt Company.
+**
+****************************************************************************/
+
+#include "dockervirtualmachine_p.h"
+
+#include "asynchronous_p.h"
+#include "sdk_p.h"
+#include "sfdkconstants.h"
+#include "usersettings_p.h"
+
+#include <utils/fileutils.h>
+#include <utils/optional.h>
+#include <utils/qtcassert.h>
+
+#include <QDir>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QProcess>
+#include <QRegularExpression>
+
+using namespace Utils;
+
+namespace Sfdk {
+
+namespace {
+const char DOCKER[] = "docker";
+const char SAILFISH_SDK_SYSTEM_DOCKER[] = "SAILFISH_SDK_SYSTEM_DOCKER";
+const char RUN[] = "run";
+const char PRIVILEGED1[] = "--privileged";
+const char PRIVILEGED2[] = "--cap-drop=NET_ADMIN";
+const char STOP[] = "stop";
+const char COMMIT[] = "commit";
+const char CONTAINER_LIST[] = "ps";
+const char DETACHED[] = "-d";
+const char PUBLISH_PORT[] = "-p";
+const char VOLUME[] = "-v";
+const char CGROUP_MOUNT[] = "/sys/fs/cgroup:/sys/fs/cgroup:ro";
+const char TMPFS[] = "--tmpfs";
+const char TMPFS1[] = "/run:exec,mode=755";
+const char TMPFS2[] = "/run/lock";
+const char TMPFS3[] = "/tmp:exec";
+const char FILTER_PREFIX[] = "--filter=ancestor=";
+const char FORMAT_CONTAINER_LIST[] = "--format={{.Image}} {{.ID}}";
+const char COMMAND_INSPECT[] = "inspect";
+const char FORMAT_LABEL_LIST[] = "--format={{json .Config.Labels}}";
+const char IMAGE[] = "image";
+const char LIST[] = "ls";
+const char FORMAT_REPOSITORY[] = "--format={{.Repository}}";
+const char UNNAMED_IMAGE[] = "<none>";
+const char COMMAND_BUILD[] = "build";
+const char DOCKERFILE_PREFIX[] = "FROM ";
+const char LABEL_PREFIX[] = "--label=";
+const char TAG[] = "--tag";
+const quint16 GUESTSSHPORT = 22;
+const quint16 GUESTWWWPORT = 9292;
+} // namespace anonymous
+
+namespace {
+
+QString dockerPath()
+{
+    static QString path;
+    if (!path.isNull())
+        return path;
+
+    const QStringList searchPaths = QProcessEnvironment::systemEnvironment()
+        .value("PATH").split(QChar(':'));
+    for (const QString &searchPath : searchPaths) {
+        QDir dir(searchPath);
+        if (dir.exists(DOCKER)) {
+            path = dir.absoluteFilePath(DOCKER);
+            break;
+        }
+    }
+
+    QTC_ASSERT(!path.isEmpty(), return path);
+    return path;
+}
+
+class DockerRunner : public ProcessRunner
+{
+    Q_OBJECT
+
+public:
+    explicit DockerRunner(const QStringList &arguments, QObject *parent = 0)
+        : ProcessRunner(path(), arguments, parent)
+    {
+        process()->setProcessEnvironment(environment());
+    }
+
+private:
+    static QString path()
+    {
+        QString path = dockerPath();
+
+        if (!SdkPrivate::customDockerPath().isEmpty())
+            path = SdkPrivate::customDockerPath();
+
+        return path;
+    }
+
+    static QProcessEnvironment environment()
+    {
+        QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
+
+        if (!SdkPrivate::customDockerPath().isEmpty())
+            env.insert(SAILFISH_SDK_SYSTEM_DOCKER, dockerPath());
+
+        return env;
+    }
+};
+
+CommandQueue *commandQueue()
+{
+    return SdkPrivate::commandQueue();
+}
+
+} // namespace anonymous
+
+/*!
+ * \class DockerVirtualMachine
+ */
+
+DockerVirtualMachine::DockerVirtualMachine(const QString &name, QObject *parent)
+    : VirtualMachine(std::make_unique<DockerVirtualMachinePrivate>(this), staticType(), name, parent)
+{
+    Q_D(DockerVirtualMachine);
+    d->setDisplayType(staticDisplayType());
+}
+
+DockerVirtualMachine::~DockerVirtualMachine()
+{
+}
+
+bool DockerVirtualMachine::isAvailable()
+{
+    return !dockerPath().isEmpty();
+}
+
+QString DockerVirtualMachine::staticType()
+{
+    return QStringLiteral("Docker");
+}
+
+QString DockerVirtualMachine::staticDisplayType()
+{
+    return tr("Docker");
+}
+
+void DockerVirtualMachine::fetchRegisteredVirtualMachines(const QObject *context,
+        const Functor<const QStringList &, bool> &functor)
+{
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    QStringList arguments;
+    arguments.append(IMAGE);
+    arguments.append(LIST);
+    arguments.append(FORMAT_REPOSITORY);
+
+    auto runner = std::make_unique<DockerRunner>(arguments);
+    QObject::connect(runner.get(), &DockerRunner::done,
+            context, [=, runner = runner.get()](bool ok) {
+        if (!ok) {
+            functor({}, false);
+            return;
+        }
+        QStringList vms = DockerVirtualMachinePrivate::listedImages(QString::fromLocal8Bit(
+                runner->process()->readAllStandardOutput()));
+        functor(vms, true);
+    });
+    commandQueue()->enqueue(std::move(runner));
+}
+
+/*!
+ * \class DockerVirtualMachinePrivate
+ * \internal
+ */
+
+void DockerVirtualMachinePrivate::fetchInfo(VirtualMachineInfo::ExtraInfos extraInfo,
+        const QObject *context, const Functor<const VirtualMachineInfo &, bool> &functor) const
+{
+    Q_Q(const DockerVirtualMachine);
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    // TODO: fetch snapshot info if requested
+    Q_UNUSED(extraInfo)
+
+    QStringList arguments;
+    arguments.append(COMMAND_INSPECT);
+    arguments.append(q->name());
+    arguments.append(FORMAT_LABEL_LIST);
+
+    auto runner = std::make_unique<DockerRunner>(arguments);
+    QObject::connect(runner.get(), &DockerRunner::done, context,
+            [this, functor, process = runner->process()](bool ok) {
+        VirtualMachineInfo info;
+        if (ok)
+            info = virtualMachineInfoFromOutput(process->readAllStandardOutput().trimmed());
+
+        functor(info, ok);
+    });
+    commandQueue()->enqueue(std::move(runner));
+}
+
+void DockerVirtualMachinePrivate::start(const QObject *context, const Functor<bool> &functor)
+{
+    Q_Q(DockerVirtualMachine);
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+    QTC_ASSERT(cachedInfo().sshPort != 0, return);
+    QTC_ASSERT(cachedInfo().wwwPort != 0, return);
+    QTC_ASSERT(!cachedInfo().sharedSrc.isEmpty(), return);
+    QTC_ASSERT(!cachedInfo().sharedHome.isEmpty(), return);
+    QTC_ASSERT(!cachedInfo().sharedTargets.isEmpty(), return);
+    QTC_ASSERT(!cachedInfo().sharedConfig.isEmpty(), return);
+    QTC_ASSERT(!cachedInfo().sharedSsh.isEmpty(), return);
+
+    QStringList arguments;
+    arguments.append(RUN);
+    arguments.append(DETACHED);
+    arguments.append(PRIVILEGED1);
+    arguments.append(PRIVILEGED2);
+    arguments.append(VOLUME);
+    arguments.append(CGROUP_MOUNT);
+
+    auto addTmpfs = [&arguments](const QString &path) {
+        arguments.append(TMPFS);
+        arguments.append(path);
+    };
+    addTmpfs(TMPFS1);
+    addTmpfs(TMPFS2);
+    addTmpfs(TMPFS3);
+
+    auto forwardPort = [&arguments](quint16 guestPort, quint16 hostPort) {
+        arguments.append(PUBLISH_PORT);
+        arguments.append(QString::number(hostPort) + ":" + QString::number(guestPort));
+    };
+    forwardPort(GUESTSSHPORT, cachedInfo().sshPort);
+    forwardPort(GUESTWWWPORT, cachedInfo().wwwPort);
+
+    auto sharePath = [&arguments](const QString &guestPath, const QString &hostPath) {
+        arguments.append(VOLUME);
+        arguments.append(hostPath + ":" + guestPath);
+    };
+    sharePath(Constants::BUILD_ENGINE_SHARED_HOME_MOUNT_POINT, cachedInfo().sharedHome);
+    sharePath(Constants::BUILD_ENGINE_SHARED_SRC_MOUNT_POINT, cachedInfo().sharedSrc);
+    sharePath(Constants::BUILD_ENGINE_SHARED_TARGET_MOUNT_POINT, cachedInfo().sharedTargets);
+    sharePath(Constants::BUILD_ENGINE_SHARED_CONFIG_MOUNT_POINT, cachedInfo().sharedConfig);
+    sharePath(Constants::BUILD_ENGINE_SHARED_SSH_MOUNT_POINT, cachedInfo().sharedSsh);
+
+    arguments.append(q->name());
+
+    auto runner = std::make_unique<DockerRunner>(arguments);
+    QObject::connect(runner.get(), &DockerRunner::done, q,
+            [this, context, functor, process = runner->process()](bool ok) {
+        if (ok)
+            containerId = QString::fromLocal8Bit(process->readAllStandardOutput().trimmed());
+        if (context)
+            functor(ok);
+    });
+    commandQueue()->enqueue(std::move(runner));
+}
+
+void DockerVirtualMachinePrivate::stop(const QObject *context, const Functor<bool> &functor)
+{
+    Q_Q(DockerVirtualMachine);
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    const QPointer<const QObject> context_{context};
+
+    auto enqueue = [=](const QStringList &arguments, CommandQueue::BatchId batch) {
+        auto runner = std::make_unique<DockerRunner>(arguments);
+        QObject::connect(runner.get(), &DockerRunner::failure, Sdk::instance(), [=]() {
+            commandQueue()->cancelBatch(batch);
+            callIf(context_, functor, false);
+        });
+        DockerRunner *runner_ = runner.get();
+        commandQueue()->enqueue(std::move(runner));
+        return runner_;
+    };
+
+    const CommandQueue::BatchId batch = commandQueue()->beginBatch();
+
+    QStringList stopArguments;
+    stopArguments.append(STOP);
+    stopArguments.append(containerId);
+
+    enqueue(stopArguments, batch);
+
+    QStringList commitArguments;
+    commitArguments.append(COMMIT);
+    commitArguments.append(containerId);
+    commitArguments.append(q->name());
+
+    enqueue(commitArguments, batch);
+
+    commandQueue()->enqueueCheckPoint(context, [=]() { functor(true); });
+    commandQueue()->endBatch();
+}
+
+void DockerVirtualMachinePrivate::probe(const QObject *context,
+        const Functor<BasicState, bool> &functor) const
+{
+    Q_Q(const DockerVirtualMachine);
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    const QPointer<const QObject> context_{context};
+
+    q->fetchRegisteredVirtualMachines(q, [=](const QStringList &registeredVms, bool ok) {
+        if (!ok) {
+            if (context_)
+                functor({}, false);
+            return;
+        }
+
+        if (!registeredVms.contains(q->name())) {
+            if (context_)
+                functor({}, true);
+            return;
+        }
+
+        QStringList arguments;
+        arguments.append(CONTAINER_LIST);
+        arguments.append(QString(FILTER_PREFIX) + q->name());
+        arguments.append(FORMAT_CONTAINER_LIST);
+
+        auto runner = std::make_unique<DockerRunner>(arguments);
+        QObject::connect(runner->process(),
+                QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished),
+                q,
+                [=, runner = runner.get()](int exitCode, QProcess::ExitStatus exitStatus) {
+            if (exitStatus != QProcess::NormalExit || exitCode != 0) {
+                if (context_)
+                    functor({}, false);
+                return;
+            }
+
+            VirtualMachinePrivate::BasicState state = VirtualMachinePrivate::Existing;
+            const QStringList runningContainers =
+                QString::fromLocal8Bit(runner->process()->readAllStandardOutput())
+                    .split('\n', QString::SkipEmptyParts);
+            for (const QString& runningContainer : runningContainers) {
+                QStringList nameAndId = runningContainer.split(' ', QString::SkipEmptyParts);
+                QTC_ASSERT(nameAndId.count() == 2 && nameAndId.first() == q->name(), continue);
+                state |= VirtualMachinePrivate::Running | VirtualMachinePrivate::Headless;
+                if (containerId.isEmpty())
+                    containerId = nameAndId.last();
+            }
+            if (context_)
+                functor(state, true);
+        });
+        commandQueue()->enqueue(std::move(runner));
+    });
+}
+
+void DockerVirtualMachinePrivate::setVideoMode(const QSize &size, int depth,
+        const QString &deviceModelName, Qt::Orientation orientation, int scaleDownFactor,
+        const QObject *context, const Functor<bool> &functor)
+{
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    Q_UNUSED(size)
+    Q_UNUSED(depth)
+    Q_UNUSED(deviceModelName)
+    Q_UNUSED(orientation)
+    Q_UNUSED(scaleDownFactor)
+    Q_UNUSED(context)
+    Q_UNUSED(functor)
+    QTC_CHECK(false);
+}
+
+void DockerVirtualMachinePrivate::doSetMemorySizeMb(int memorySizeMb, const QObject *context,
+        const Functor<bool> &functor)
+{
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    Q_UNUSED(memorySizeMb)
+    Q_UNUSED(context)
+    Q_UNUSED(functor)
+    QTC_CHECK(false);
+}
+
+void DockerVirtualMachinePrivate::doSetCpuCount(int cpuCount, const QObject *context,
+        const Functor<bool> &functor)
+{
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    Q_UNUSED(cpuCount)
+    Q_UNUSED(context)
+    Q_UNUSED(functor)
+    QTC_CHECK(false);
+}
+
+void DockerVirtualMachinePrivate::doSetStorageSizeMb(int storageSizeMb, const QObject *context,
+        const Functor<bool> &functor)
+{
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    Q_UNUSED(storageSizeMb)
+    Q_UNUSED(context)
+    Q_UNUSED(functor)
+    QTC_CHECK(false);
+}
+
+void DockerVirtualMachinePrivate::doSetSharedPath(SharedPath which, const FileName &path,
+        const QObject *context, const Functor<bool> &functor)
+{
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    buildWithLabel(labelFor(which), path.toString(), context, functor);
+}
+
+void DockerVirtualMachinePrivate::doAddPortForwarding(const QString &ruleName,
+        const QString &protocol, quint16 hostPort, quint16 emulatorVmPort,
+        const QObject *context, const Functor<bool> &functor)
+{
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    Q_UNUSED(ruleName)
+    Q_UNUSED(protocol)
+    Q_UNUSED(hostPort)
+    Q_UNUSED(emulatorVmPort)
+    Q_UNUSED(context)
+    Q_UNUSED(functor)
+    QTC_CHECK(false);
+}
+
+void DockerVirtualMachinePrivate::doRemovePortForwarding(const QString &ruleName,
+        const QObject *context, const Functor<bool> &functor)
+{
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    Q_UNUSED(ruleName)
+    Q_UNUSED(context)
+    Q_UNUSED(functor)
+    QTC_CHECK(false);
+}
+
+void DockerVirtualMachinePrivate::doSetReservedPortForwarding(ReservedPort which, quint16 port,
+        const QObject *context, const Functor<bool> &functor)
+{
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    buildWithLabel(labelFor(which), QString::number(port), context, functor);
+}
+
+void DockerVirtualMachinePrivate::doSetReservedPortListForwarding(ReservedPortList which,
+        const QList<Utils::Port> &ports, const QObject *context,
+        const Functor<const QMap<QString, quint16> &, bool> &functor)
+{
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    Q_UNUSED(which)
+    Q_UNUSED(ports)
+    Q_UNUSED(context)
+    Q_UNUSED(functor)
+    QTC_CHECK(false);
+}
+
+void DockerVirtualMachinePrivate::doRestoreSnapshot(const QString &snapshotName, const QObject *context,
+        const Functor<bool> &functor)
+{
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    Q_UNUSED(snapshotName)
+    Q_UNUSED(context)
+    Q_UNUSED(functor)
+    QTC_CHECK(false);
+}
+
+QStringList DockerVirtualMachinePrivate::listedImages(const QString &output)
+{
+    QStringList images;
+    const QStringList lines = output.split(QRegularExpression("[\r\n]"), QString::SkipEmptyParts);
+    for (const QString &line : lines) {
+        if (line != UNNAMED_IMAGE)
+            images.append(line);
+    }
+    return images;
+}
+
+VirtualMachineInfo DockerVirtualMachinePrivate::virtualMachineInfoFromOutput(const QByteArray &output)
+{
+    VirtualMachineInfo info;
+
+    QJsonDocument document = QJsonDocument::fromJson(output);
+    QTC_ASSERT(!document.isNull(), return info);
+
+    info.memorySizeMb = VirtualMachine::availableMemorySizeMb();
+    info.cpuCount = VirtualMachine::availableCpuCount();
+
+    QJsonObject labels = document.object();
+
+    auto parsePort = [=](VirtualMachinePrivate::ReservedPort which, quint16 *port) {
+        const QString key = labelFor(which);
+        QTC_ASSERT(labels.contains(key), return);
+        *port = labels.value(key).toString().toUShort();
+    };
+
+    parsePort(VirtualMachinePrivate::SshPort, &info.sshPort);
+    parsePort(VirtualMachinePrivate::WwwPort, &info.wwwPort);
+
+    auto parsePath = [=](VirtualMachinePrivate::SharedPath which, QString *path) {
+        const QString key = labelFor(which);
+        QTC_ASSERT(labels.contains(key), return);
+        *path = labels.value(key).toString();
+    };
+
+    parsePath(VirtualMachinePrivate::SharedSrc, &info.sharedSrc);
+    parsePath(VirtualMachinePrivate::SharedSsh, &info.sharedSsh);
+    parsePath(VirtualMachinePrivate::SharedHome, &info.sharedHome);
+    parsePath(VirtualMachinePrivate::SharedConfig, &info.sharedConfig);
+    parsePath(VirtualMachinePrivate::SharedTargets, &info.sharedTargets);
+
+    return info;
+}
+
+void DockerVirtualMachinePrivate::buildWithLabel(const QString& key, const QString& value,
+        const QObject *context, const Functor<bool> &functor)
+{
+    Q_Q(DockerVirtualMachine);
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    QStringList arguments;
+    arguments.append(COMMAND_BUILD);
+    arguments.append(LABEL_PREFIX + key + "=" + value);
+    arguments.append(TAG);
+    arguments.append(q->name());
+    arguments.append("-");
+
+    auto runner = std::make_unique<DockerRunner>(arguments);
+
+    QObject::connect(runner->process(), &QProcess::started, q, [q, process = runner->process()]() {
+        QString dockerFile(DOCKERFILE_PREFIX + q->name());
+        process->write(dockerFile.toUtf8());
+        process->closeWriteChannel();
+    });
+
+    QObject::connect(runner.get(), &DockerRunner::done, context, functor);
+    commandQueue()->enqueue(std::move(runner));
+}
+
+QString DockerVirtualMachinePrivate::labelFor(VirtualMachinePrivate::SharedPath which)
+{
+    switch (which) {
+    case SharedSrc:
+        return Constants::BUILD_ENGINE_SHARED_SRC;
+    case SharedHome:
+        return Constants::BUILD_ENGINE_SHARED_HOME;
+    case SharedConfig:
+        return Constants::BUILD_ENGINE_SHARED_CONFIG;
+    case SharedSsh:
+        return Constants::BUILD_ENGINE_SHARED_SSH;
+    case SharedTargets:
+        return Constants::BUILD_ENGINE_SHARED_TARGET;
+    }
+    Q_ASSERT(false);
+    return {};
+}
+
+QString DockerVirtualMachinePrivate::labelFor(VirtualMachinePrivate::ReservedPort which)
+{
+    switch (which) {
+    case SshPort:
+        return Constants::BUILD_ENGINE_SSH_PORT;
+    case WwwPort:
+        return Constants::BUILD_ENGINE_WWW_PORT;
+    }
+    Q_ASSERT(false);
+    return {};
+}
+
+} // namespace Sfdk
+
+#include "dockervirtualmachine.moc"
--- a/src/libs/sfdk/dockervirtualmachine_p.h
+++ b/src/libs/sfdk/dockervirtualmachine_p.h
@@ -1,0 +1,106 @@
+/****************************************************************************
+**
+** Copyright (C) 2019 Open Mobile Platform LLC.
+** Contact: http://jolla.com/
+**
+** This file is part of Qt Creator.
+**
+** GNU Lesser General Public License Usage
+** Alternatively, this file may be used under the terms of the GNU Lesser
+** General Public License version 2.1 as published by the Free Software
+** Foundation and appearing in the file LICENSE.LGPL included in the
+** packaging of this file.  Please review the following information to
+** ensure the GNU Lesser General Public License version 2.1 requirements
+** will be met: http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html.
+**
+** Other Usage
+**
+** Alternatively, this file may be used in accordance with the terms and
+** conditions contained in a signed written agreement between you and
+** The Qt Company.
+**
+****************************************************************************/
+
+#pragma once
+
+#include "virtualmachine_p.h"
+
+namespace Sfdk {
+
+class DockerVirtualMachinePrivate;
+class DockerVirtualMachine : public VirtualMachine
+{
+    Q_OBJECT
+
+public:
+    explicit DockerVirtualMachine(const QString &name, QObject *parent = nullptr);
+    ~DockerVirtualMachine() override;
+
+    static bool isAvailable();
+    static QString staticType();
+    static QString staticDisplayType();
+    static void fetchRegisteredVirtualMachines(const QObject *context,
+            const Functor<const QStringList &, bool> &functor);
+
+private:
+    Q_DISABLE_COPY(DockerVirtualMachine)
+    Q_DECLARE_PRIVATE(DockerVirtualMachine)
+};
+
+class DockerVirtualMachinePrivate : public VirtualMachinePrivate
+{
+    Q_DECLARE_PUBLIC(DockerVirtualMachine)
+
+public:
+    using VirtualMachinePrivate::VirtualMachinePrivate;
+
+    void fetchInfo(VirtualMachineInfo::ExtraInfos extraInfo, const QObject *context,
+            const Functor<const VirtualMachineInfo &, bool> &functor) const override;
+
+    void start(const QObject *context, const Functor<bool> &functor) override;
+    void stop(const QObject *context, const Functor<bool> &functor) override;
+    void probe(const QObject *context,
+            const Functor<BasicState, bool> &functor) const override;
+
+    void setVideoMode(const QSize &size, int depth, const QString &deviceModelName,
+            Qt::Orientation orientation, int scaleDownFactor, const QObject *context,
+            const Functor<bool> &functor) override;
+
+protected:
+    void doSetMemorySizeMb(int memorySizeMb, const QObject *context,
+            const Functor<bool> &functor) override;
+    void doSetCpuCount(int cpuCount, const QObject *context,
+            const Functor<bool> &functor) override;
+    void doSetStorageSizeMb(int storageSizeMb, const QObject *context,
+            const Functor<bool> &functor) override;
+
+    void doSetSharedPath(SharedPath which, const Utils::FileName &path, const QObject *context,
+            const Functor<bool> &functor) override;
+
+    void doAddPortForwarding(const QString &ruleName,
+            const QString &protocol, quint16 hostPort, quint16 emulatorVmPort,
+            const QObject *context, const Functor<bool> &functor) override;
+    void doRemovePortForwarding(const QString &ruleName,
+            const QObject *context, const Functor<bool> &functor) override;
+    void doSetReservedPortForwarding(ReservedPort which, quint16 port,
+            const QObject *context, const Functor<bool> &functor) override;
+    void doSetReservedPortListForwarding(ReservedPortList which, const QList<Utils::Port> &ports,
+            const QObject *context, const Functor<const QMap<QString, quint16> &, bool> &functor)
+        override;
+
+    void doRestoreSnapshot(const QString &snapshotName, const QObject *context,
+            const Functor<bool> &functor) override;
+
+private:
+    static QStringList listedImages(const QString &output);
+    static VirtualMachineInfo virtualMachineInfoFromOutput(const QByteArray &output);
+
+    void buildWithLabel(const QString& key, const QString& value,
+                        const QObject *context, const Functor<bool> &functor);
+    static QString labelFor(SharedPath which);
+    static QString labelFor(ReservedPort which);
+
+    mutable QString containerId;
+};
+
+} // namespace Sfdk
--- a/src/libs/sfdk/sfdkconstants.h
+++ b/src/libs/sfdk/sfdkconstants.h
@@ -60,9 +60,14 @@ const char BUILD_ENGINE_WWW_PROXY_EXCLUDES[] = "WwwProxyExcludes";
 const char BUILD_ENGINE_HEADLESS[] = "Headless";
 const char BUILD_ENGINE_TARGETS_COUNT_KEY[] = "BuildTargets.Count";
 const char BUILD_ENGINE_TARGET_DATA_KEY_PREFIX[] = "BuildTarget.";
-
+const char BUILD_ENGINE_HOST_TARGET_PATH_POSTFIX[] = "/mersdk/targets";
+const char BUILD_ENGINE_HOST_SSH_PATH_POSTFIX[] = "/mersdk/ssh";
+const char BUILD_ENGINE_HOST_CONFIG_PATH_POSTFIX[] = "/vmshare";
+const char BUILD_ENGINE_SHARED_CONFIG_MOUNT_POINT[] = "/etc/mersdk/share";
 const char BUILD_ENGINE_SHARED_HOME_MOUNT_POINT[] = "/home/mersdk/share";
 const char BUILD_ENGINE_SHARED_SRC_MOUNT_POINT[] = "/home/src1";
+const char BUILD_ENGINE_SHARED_TARGET_MOUNT_POINT[] = "/host_targets";
+const char BUILD_ENGINE_SHARED_SSH_MOUNT_POINT[] = "/etc/ssh/authorized_keys";
 
 const char BUILD_TARGET_NAME[] = "Name";
 const char BUILD_TARGET_GCC_DUMP_MACHINE[] = "GccDumpMachine";
--- a/src/libs/sfdk/sdk.cpp
+++ b/src/libs/sfdk/sdk.cpp
@@ -80,7 +80,8 @@ Sdk::Sdk(Options options)
     d->commandQueue_ = std::make_unique<CommandQueue>(this);
 
     d->virtualMachineFactory = std::make_unique<VirtualMachineFactory>(this);
-    d->virtualMachineFactory->registerType<VBoxVirtualMachine>();
+    if (VBoxVirtualMachine::isAvailable())
+        d->virtualMachineFactory->registerType<VBoxVirtualMachine>();
     if (DockerVirtualMachine::isAvailable())
         d->virtualMachineFactory->registerType<DockerVirtualMachine>();
 
--- a/src/libs/sfdk/sdk.cpp
+++ b/src/libs/sfdk/sdk.cpp
@@ -26,6 +26,7 @@
 #include "buildengine_p.h"
 #include "device_p.h"
 #include "emulator_p.h"
+#include "dockervirtualmachine_p.h"
 #include "sfdkconstants.h"
 #include "sfdk_version_p.h"
 #include "usersettings_p.h"
@@ -50,6 +51,7 @@ namespace {
 const char SDK_MAINTENANCE_TOOL_DATA_FILE[] = "SDKMaintenanceTool.dat";
 const char GENERAL_SETTINGS_GROUP[] = "General";
 const char VBOXMANAGE_PATH[] = "VBoxManagePath";
+const char DOCKER_PATH[] = "DockerPath";
 }
 
 /*!
@@ -79,6 +81,8 @@ Sdk::Sdk(Options options)
 
     d->virtualMachineFactory = std::make_unique<VirtualMachineFactory>(this);
     d->virtualMachineFactory->registerType<VBoxVirtualMachine>();
+    if (DockerVirtualMachine::isAvailable())
+        d->virtualMachineFactory->registerType<DockerVirtualMachine>();
 
     d->buildEngineManager = std::make_unique<BuildEngineManager>(this);
 
@@ -413,6 +417,7 @@ void SdkPrivate::readGeneralSettings()
 
     settings.beginGroup(GENERAL_SETTINGS_GROUP);
     customVBoxManagePath_ = settings.value(VBOXMANAGE_PATH).toString();
+    customDockerPath_ = settings.value(DOCKER_PATH).toString();
     settings.endGroup();
 }
 
--- a/src/libs/sfdk/sdk_p.h
+++ b/src/libs/sfdk/sdk_p.h
@@ -70,6 +70,7 @@ public:
     static Utils::FileName cacheLocation();
 
     static QString customVBoxManagePath() { return instance()->customVBoxManagePath_; }
+    static QString customDockerPath() { return instance()->customDockerPath_; }
 
 signals:
     void enableUpdatesRequested();
@@ -88,6 +89,7 @@ private:
     std::unique_ptr<EmulatorManager> emulatorManager;
     std::unique_ptr<DeviceManager> deviceManager;
     QString customVBoxManagePath_;
+    QString customDockerPath_;
 };
 
 } // namespace Sfdk
--- a/src/libs/sfdk/sfdk.pro
+++ b/src/libs/sfdk/sfdk.pro
@@ -11,6 +11,7 @@ SOURCES += \
     asynchronous.cpp \
     buildengine.cpp \
     device.cpp \
+    dockervirtualmachine.cpp \
     emulator.cpp \
     sfdkglobal.cpp \
     sdk.cpp \
@@ -28,6 +29,7 @@ HEADERS += \
     buildengine_p.h \
     device.h \
     device_p.h \
+    dockervirtualmachine_p.h \
     emulator.h \
     emulator_p.h \
     sfdkconstants.h \
--- a/src/libs/sfdk/vboxvirtualmachine.cpp
+++ b/src/libs/sfdk/vboxvirtualmachine.cpp
@@ -212,6 +212,11 @@ VBoxVirtualMachine::~VBoxVirtualMachine()
 {
 }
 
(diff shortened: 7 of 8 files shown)


vigejolla:
rebased on top of master

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+    arguments[0] = QLatin1String(COMMIT);
+    arguments.append(q->name());
+
+    enqueue(arguments, batch);
+
+    SdkPrivate::commandQueue()->enqueueCheckPoint(context,  [=]() { functor(true); });
+    SdkPrivate::commandQueue()->endBatch();
+}
+
+void DockerContainerPrivate::probe(const QObject *context, const Functor<VirtualMachinePrivate::BasicState, bool> &functor) const
```
martyone: Long line

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+{
+    Q_Q(const DockerContainer);
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    DockerManager::fetchRegisteredVirtualMachines(q,
+        [=](const QStringList &registeredVms, bool ok) {
```
martyone: I would join that with the previous line and decrease indent of the body

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+            if (!ok) {
+                VirtualMachinePrivate::BasicState state;
```
martyone: It would be more understandable if defined just once at outer block scope. From that I would understand that this is the outcome that is to be computed.

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+                functor(state, false);
+            } else {
+                if (!registeredVms.contains(q->name())) {
+                    VirtualMachinePrivate::BasicState state;
+                    functor(state, true);
```
martyone: Missing return? If so, then it would be better to instead make it an `else if` block in the upper level if-else chain.

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+                }
+                QStringList arguments;
+                arguments.append(QLatin1String(CONTAINER_LIST));
+                arguments.append(QLatin1String(FILTER_PREFIX)+q->name());
```
martyone: spaces around operator

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+                arguments.append(QLatin1String(FORMAT));
+                auto runner = std::make_unique<ProcessRunner>(DockerManager::dockerPath(), arguments);
+                QObject::connect(runner->process(), QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished), context,
+                        [=, runner = runner.get()](int exitCode, QProcess::ExitStatus exitStatus) {
+                    VirtualMachinePrivate::BasicState state = VirtualMachinePrivate::Existing;
+                    if (exitStatus != QProcess::NormalExit) {
+                        functor(state, false);
+                        return;
+                    }
+                    if (exitCode != 0) {
```
martyone: Handle those two with single `if`?

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+                        functor(state, false);
+                        return;
+                    }
+
+                    QString runningContainers = QString::fromLocal8Bit(runner->process()->readAllStandardOutput());
+                    bool isRunning = runningContainers.contains(q->name());
```
martyone: Restrict to whole-line match or so?

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+                    if (isRunning) {
+                        state |= (VirtualMachinePrivate::Running | VirtualMachinePrivate::Headless);
```
martyone: Ugly superfluous parentheses :) :)

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+                        if (q->m_containerId.isEmpty())
+                            q->m_containerId = runningContainers.split(' ')[1].trimmed();
```
martyone: One never knows.. check for required capacity?

Second, not that it should make any harm to performance here, but I guess `operator[]` detaches unnecessarily even in this context - I would simply avoid its use for reading.

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+                    }
+                    functor(state, true);
+                });
+                SdkPrivate::commandQueue()->enqueue(std::move(runner));
+
```
martyone: Extra blank line

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+    Q_UNUSED(context)
+    Q_UNUSED(functor)
+    Q_ASSERT(false);
+}
+
+void DockerContainerPrivate::doSetMemorySizeMb(int memorySizeMb, const QObject *context, const Functor<bool> &functor)
+{
+    Q_UNUSED(memorySizeMb)
+    Q_UNUSED(context)
+    Q_UNUSED(functor)
```
martyone: assert false? similar in other unimplemented functions

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+                                          const Functor<bool> &functor)
+{
+    Q_UNUSED(size)
+    Q_UNUSED(depth)
+    Q_UNUSED(deviceModelName)
+    Q_UNUSED(orientation)
+    Q_UNUSED(scaleDownFactor)
+    Q_UNUSED(context)
+    Q_UNUSED(functor)
+    Q_ASSERT(false);
```
martyone: QTC_CHECK would be more convenient.

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+}
+
+void DockerContainerPrivate::doSetStorageSizeMb(int storageSizeMb, const QObject *context, const Functor<bool> &functor)
+{
+    Q_UNUSED(storageSizeMb)
+    Q_UNUSED(context)
+    Q_UNUSED(functor)
+}
+
+void DockerContainerPrivate::doSetSharedPath(VirtualMachinePrivate::SharedPath which, const Utils::FileName &path, const QObject *context, const Functor<bool> &functor)
```
martyone: Long line, more occurrences nearby

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+{
+    QPair<QString,QString>& paths = m_sharedPaths[which];
```
martyone: Space

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+    paths.first = path.toString();
+    SdkPrivate::commandQueue()->enqueueCheckPoint(context, [=]() { functor(true); });
```
martyone: Shouldn't we ensure it is stored persistently at this point? More similar cases.

src/libs/sfdk/dockercontainer.h
```diff
@@ -0,0 +1,54 @@
...
+#ifndef DOCKERCONTAINER_H
+#define DOCKERCONTAINER_H
+
+#include "virtualmachine.h"
+
+namespace Sfdk {
+
+class DockerContainerPrivate;
+
+class SFDK_EXPORT DockerContainer : public VirtualMachine
```
martyone: The name DockerVirtualMachine should be used

src/libs/sfdk/dockercontainer.h
```diff
@@ -0,0 +1,54 @@
...
+{
+    Q_OBJECT
+
+public:
+    explicit DockerContainer(const QString& name, QObject* parent = nullptr);
```
martyone: spaces around operators

src/libs/sfdk/dockercontainer.h
```diff
@@ -0,0 +1,54 @@
+/****************************************************************************
```
martyone: The content of this file should be moved to the *_p.h file - concrete VirtualMachine implementations are not public.

src/libs/sfdk/dockercontainer_p.h
```diff
@@ -0,0 +1,83 @@
...
+**
+** Other Usage
+**
+** Alternatively, this file may be used in accordance with the terms and
+** conditions contained in a signed written agreement between you and
+** The Qt Company.
+**
+****************************************************************************/
+#ifndef DOCKERCONTAINER_P_H
+#define DOCKERCONTAINER_P_H
```
martyone: `#pragma once`, missing blank line

src/libs/sfdk/dockercontainer_p.h
```diff
@@ -0,0 +1,83 @@
...
+    DockerContainerPrivate(VirtualMachine* virtualMachine);
+    virtual ~DockerContainerPrivate();
+
+    void fetchInfo(VirtualMachineInfo::ExtraInfos extraInfo, const QObject *context,
+                   const Functor<const VirtualMachineInfo &, bool> &functor) const;
+
+    void start(const QObject *context, const Functor<bool> &functor);
+    void stop(const QObject *context, const Functor<bool> &functor);
+    void probe(const QObject *context,
+               const Functor<BasicState, bool> &functor) const;
```
martyone: Missing `override` keywords

src/libs/sfdk/dockercontainer_p.h
```diff
@@ -0,0 +1,83 @@
...
+
+namespace Sfdk {
+
+class DockerContainerPrivate : public VirtualMachinePrivate
+{
+    Q_DECLARE_PUBLIC(DockerContainer)
+
+public:
+    DockerContainerPrivate(VirtualMachine* virtualMachine);
+    virtual ~DockerContainerPrivate();
```
martyone: Empty body - should be dropped.

src/libs/sfdk/dockercontainer_p.h
```diff
@@ -0,0 +1,83 @@
...
+#include "dockercontainer.h"
+
+namespace Sfdk {
+
+class DockerContainerPrivate : public VirtualMachinePrivate
+{
+    Q_DECLARE_PUBLIC(DockerContainer)
+
+public:
+    DockerContainerPrivate(VirtualMachine* virtualMachine);
```
martyone: Initialization should be done in public class' ctor instead. Private ctor can be inherited then:

```
using VirtualMachinePrivate::VirtualMachinePrivate;
```

src/libs/sfdk/dockercontainer_p.h
```diff
@@ -0,0 +1,83 @@
...
+            const QObject *context, const Functor<bool> &functor) override;
+    void doSetReservedPortListForwarding(ReservedPortList which, const QList<Utils::Port> &ports,
+            const QObject *context, const Functor<const QMap<QString, quint16> &, bool> &functor)
+        override;
+
+    void doRestoreSnapshot(const QString &snapshotName, const QObject *context,
+        const Functor<bool> &functor) override;
+
+private:
+
```
martyone: Extra blank line

src/libs/sfdk/dockercontainer_p.h
```diff
@@ -0,0 +1,83 @@
...
+    QMap<ReservedPort, QPair<quint16, quint16>> m_reservedPortMappings;
+    QMap<SharedPath,QPair<QString,QString>> m_sharedPaths;
```
martyone: Spaces

src/libs/sfdk/dockercontainer_p.h
```diff
@@ -0,0 +1,83 @@
...
+};
+
+}
```
martyone: `// namespace Sfdk`

src/libs/sfdk/dockermanager.cpp
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+    return instance()->m_dockerPath;
+}
+
+void DockerManager::useImage(const QString &image)
+{
+    if (--(instance()->m_usedImages[image]) == 0) {
+        instance()->m_usedImages.remove(image);
+    }
```
martyone: This is now enforced by VirtualMachineFactory and can be removed here. The other static functions should be moved to the DockerVirtualMachinePrivate class in a fashion similar to how it is done with VBoxVirtualMachinePrivate class.

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+** will be met: http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html.
+**
+** Other Usage
+**
+** Alternatively, this file may be used in accordance with the terms and
+** conditions contained in a signed written agreement between you and
+** The Qt Company.
+**
+****************************************************************************/
+#include "dockercontainer.h"
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+#include "dockercontainer_p.h"
+
+#include "asynchronous_p.h"
+#include "dockermanager.h"
+#include "sdk_p.h"
+
+#include <QDir>
+#include <QProcess>
+#include <utils/fileutils.h>
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+namespace{
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+    const quint16 HOSTSSHPORT = 2221;
+    const quint16 GUESTWWWPORT = 9292;
+    const quint16 HOSTWWWPORT = 8080;
+    const char GUESTTARGETPATH[] = "/host_targets";
+    const char HOSTTARGETPATHPOSTFIX[] = "/mersdk/targets";
+    const char GUESTSSHPATH[] = "/etc/ssh/authorized_keys";
+    const char HOSTSSHPATHPOSTFIX[] = "/mersdk/ssh";
+    const char HOSTCONFIGPATHPOSTFIX[] = "/vmshare";
+    const char GUESTCONFIGPATH[] = "/etc/mersdk/share";
+    const char GUESTSHAREDHOME[] = "/home/mersdk/share";
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+}
+
+namespace Sfdk {
+
+
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+    auto info = std::make_shared<VirtualMachineInfo>();
+
+    info->sshPort=m_reservedPortMappings[VirtualMachinePrivate::SshPort].first;
+    info->wwwPort=m_reservedPortMappings[VirtualMachinePrivate::WwwPort].first;
+    info->sharedSsh=m_sharedPaths[VirtualMachinePrivate::SharedSsh].first;
+    info->sharedTargets=m_sharedPaths[VirtualMachinePrivate::SharedTargets].first;
+    info->sharedConfig=m_sharedPaths[VirtualMachinePrivate::SharedConfig].first;
+    info->sharedHome=m_sharedPaths[VirtualMachinePrivate::SharedHome].first;
+    info->memorySizeMb=VirtualMachine::availableMemorySizeMb();
+    info->cpuCount=VirtualMachine::availableCpuCount();
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+    m_sharedPaths[VirtualMachinePrivate::SharedConfig] = QPair<QString,QString>(configDir, QLatin1String(GUESTCONFIGPATH));
+    m_sharedPaths[VirtualMachinePrivate::SharedHome] = QPair<QString,QString>(QDir::homePath(), QLatin1String(GUESTSHAREDHOME));
+}
+
+DockerContainerPrivate::~DockerContainerPrivate()
+{
+
+}
+
+void DockerContainerPrivate::fetchInfo(VirtualMachineInfo::ExtraInfos extraInfo, const QObject *context, const Functor<const VirtualMachineInfo &, bool> &functor) const
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+    QDir installationDir = QCoreApplication::applicationDirPath();
+    installationDir.cdUp();
+
+    QString targetDir(installationDir.absolutePath() + HOSTTARGETPATHPOSTFIX);
+    m_sharedPaths[VirtualMachinePrivate::SharedTargets] = QPair<QString,QString>(targetDir, QLatin1String(GUESTTARGETPATH));
+    QString sshDir(installationDir.absolutePath() + HOSTSSHPATHPOSTFIX);
+    m_sharedPaths[VirtualMachinePrivate::SharedSsh] = QPair<QString,QString>(sshDir, QLatin1String(GUESTSSHPATH));
+    QString configDir(installationDir.absolutePath() + HOSTCONFIGPATHPOSTFIX);
+    m_sharedPaths[VirtualMachinePrivate::SharedConfig] = QPair<QString,QString>(configDir, QLatin1String(GUESTCONFIGPATH));
+    m_sharedPaths[VirtualMachinePrivate::SharedHome] = QPair<QString,QString>(QDir::homePath(), QLatin1String(GUESTSHAREDHOME));
```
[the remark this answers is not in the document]
vigejolla: suggestions are welcome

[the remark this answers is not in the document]
vigejolla: long lines fixed

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+    arguments.append(QLatin1String(PRIVILEGED1));
+    arguments.append(QLatin1String(PRIVILEGED2));
+    arguments.append(QLatin1String(VOLUME));
+    arguments.append(QLatin1String(CGROUP_MOUNT));
+    arguments.append(QLatin1String(TMPFS));
+    arguments.append(QLatin1String(TMPFS1));
+    arguments.append(QLatin1String(TMPFS));
+    arguments.append(QLatin1String(TMPFS2));
+    arguments.append(QLatin1String(TMPFS));
+    arguments.append(QLatin1String(TMPFS3));
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+    for (auto& ports : m_reservedPortMappings) {
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    arguments.append(q->name());
+
+    auto runner = std::make_unique<ProcessRunner>(DockerManager::dockerPath(), arguments);
+    QObject::connect(runner.get(), &ProcessRunner::done, q, [q, process = runner->process()](bool ok) {
+        if (ok)
+            q->m_containerId = QString::fromLocal8Bit(process->readAllStandardOutput().trimmed());
+    });
+    QObject::connect(runner.get(), &ProcessRunner::done, context, functor);
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+           QObject::connect(runner.get(), &ProcessRunner::failure, Sdk::instance(), [=]() {
+               SdkPrivate::commandQueue()->cancelBatch(batch);
+               callIf(context_, functor, false);
+           });
+           ProcessRunner *runner_ = runner.get();
+           SdkPrivate::commandQueue()->enqueue(std::move(runner));
+           return runner_;
+       };
+
+    CommandQueue::BatchId batch = SdkPrivate::commandQueue()->beginBatch();
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+    QStringList arguments;
+    arguments.append(QLatin1String(STOP));
+    arguments.append(q->m_containerId);
+
+    enqueue(arguments, batch);
+
+    arguments[0] = QLatin1String(COMMIT);
+    arguments.append(q->name());
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+    enqueue(arguments, batch);
+
+    SdkPrivate::commandQueue()->enqueueCheckPoint(context,  [=]() { functor(true); });
```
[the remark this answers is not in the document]
vigejolla: Which space are you referring to? The empty line above?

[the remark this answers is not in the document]
martyone: Two spaces after comma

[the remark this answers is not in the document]
vigejolla: I stared and stared and just didn't see it... :) fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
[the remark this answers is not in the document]
martyone: > suggestions are welcome

This compiles with GCC 8 on Linux:

```
const QString prefix = Sdk::installationPath();

// default values hard coded now
m_reservedPortMappings[SshPort] = {HOSTSSHPORT, GUESTSSHPORT};
m_reservedPortMappings[WwwPort] = {HOSTWWWPORT, GUESTWWWPORT};

m_sharedPaths[SharedTargets] = {prefix + HOSTTARGETPATHPOSTFIX, GUESTTARGETPATH};
m_sharedPaths[SharedSsh] = {prefix + HOSTSSHPATHPOSTFIX, GUESTSSHPATH};
m_sharedPaths[SharedConfig] = {prefix + HOSTCONFIGPATHPOSTFIX, GUESTCONFIGPATH};
m_sharedPaths[SharedHome] = {QDir::homePath(), GUESTSHAREDHOME};
```

The constants suffer from bad readability. Some are already available from some *constants.h, at least we have

```
const char MER_SDK_SHARED_HOME_MOUNT_POINT[] = "/home/mersdk/share";
```

in merconstants.h and it even has a nice name to start with :) so we should make them all available from sfdkconstants.h, omiting the "MER_" prefix.

I would only do that for the paths though, not for the ports, at least at this point.

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,337 @@
...
+const char VOLUME[] = "-v";
+const char CGROUP_MOUNT[] = "/sys/fs/cgroup:/sys/fs/cgroup:ro";
+const char TMPFS[] = "--tmpfs";
+const char TMPFS1[] = "/run:exec,mode=755";
+const char TMPFS2[] = "/run/lock";
+const char TMPFS3[] = "/tmp:exec";
+const char FILTER_PREFIX[] = "--filter=ancestor=";
+const char FORMAT[] = "--format={{.Image}} {{.ID}}";
+const quint16 GUESTSSHPORT = 22;
+const quint16 HOSTSSHPORT = 2221;
```
martyone: We can use 2222 now.

src/libs/sfdk/dockermanager.cpp
```diff
@@ -0,0 +1,129 @@
...
+    return images;
+}
+
+void DockerManager::fetchRegisteredVirtualMachines(const QObject *context, const Functor<const QStringList &, bool> &functor)
+{
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    QStringList arguments;
+    arguments.append(QLatin1String(IMAGE));
```
martyone: Some more cases for QLatin1String elimination?

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,337 @@
...
+void DockerContainerPrivate::stop(const QObject *context, const Functor<bool> &functor)
+{
+    Q_Q(DockerContainer);
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    const QPointer<const QObject> context_{context};
+
+    auto enqueue = [=](const QStringList &arguments, CommandQueue::BatchId batch) {
+           auto runner = std::make_unique<ProcessRunner>(DockerManager::dockerPath(), arguments);
```
martyone: Extra indentation

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,337 @@
...
+            functor(state, true);
+        } else {
+            QStringList arguments;
+            arguments.append(CONTAINER_LIST);
+            arguments.append(QString(FILTER_PREFIX) + q->name());
+            arguments.append(FORMAT);
+            auto runner = std::make_unique<ProcessRunner>(DockerManager::dockerPath(), arguments);
+            QObject::connect(runner->process(), QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished), context,
+                    [=, runner = runner.get()](int exitCode, QProcess::ExitStatus exitStatus) mutable {
+                state = VirtualMachinePrivate::Existing;
```
martyone: It makes no difference but it would look more sanely if assigned only after the checking for process exit status/code.

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,337 @@
...
+                if ((exitStatus != QProcess::NormalExit) || (exitCode != 0)) {
```
martyone: Please avoid that https://pastebin.com/raw/PRWJsWip

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,337 @@
...
+                bool isRunning = runningContainers.contains(q->name());
+                if (isRunning) {
+                    state |= (VirtualMachinePrivate::Running | VirtualMachinePrivate::Headless);
+                    if (q->m_containerId.isEmpty())
+                    q->m_containerId = runningContainers.split(' ')[1].trimmed();
+                }
+                functor(state, true);
+            });
+            SdkPrivate::commandQueue()->enqueue(std::move(runner));
+
```
martyone: Extra blank line

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,337 @@
...
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    DockerManager::fetchRegisteredVirtualMachines(q, [=](const QStringList &registeredVms, bool ok) {
+        VirtualMachinePrivate::BasicState state;
+        if (!ok) {
+            functor(state, false);
+        } else if (!registeredVms.contains(q->name())) {
+            functor(state, true);
+        } else {
```
martyone: Looking at it now, it would be better to explicitly return from the first two blocks and omit the else block, reducing indentation and possibly adding some blank lines  e.g. after arguments list building.

src/libs/sfdk/dockercontainer.cpp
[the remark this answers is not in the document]
vigejolla: These should be now fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: What do you mean by required capacity?

martyone: I mean to check that `runningContainers.split(' ')` has at least two elements. Out of bound access results in abortion, so I would prefer checking that.

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: Right. Thanks. Fixed now - if the output contains something else than what's expected, it will print an error message and assume that the container is not running.

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed all long lines

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,361 @@
...
+            QStringList arguments;
+            arguments.append(CONTAINER_LIST);
+            arguments.append(QString(FILTER_PREFIX) + q->name());
+            arguments.append(FORMAT);
+            auto runner = std::make_unique<ProcessRunner>(DockerManager::dockerPath(), arguments);
+            QObject::connect(runner->process(),
+                QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished), context,
```
martyone: Should be indented once more

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,361 @@
...
+                state = VirtualMachinePrivate::Existing;
+                if ((exitStatus != QProcess::NormalExit) || (exitCode != 0)) {
+                    functor(state, false);
+                    return;
+                }
+
+                QStringList runningContainers =
+                    QString::fromLocal8Bit(runner->process()
+                        ->readAllStandardOutput()).split('\n', QString::SkipEmptyParts);
+                for (const QString runningContainer : qAsConst(runningContainers)) {
```
martyone: Better declare `runningContainers` as const. Use `const QString &` for looping.

I would consider breaking the above line before `.split` instead.

src/libs/sfdk/dockercontainer.cpp
```diff
@@ -0,0 +1,361 @@
...
+                    QStringList nameAndId = runningContainer.split(' ', QString::SkipEmptyParts);
+                    if ((nameAndId.count() == 2) && (nameAndId.first() == q->name())) {
+                        state |= VirtualMachinePrivate::Running | VirtualMachinePrivate::Headless;
+                        if (q->m_containerId.isEmpty()) {
+                            q->m_containerId = nameAndId.last();
+                        }
```
martyone: Extra braces

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed - and most likely introduced a dozen new problems while fixing ;)

src/libs/sfdk/dockercontainer.cpp
martyone: I would prefer using just `QTC_ASSERT(nameAndId.count() == 2 && nameAndId.first() == q->name(), continue)` there.

src/libs/sfdk/dockercontainer.cpp
martyone: There are some occurrences over 120 chars wide yet (which I'd take as a hard limit) and some more over 100 chars (which is recommended by the coding rules) that could be wrapped :)

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+    // TODO: fetch snapshot info if requested
+    Q_UNUSED(extraInfo)
+
+    QStringList arguments;
+    arguments.append(COMMAND_INSPECT);
+    arguments.append(q->name());
+    arguments.append(FORMAT_LABEL_LIST);
+
+    auto info = std::make_shared<VirtualMachineInfo>();
```
martyone: In this class the info is built in single step, so it is ok to use an automatic variable scoped inside the lambda.

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+void DockerVirtualMachinePrivate::start(const QObject *context, const Functor<bool> &functor)
+{
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+    Q_Q(DockerVirtualMachine);
+
+    CommandQueue::BatchId batch = SdkPrivate::commandQueue()->beginBatch();
+
+    if (m_reservedPortMappings.size() == 0) {
```
martyone: I think we should assume (and ensure*) that `start` is never called before the info is fetched.

*) If it happened for you, then `enableUpdates` and `updateOnce` under `BuildEnginePrivate` and `EmulatorPrivate` may need to be fixed, delaying the `setAutoConnectEnabled` call until the `updateVmProperties` call finishes.

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,494 @@
...
+    return label;
+}
+
+void DockerVirtualMachinePrivate::buildWithLabel(const QString& key, const QString& value,
+        const QObject *context, const Functor<bool> &functor)
+{
+    Q_Q(DockerVirtualMachine);
+    QStringList arguments;
+    arguments.append(COMMAND_BUILD);
+    arguments.append(DOCKERFILE_PREFIX + Sdk::installationPath() + DOCKERFILE_POSTFIX);
```
martyone: It seems possible to pass "-" instead and write the dockerfile on stdin with something like

```
const QByteArray dockerfile = ...;
connect(runner.get(), &QProcess::started, [=]() {
    runner->process()->write(dockerfile);
    runner->process()->closeWriteChannel();
});
```

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,494 @@
...
+    arguments.append(TAG);
+    arguments.append(q->name());
+    arguments.append(Sdk::installationPath());
+
+    const QPointer<const QObject> context_{context};
+
+    auto runner = std::make_unique<ProcessRunner>(DockerManager::dockerPath(), arguments);
+    QObject::connect(runner.get(), &ProcessRunner::done, q, [context_, functor, process = runner->process()](bool ok) {
+        callIf(context_, functor, ok);
+    });
```
martyone: `QObject::connect(runner.get(), &ProcessRunner::done, context, functor);` should be enough here.

src/libs/sfdk/dockervirtualmachine.h
```diff
@@ -0,0 +1,57 @@
...
+    static QString staticDisplayType();
+    static void fetchRegisteredVirtualMachines(const QObject *context,
+            const Functor<const QStringList &, bool> &functor);
+
+private:
+    Q_DISABLE_COPY(DockerVirtualMachine)
+    Q_DECLARE_PRIVATE(DockerVirtualMachine)
+
+    mutable QString m_containerId;
+    const std::unique_ptr<UserSettings> m_userSettings;
```
martyone: m_userSettings unused now. m_containerId should go to the private class.

src/libs/sfdk/dockervirtualmachine_p.h
```diff
@@ -0,0 +1,96 @@
...
+            const QObject *context, const Functor<bool> &functor) override;
+    void doSetReservedPortListForwarding(ReservedPortList which, const QList<Utils::Port> &ports,
+            const QObject *context, const Functor<const QMap<QString, quint16> &, bool> &functor)
+        override;
+
+    void doRestoreSnapshot(const QString &snapshotName, const QObject *context,
+        const Functor<bool> &functor) override;
+
+private:
+    void doSetReservedPortForwarding(ReservedPort which, quint16 port);
```
martyone: Seems unused now

src/libs/sfdk/dockervirtualmachine_p.h
```diff
@@ -0,0 +1,96 @@
...
+
+    void setSharedPathFromJSon(SharedPath which, const QJsonObject& json, VirtualMachineInfo& info);
```
martyone: Seems unused

src/libs/sfdk/dockervirtualmachine_p.h
```diff
@@ -0,0 +1,96 @@
...
+    VirtualMachineInfo virtualMachineInfoFromOutput(const QByteArray& output) const;
+
+    static QString label(SharedPath which);
+    static QString label(ReservedPort which);
```
martyone: I would prefer naming it `labelFor` (the argument is not of a type dedicated to select a label).

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: I'll take your word for it - and add an assert.

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,494 @@
...
+    if (document.isNull())
+        return info;
+
+    QJsonObject labels = document.object();
+
+    auto parsePort = [=](VirtualMachinePrivate::ReservedPort which, quint16& port, quint16 guestPort) {
+        QString key = label(which);
+        if (labels.contains(key)) {
+            port = labels.value(key).toInt();
+            m_reservedPortMappings[which] = QPair<quint16, quint16>(port, guestPort);
```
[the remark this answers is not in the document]
vigejolla: I wondered about this as well yesterday, but couldn't make up my mind on which way is better. Now, after sleep, it feels quite obvious that it's better to ditch the member variables (gets rid of the ugly mutable keywords etc).

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,494 @@
...
+        }
+    };
+
+    parsePort(VirtualMachinePrivate::SshPort, info.sshPort, GUESTSSHPORT);
+    parsePort(VirtualMachinePrivate::WwwPort, info.wwwPort, GUESTWWWPORT);
+
+    auto parsePath = [=](VirtualMachinePrivate::SharedPath which, QString& path, const QString& guestPath) {
+        QString key = label(which);
+        if (labels.contains(key)) {
+            path = labels.value(key).toInt();
```
vigejolla: note to self: converting a path to int is not a good idea

src/libs/sfdk/dockercontainer.cpp
vigejolla: makes sense - fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: Fixed again

src/libs/sfdk/dockercontainer.h
vigejolla: fixed

src/libs/sfdk/dockercontainer_p.h
vigejolla: fixed

src/libs/sfdk/dockercontainer_p.h
vigejolla: fixed

src/libs/sfdk/dockercontainer_p.h
vigejolla: fixed

src/libs/sfdk/dockercontainer_p.h
vigejolla: fixed

src/libs/sfdk/dockercontainer_p.h
vigejolla: fixed

src/libs/sfdk/dockercontainer_p.h
vigejolla: Instead of fixing spaces, removed  the members entirely

src/libs/sfdk/dockercontainer_p.h
vigejolla: fixed

src/libs/sfdk/dockermanager.cpp
vigejolla: fixed

src/libs/sfdk/dockermanager.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: Old habits die hard - When you have been coding last 20 years following the instructions that you should always add parenthesis to make it easier to the reader, it kind of comes automatically.

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,531 @@
...
+        arguments.append(CONTAINER_LIST);
+        arguments.append(QString(FILTER_PREFIX) + q->name());
+        arguments.append(FORMAT_CONTAINER_LIST);
+
+        auto runner = std::make_unique<ProcessRunner>(dockerPath(), arguments);
+        QObject::connect(runner->process(),
+            QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished), context,
+                [=, runner = runner.get()](int exitCode,
+                                           QProcess::ExitStatus exitStatus) mutable {
+                if (exitStatus != QProcess::NormalExit || exitCode != 0) {
```
martyone: Bad indent of the second argument and body. I would prefer this:

```
QObject::connect(runner->process(),
        QOverload<int, QProcess::ExitStatus>::of(&QProcess::finished),
        context,
        [=, runner = runner.get()](int exitCode, QProcess::ExitStatus exitStatus) mutable {
    if (exitStatus != QProcess::NormalExit || exitCode != 0) {
```

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,531 @@
...
+
+QString DockerVirtualMachinePrivate::dockerPath()
+{
+    if (s_dockerPath()->isNull()) {
+        QStringList searchPaths = QProcessEnvironment::systemEnvironment()
+            .value("PATH").split(QChar(':'));
+        foreach (const QString &searchPath, searchPaths) {
+            QDir dir(searchPath);
+            if (dir.exists(DOCKER)) {
+                s_dockerPath()->operator=(dir.absoluteFilePath(DOCKER));
```
martyone: IIUC you use Q_GLOBAL_STATIC to avoid searching for docker endlessly when it is not installed. I think we should instead guard the `d->virtualMachineFactory->registerType<DockerVirtualMachine>();` call with `if (!dockerPath().isNull())` and do similarly for vbox which now becomes optional too.

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,531 @@
...
+        }
+    }
+    return *s_dockerPath();
+}
+
+QStringList DockerVirtualMachinePrivate::listedImages(const QString &output)
+{
+    QStringList images;
+    QStringList lines = output.split(QRegularExpression("[\r\n]"), QString::SkipEmptyParts);
+    for (QString line : lines) {
```
martyone: const QStringList lines, const QString &line

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,531 @@
...
+
+QString DockerVirtualMachine::staticDisplayType()
+{
+    return tr("Docker");
+}
+
+void DockerVirtualMachine::fetchRegisteredVirtualMachines(const QObject *context,
+        const Functor<const QStringList &, bool> &functor)
+{
+    DockerVirtualMachinePrivate::fetchRegisteredVirtualMachines(context, functor);
```
martyone: Unnecessary indirection now

src/plugins/mer/merdeploysteps.cpp
```diff
@@ -689,7 +689,7 @@ void MerMb2RpmBuildStep::stdOutput(const QString &line)
     if (rexp.indexIn(line) != -1) {
         QString file = rexp.cap(1);
         //TODO First replace shared home and then shared src (error prone!)
-        file.replace(QRegExp(QString("^") + Constants::MER_SDK_SHARED_HOME_MOUNT_POINT), m_sharedHome);
+        file.replace(QRegExp(QString("^") + Sfdk::Constants::BUILD_ENGINE_GUEST_SHARED_HOME), m_sharedHome);
         file.replace(QRegExp(QString("^") + Constants::MER_SDK_SHARED_SRC_MOUNT_POINT), m_sharedSrc);
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: No, actually it was the next line which had wrong indentation level. Please check if you like the current state.

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,494 @@
...
+
+    return data;
+}
+
+VirtualMachineInfo DockerVirtualMachinePrivate::virtualMachineInfoFromOutput(const QByteArray &output) const
+{
+    QJsonDocument document = QJsonDocument::fromJson(output);
+    VirtualMachineInfo info = {};
+
+    if (document.isNull())
```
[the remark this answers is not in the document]
vigejolla: replaced with QTC_ASSERT

src/libs/sfdk/dockercontainer.cpp
martyone: Argument list (first level parentheses) should be indented twice, so really, this line was wrong. If this is the current code https://github.com/sailfishos/sailfish-qtcreator/pull/223/files#diff-8c6731e3f6adcdd8ada5e49b9f656c07R229, then there is a newer comment with suggested wrapping.

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,494 @@
...
+        return info;
+
+    QJsonObject labels = document.object();
+
+    auto parsePort = [=](VirtualMachinePrivate::ReservedPort which, quint16& port, quint16 guestPort) {
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: fixed

src/libs/sfdk/dockercontainer.cpp
vigejolla: Fixed now according to the other comment

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: Implemented now this way

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.h
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine_p.h
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,464 @@
...
+{
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+    Q_ASSERT(cachedInfo().sshPort != 0);
+    Q_ASSERT(cachedInfo().wwwPort != 0);
+    Q_ASSERT(!cachedInfo().sharedSrc.isEmpty());
+    Q_ASSERT(!cachedInfo().sharedHome.isEmpty());
+    Q_ASSERT(!cachedInfo().sharedTargets.isEmpty());
+    Q_ASSERT(!cachedInfo().sharedConfig.isEmpty());
+    Q_ASSERT(!cachedInfo().sharedSsh.isEmpty());
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,464 @@
...
+    auto sharedPath = [&arguments](const QString& hostPath, const QString& guestPath) {
+        arguments.append(VOLUME);
+        arguments.append(hostPath + ":" + guestPath);
+    };
+
+    sharedPath(cachedInfo().sharedHome, Constants::BUILD_ENGINE_GUEST_SHARED_HOME);
+    sharedPath(cachedInfo().sharedSrc, Constants::BUILD_ENGINE_GUEST_SHARED_SRC);
+    sharedPath(cachedInfo().sharedTargets, Constants::BUILD_ENGINE_GUEST_SHARED_TARGET);
+    sharedPath(cachedInfo().sharedConfig, Constants::BUILD_ENGINE_GUEST_SHARED_CONFIG);
+    sharedPath(cachedInfo().sharedSsh, Constants::BUILD_ENGINE_GUEST_SHARED_SSH);
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,531 @@
...
+
+#include <utils/fileutils.h>
+#include <utils/optional.h>
+#include <utils/qtcassert.h>
+
+#include <QDir>
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QProcess>
+#include <QGlobalStatic>
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,529 @@
...
+    Q_UNUSED(extraInfo)
+
+    QStringList arguments;
+    arguments.append(COMMAND_INSPECT);
+    arguments.append(q->name());
+    arguments.append(FORMAT_LABEL_LIST);
+
+    auto runner = std::make_unique<ProcessRunner>(dockerPath(), arguments);
+    QObject::connect(runner.get(), &ProcessRunner::done, q,
+                     [this, functor, process = runner->process()](bool ok) {
```
martyone: `context` should be passed instead of `q`.

BTW, aligning arguments under opening brace often hits space limitations, so it can hardly be applied consistently, so something I would rather avoid. Please consider realigning such code to just plain two levels shift.

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,529 @@
...
+    arguments.append(PRIVILEGED1);
+    arguments.append(PRIVILEGED2);
+    arguments.append(VOLUME);
+    arguments.append(CGROUP_MOUNT);
+
+    auto addTmpfs = [&arguments](const QString& path) {
+        arguments.append(TMPFS);
+        arguments.append(path);
+    };
+
```
martyone: nitpick: I would stick the definition and use together

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,529 @@
...
+void DockerVirtualMachinePrivate::doSetStorageSizeMb(int storageSizeMb, const QObject *context,
+        const Functor<bool> &functor)
+{
+    Q_UNUSED(storageSizeMb)
+    Q_UNUSED(context)
+    Q_UNUSED(functor)
+    QTC_CHECK(false);
+}
+
+
```
martyone: extra blank

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,529 @@
...
+    };
+
+    parsePath(VirtualMachinePrivate::SharedSrc, &info.sharedSrc);
+    parsePath(VirtualMachinePrivate::SharedSsh, &info.sharedSsh);
+    parsePath(VirtualMachinePrivate::SharedHome, &info.sharedHome);
+    parsePath(VirtualMachinePrivate::SharedConfig, &info.sharedConfig);
+    parsePath(VirtualMachinePrivate::SharedTargets, &info.sharedTargets);
+
+    info.memorySizeMb = VirtualMachine::availableMemorySizeMb();
+    info.cpuCount = VirtualMachine::availableCpuCount();
```
martyone: I would consider moving these two near the beginning

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,529 @@
...
+        break;
+    case SharedConfig:
+        label = Constants::BUILD_ENGINE_SHARED_CONFIG;
+        break;
+    case SharedSsh:
+        label = Constants::BUILD_ENGINE_SHARED_SSH;
+        break;
+    case SharedTargets:
+        label = Constants::BUILD_ENGINE_SHARED_TARGET;
+    break;
```
martyone: bad indent.

Better return immediately from the cases?

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,529 @@
...
+    return tr("Docker");
+}
+
+void DockerVirtualMachine::fetchRegisteredVirtualMachines(const QObject *context,
+        const Functor<const QStringList &, bool> &functor)
+{
+    DockerVirtualMachinePrivate::fetchRegisteredVirtualMachines(context, functor);
+}
+
+}
```
martyone: `// namespace Sfdk`

src/libs/sfdk/dockervirtualmachine_p.h
```diff
@@ -0,0 +1,110 @@
...
+****************************************************************************/
+
+#pragma once
+
+#include "virtualmachine_p.h"
+
+namespace Sfdk {
+
+class DockerVirtualMachinePrivate;
+
```
martyone: extra blank

src/libs/sfdk/sfdkconstants.h
```diff
@@ -182,5 +190,7 @@ const char HARDWARE_DEVICE_QML_LIVE_PORTS[] = "QmlLivePorts";
 
 const char EMULATOR_DEVICE_EMULATOR_URI[] = "EmulatorUri";
 
+const char DOCKER_SETTINGS_FILE_NAME[] = "docker.xml";
+const char DOCKER_SETTINGS_DOC_TYPE[] = "QtCreatorSfdkDocker";
```
martyone: Unused now

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,531 @@
...
+
+void DockerVirtualMachinePrivate::probe(const QObject *context,
+        const Functor<VirtualMachinePrivate::BasicState, bool> &functor) const
+{
+    Q_Q(const DockerVirtualMachine);
+    Q_ASSERT(context);
+    Q_ASSERT(functor);
+
+    fetchRegisteredVirtualMachines(q, [=](const QStringList &registeredVms, bool ok) {
```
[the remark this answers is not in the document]
vigejolla: Using now <code>context</code> as context, as there is no longer any reason to use <code>q</code>.

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: Implemented now using DockerVirtualMachine::isAvailable() and VBoxVirtualMachine::isAvailable()

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,529 @@
...
+    QTC_CHECK(false);
+}
+
+VirtualMachineInfo DockerVirtualMachinePrivate::virtualMachineInfoFromOutput(
+        const QByteArray &output) const
+{
+    QJsonDocument document = QJsonDocument::fromJson(output);
+    VirtualMachineInfo info = {};
+
+    QTC_ASSERT(!document.isNull(), return info);
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,529 @@
...
+    sharePath(Constants::BUILD_ENGINE_SHARED_SSH_MOUNT_POINT, cachedInfo().sharedSsh);
+
+    arguments.append(q->name());
+
+    auto runner = std::make_unique<ProcessRunner>(dockerPath(), arguments);
+    QObject::connect(runner.get(), &ProcessRunner::done, q,
+                     [this, functor, process = runner->process()](bool ok) {
+        if (ok)
+            m_containerId = QString::fromLocal8Bit(process->readAllStandardOutput().trimmed());
+        functor(ok);
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,529 @@
...
+    QJsonDocument document = QJsonDocument::fromJson(output);
+    VirtualMachineInfo info = {};
+
+    QTC_ASSERT(!document.isNull(), return info);
+
+    QJsonObject labels = document.object();
+
+    auto parsePort = [=](VirtualMachinePrivate::ReservedPort which, quint16 *port) {
+        QString key = labelFor(which);
+        if (labels.contains(key))
```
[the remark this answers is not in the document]
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: Fixed. And modified my emacs indentation rules according to your suggestion.

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: you mean <code>arguments.append("--tmpfs")</code> instead of <code>arguments.append(TMPFS)</code>?

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine_p.h
vigejolla: removed

src/libs/sfdk/sfdkconstants.h
vigejolla: removed

src/libs/sfdk/dockervirtualmachine.cpp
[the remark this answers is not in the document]
martyone: Unfortunately it is not that easy. ``m_container` is still accessed there and that needs to be guarded with `q`. Even if it can be accessed directly now, its lifetime is bound to `q`. Wouldn't it be possible to initialize `m_containerId` from `fetchInfo` instead to get rid of that?

src/libs/sfdk/dockervirtualmachine.cpp
martyone: Oops, misplaced `&` yet.

src/libs/sfdk/dockervirtualmachine.cpp
martyone: Nice :) There is one more case on line 155 left.

src/libs/sfdk/dockervirtualmachine.cpp
martyone: I mean I would omit the blank line between lambda definition and its calls here to help eyes spotting their purpose without actually reading the code :)

src/libs/sfdk/dockervirtualmachine.cpp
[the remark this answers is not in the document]
vigejolla: We don't know the container id in fetchInfo, as that method only deals with images (not containers)

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,519 @@
...
+        return Constants::BUILD_ENGINE_SHARED_HOME;
+    case SharedConfig:
+        return Constants::BUILD_ENGINE_SHARED_CONFIG;
+    case SharedSsh:
+        return Constants::BUILD_ENGINE_SHARED_SSH;
+    case SharedTargets:
+        return Constants::BUILD_ENGINE_SHARED_TARGET;
+    }
+    Q_ASSERT(false);
+    return QString();
```
martyone: nitpick: I enjoy using `return {}` for these.

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,519 @@
...
+QString DockerVirtualMachinePrivate::dockerPath()
+{
+    static QString s_dockerPath;
+    if (s_dockerPath.isNull()) {
+        QStringList searchPaths = QProcessEnvironment::systemEnvironment()
+            .value("PATH").split(QChar(':'));
+        foreach (const QString &searchPath, searchPaths) {
+            QDir dir(searchPath);
+            if (dir.exists(DOCKER)) {
+                s_dockerPath=dir.absoluteFilePath(DOCKER);
```
martyone: Spaces

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,519 @@
...
+        if (!ok) {
+            functor({}, false);
+            return;
+        }
+        QStringList vms = DockerVirtualMachinePrivate::listedImages(QString::fromLocal8Bit(
+                runner->process()->readAllStandardOutput()));
+        functor(vms, true);
+    });
+    SdkPrivate::commandQueue()->enqueue(std::move(runner));
+
```
martyone: extra blank line

src/libs/sfdk/dockervirtualmachine_p.h
```diff
@@ -0,0 +1,109 @@
...
+public:
+    explicit DockerVirtualMachine(const QString &name, QObject *parent = nullptr);
+    ~DockerVirtualMachine() override;
+
+    static QString staticType();
+    static QString staticDisplayType();
+    static void fetchRegisteredVirtualMachines(const QObject *context,
+            const Functor<const QStringList &, bool> &functor);
+    static bool isAvailable();
```
martyone: I would prefer moving `isAvailable` up - it's the first and possibly last function to call after all :) Similarly for vbox.

src/libs/sfdk/dockervirtualmachine_p.h
```diff
@@ -0,0 +1,109 @@
...
+    void doAddPortForwarding(const QString &ruleName,
+            const QString &protocol, quint16 hostPort, quint16 emulatorVmPort,
+            const QObject *context, const Functor<bool> &functor) override;
+    void doRemovePortForwarding(const QString &ruleName,
+            const QObject *context, const Functor<bool> &functor) override;
+    void doSetReservedPortForwarding(ReservedPort which, quint16 port,
+            const QObject *context, const Functor<bool> &functor) override;
+    void doSetReservedPortListForwarding(ReservedPortList which, const QList<Utils::Port> &ports,
+            const QObject *context, const Functor<const QMap<QString, quint16> &, bool> &functor)
+            override;
```
martyone: nitpick: `override` should be indented one level less

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: Fixed this and a few other occurrences as well

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: fixed that as well

src/libs/sfdk/dockervirtualmachine.cpp
martyone: I was looking for the variable in class declaration... only now I noticed it is local to this function :) omit the `s_` please.

src/libs/sfdk/dockervirtualmachine_p.h
```diff
@@ -0,0 +1,109 @@
...
+    static QString labelFor(ReservedPort which);
+
+    static QString dockerPath();
+
+    static QStringList listedImages(const QString &output);
+
+    void buildWithLabel(const QString& key, const QString& value,
+                        const QObject *context, const Functor<bool> &functor);
+
+    mutable QString m_containerId;
```
martyone: It is now in the private class, so `m_` should not be used.

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -0,0 +1,524 @@
...
+    QStringList arguments;
+    arguments.append(COMMAND_BUILD);
+    arguments.append(LABEL_PREFIX + key + "=" + value);
+    arguments.append(TAG);
+    arguments.append(q->name());
+    arguments.append("-");
+
+    auto runner = std::make_unique<ProcessRunner>(dockerPath(), arguments);
+
+    QObject::connect(runner->process(), &QProcess::started, [q, process = runner->process()]() {
```
martyone: `q` should be used as contex

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: fixed this and one other location

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine_p.h
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine_p.h
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine_p.h
vigejolla: fixed

src/libs/sfdk/dockervirtualmachine.cpp
vigejolla: fixed

src/plugins/mer/merdeploysteps.cpp
```diff
@@ -689,8 +689,8 @@ void MerMb2RpmBuildStep::stdOutput(const QString &line)
     if (rexp.indexIn(line) != -1) {
         QString file = rexp.cap(1);
         //TODO First replace shared home and then shared src (error prone!)
-        file.replace(QRegExp(QString("^") + Constants::MER_SDK_SHARED_HOME_MOUNT_POINT), m_sharedHome);
-        file.replace(QRegExp(QString("^") + Constants::MER_SDK_SHARED_SRC_MOUNT_POINT), m_sharedSrc);
+        file.replace(QRegExp(QString("^") + Sfdk::Constants::BUILD_ENGINE_SHARED_HOME_MOUNT_POINT), m_sharedHome);
+        file.replace(QRegExp(QString("^") + Sfdk::Constants::BUILD_ENGINE_SHARED_SRC_MOUNT_POINT), m_sharedSrc);
```
martyone: I would extract these into a separate commit "Mer: Move constants for shared paths to Sfdk". That way it will also become more clear that all the enablers for docker build engine exist on sfdk lib side only.

vigejolla: moved to a separate commit

PR #225, opened 2019-10-31
Jb45560+

martyone:


src/tools/sfdk/command.cpp
```diff
@@ -647,3 +1045,5 @@ std::unique_ptr<Worker> EngineWorker::fromMap(const QVariantMap &data, int versi
     return worker;
 #endif
 }
+
+#include "command.moc"
```
vigejolla: fatal error: command.moc: No such file or directory
 #include "command.moc"
          ^~~~~~~~~~~~~
compilation terminated.

I believe this is because Qt does not notice the Q_OBJECT for SetPropertiesTask class, as it is not in a header file, and that's why it doesn't generate the moc file. Not that having class declarations in cpp files is a good habit anyway.

martyone: you need to run qmake again (make -C src/tools/sfdk qmake)

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -310,6 +314,87 @@ void SdkManager::setEnableReversePathMapping(bool enable)
     s_instance->m_enableReversePathMapping = enable;
 }
 
+Device *SdkManager::configuredDevice(QString *errorMessage)
+{
+    Q_ASSERT(errorMessage);
+
+    const Option *const deviceOption = Dispatcher::option("device"); // FIXME
```
vigejolla: Something you've forgotten or planning to fix in the near future?

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -310,6 +314,87 @@ void SdkManager::setEnableReversePathMapping(bool enable)
...
+    QTC_ASSERT(deviceOption, return nullptr);
+
+    const Utils::optional<OptionEffectiveOccurence> effectiveDeviceOption =
+        Configuration::effectiveState(deviceOption);
+    if (!effectiveDeviceOption) {
+        *errorMessage = tr("No device selected");
+        return nullptr;
+    }
+    QTC_ASSERT(!effectiveDeviceOption->isMasked(), return nullptr);
```
vigejolla: Shouldn't we put something into the errorMessage even in this case? It might help debug future bugs.

martyone: I prefer not doing that - it would be practically a dead code (assertions should only fail on internal errors)

src/tools/sfdk/command.cpp
```diff
@@ -59,7 +62,309 @@ const char SDK_MAINTENANCE_TOOL[] = "SDKMaintenanceTool.app/Contents/MacOS/SDKMa
...
+private:
+    QPointer<Emulator> m_emulator;
+    std::unique_ptr<VirtualMachinePropertiesAccessor> m_vmAccessor;
+};
+
+/*!
+ * \class BuildEnginePropertiesAccessor
+ */
+
+class BuildEnginePropertiesAccessor : public PropertiesAccessor
```
vigejolla: To me it looks like BuildEnginePropertiesAccessor and EmulatorPropertiesAccessor are identical except for the contained type (Emulator vs BuildEngine). Consider using a template?

src/tools/sfdk/sdkmanager.cpp
martyone: The use of the string literal here was a bit controversial to me.

I prefer NOT to use constants blindly for every primitive value. E.g. when a constant is used for a value that is specified by an external entity, where the context is clear and the value is not likely to change, all it adds is an unnecessary overhead for both the writer and readers of the code. Consider embedding an SQL query in C++ code - would we define a constant for every SQL keyword and build the queries using that? - not likely. A deterrent example is how do we work with vboxmanage :)

The case with "device" option is not that straightforward. It comes from sfdk module definition, which isn't really external to sfdk. It is only external to its C++ side. At the same time there is currently no mechanism to share definitions between those. So what the FIXME follows here is an opportunity to add such a mechanism and use it here.

Given that 1) I am not aware of any other case similar to this, 2) it is likely to fail soon when the other side changes and 3) improving it wouldn't be for free, I think relying on the assert below is just enough. I will remove the FIXME.

vigejolla: Ok. Thanks for the explanation :)

src/tools/sfdk/command.cpp
```diff
@@ -59,7 +62,309 @@ const char SDK_MAINTENANCE_TOOL[] = "SDKMaintenanceTool.app/Contents/MacOS/SDKMa
...
+            return true;
+        } else if (name == VM_CPU_COUNT) {
+            if (!parsePositiveInt(&m_cpuCount, value, errorString))
+                return false;
+            if (m_cpuCount > VirtualMachine::availableCpuCount()) {
+                *errorString = valueTooBigMessage();
+                return false;
+            }
+            return true;
+        } else if (name == VM_VDI_CAPACITY_MB) {
```
jlehtoranta: Have we got any easy source for getting the free disk space amount on host? Does Virtualbox allow setting the VDI capacity bigger than the free disk space on host?

martyone: > Does Virtualbox allow setting the VDI capacity bigger than the free disk space on host?

For dynamic size images it does not care I believe. When all space is used, it just pauses the VM and let's you free some (I saw this on SDK builder many times)

> Have we got any easy source for getting the free disk space amount on host?

No answer at hand, but since the same issue that we do not bound the value exists on Qt Creator side, I would only care about that in a separate task.

jlehtoranta: Ok, let's go with this approach for the time being 👍

PR #227, opened 2019-11-11
Mer: Install SDK tools if they are missing

Tomin1:


src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -237,6 +263,118 @@ QString MerHardwareDeviceWizardSelectionPage::detectDeviceName(
...
+    loop.exec();
+
+    if (!runner.lastConnectionErrorString().isEmpty()
+            || runner.processExitStatus() != SshRemoteProcess::NormalExit
+            || runner.processExitCode() != 0) {
+        qWarning() << "Failed to install patterns-sailfish-sdk-client-tools";
+        return false;
+    }
+
+    passwordData.truncate(0);
```
martyone: Why only in the successful case? Also if I am correct that you do it to clear the password from process memory, I wouldn't be that optimistic about the effect of truncate().

Tomin1: No reason, just a simple mistake. Also meant that truncate(0) to be clear() which is used in devicelock for this purpose.

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -237,6 +263,118 @@ QString MerHardwareDeviceWizardSelectionPage::detectDeviceName(
...
+        qInfo() << "patterns-sailfish-sdk-client-tools is not installed";
+        return false;
+    }
+
+    qInfo() << "patterns-sailfish-sdk-client-tools is installed";
+    *ok = true;
+    return true;
+}
+
+bool MerHardwareDeviceWizardSelectionPage::askInstallSdkClientTools(QString &password)
```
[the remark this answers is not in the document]
Tomin1: Let's see if I can do that.

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -237,6 +263,118 @@ QString MerHardwareDeviceWizardSelectionPage::detectDeviceName(
...
+    const QString osVersion = QString::fromLatin1(runner.readAllStandardOutput()).trimmed();
+    if (osVersion.isEmpty()) {
+        *ok = false;
+        qWarning() << "Empty output from querying os version name on target";
+        return false;
+    }
+
+    if (QVersionNumber::fromString(osVersion) < QVersionNumber(3, 2, 1)) {
+        *ok = true;
+        qDebug() << "Sailfish OS" << osVersion << "assume that packages are installed";
```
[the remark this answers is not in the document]
rainemak: Try to get rid of mer logging and replace it with "org.sailfishos.sdk.*"

"org.sailfishos.sdk.qmakeArgs"
"org.sailfishos.sdk.qmlLive"
"org.sailfishos.sdk"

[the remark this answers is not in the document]
martyone: nope, only "qtc.sailfishos.*" would be compliant with upstream conventions (And only if we really renamed the plugin from Mer to SailfishOS, which we didn't. There are more issues at the code level when comes to the legacy naming, that are not likely to disappear soon - I am not a fan of breaking git history just for fun)

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -148,11 +153,32 @@ void MerHardwareDeviceWizardSelectionPage::handleTestConnectionClicked()
...
 
+    m_ui->connectionTestLabel->setText(tr("Detecting SDK tools..."));
+    m_sdkClientToolsInstalled = detectSdkClientTools(sshParams, &m_connectionTestOk);
+    if (!m_connectionTestOk) {
+        m_ui->connectionTestLabel->setText(tr("Could not autodetect SDK tools"));
+        goto end;
+    }
+    if (!m_sdkClientToolsInstalled) {
+        m_ui->connectionTestLabel->setText(tr("SDK tools are not installed"));
+        if (askInstallSdkClientTools(password)) {
```
vigejolla: password is only used in this block; consider moving the definition here

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -148,11 +153,32 @@ void MerHardwareDeviceWizardSelectionPage::handleTestConnectionClicked()
...
+            goto end;
+        }
+    }
+
     m_ui->connectionTestLabel->setText(tr("Connected %1 device").arg(m_deviceName));
 
 end:
     m_ui->testButton->setEnabled(true);
     m_isIdle = true;
+    password.truncate(0);
```
vigejolla: password is about to go out of scope, what's the point of truncate?

src/plugins/mer/merhardwaredevicewizardpages.h
```diff
@@ -67,11 +67,15 @@ private slots:
     static ProjectExplorer::Abi::Architecture detectArchitecture(
             const QSsh::SshConnectionParameters &sshParams, bool *ok);
     static QString detectDeviceName(const QSsh::SshConnectionParameters &sshParams, bool *ok);
+    static bool detectSdkClientTools(const QSsh::SshConnectionParameters &sshParams, bool *ok);
```
vigejolla: Since this method is only called from a non-static method, it doesn't have to be static. And since the return value is used for setting m_sdkClientToolsInstalled, I would even go so far that I would manipulate m_sdkClientToolsInstalled and m_connectionTestOk directly in detectSdkClientTools().

vigejolla: ... I can see that you are following the example set by detectDeviceName here, so may be it's not worth it changing the implementation here. Consider the comment above more like just food for thought than a change request.

src/plugins/mer/merhardwaredevicewizardpages.cpp
Tomin1: It was meant to be clear() but anyway looking at qt code calling clear doesn't do anything that destroying QString wouldn't do, so I'll just remove these.

In devicelock code values containing security code are always cleared so I did that here too but that's not really useful.

src/plugins/mer/merhardwaredevicewizardpages.cpp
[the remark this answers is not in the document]
rainemak: These have nothing to do with upstream conventions.

src/plugins/mer/merhardwaredevicewizardpages.cpp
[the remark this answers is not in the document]
Tomin1: Hmmh. Doesn't seem to be doable AFAICT. pkexec normally asks for authorization though agent but in our case nothing has registered as an agent so it doesn't work. Devicelock doesn't provide such feature either. If someone knows better please tell.

src/plugins/mer/merhardwaredevicewizardpages.cpp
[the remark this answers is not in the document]
martyone: I think you missed the context these sources live in :)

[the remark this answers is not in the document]
Tomin1: Changed it to "mer" for now, if this debate leads to another one I'll change it. It's quite a bit easier change anyway.

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -237,6 +262,117 @@ QString MerHardwareDeviceWizardSelectionPage::detectDeviceName(
...
+    const QString osVersion = QString::fromLatin1(runner.readAllStandardOutput()).trimmed();
+    if (osVersion.isEmpty()) {
+        *ok = false;
+        qWarning() << "Empty output from querying os version name on target";
+        return false;
+    }
+
+    if (QVersionNumber::fromString(osVersion) < QVersionNumber(3, 2, 1)) {
+        *ok = true;
+        qCDebug(Mer::Log::mer) << "Sailfish OS" << osVersion << "assume that packages are installed";
```
martyone: Just `Log::mer` please :)

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -237,6 +262,117 @@ QString MerHardwareDeviceWizardSelectionPage::detectDeviceName(
...
+            &loop, &QEventLoop::quit);
+    runner.run("sed -n '/^VERSION_ID=/s/^VERSION_ID=\\([0-9]\\+\\.[0-9]\\+\\.[0-9]\\+\\).*/\\1/p'"
+               " /etc/os-release", sshParams);
+    loop.exec();
+
+    if (!runner.lastConnectionErrorString().isEmpty()
+            || runner.processExitStatus() != SshRemoteProcess::NormalExit
+            || runner.processExitCode() != 0) {
+        *ok = false;
+        qWarning() << "Failed to query os version name on target";
```
rainemak: qCWarning

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -237,6 +262,117 @@ QString MerHardwareDeviceWizardSelectionPage::detectDeviceName(
...
+        return false;
+    }
+
+    const QString osVersion = QString::fromLatin1(runner.readAllStandardOutput()).trimmed();
+    if (osVersion.isEmpty()) {
+        *ok = false;
+        qWarning() << "Empty output from querying os version name on target";
```
rainemak: qCWarning

Tomin1: Changed all of the qWarnings in that file to use it, I hope that's fine.

Tomin1:
I wonder if I should change this to try installing first, then refresh if needed and try again because refreshing repositories can take a lot of time.

Tomin1:
> I wonder if I should change this to try installing first, then refresh if needed and try again because refreshing repositories can take a lot of time.

Decided to implement it anyway, I think it improves it a lot. Failing at wrong password is still fast and I don't see any harm trying it first.

src/plugins/mer/merhardwaredevicewizardpages.cpp
martyone: Could you please change it in all files under src/plugins/mer then? There is just few occurrences.

```
merconnectionmanager.cpp:        qWarning() << "VirtualMachine::update() - unknown state";
mersdkmanager.cpp:            qWarning() << "MerToolChain without target found. Removing toolchain.";
mersdkmanager.cpp:            qWarning() << "MerQtVersion without target found. Removing qtversion.";
mersdkmanager.cpp:                qWarning() << "Broken Mer kit found! Removing kit.";
mersdkmanager.cpp:        qWarning() << "Sysroot does not exist" << buildTarget.sysRoot.toString();
mersettings.cpp:        qWarning() << "The environment variable"

```

In `merconnectionmanager.cpp` you may rather remove the `default:` case with the warning. This is checked by the compiler nowadays.

Tomin1: Sure.

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -148,6 +152,27 @@ void MerHardwareDeviceWizardSelectionPage::handleTestConnectionClicked()
         goto end;
     }
 
+    m_ui->connectionTestLabel->setText(tr("Detecting SDK tools..."));
```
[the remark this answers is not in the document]
vigejolla: "SDK Client Tools" doesn't really say anything to the users. How about "SDK Device Tools", which would indicate that these are tools that are needed on the device in order for it to work with the SDK.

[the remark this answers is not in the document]
Tomin1: I can change it to whatever feels correct. SDK Device Tools seems like a good one and there is not too much space for long descriptions.

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -148,6 +152,27 @@ void MerHardwareDeviceWizardSelectionPage::handleTestConnectionClicked()
...
+    m_sdkClientToolsInstalled = detectSdkClientTools(sshParams, &m_connectionTestOk);
+    if (!m_connectionTestOk) {
+        m_ui->connectionTestLabel->setText(tr("Could not autodetect SDK tools"));
```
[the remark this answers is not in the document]
Tomin1: Do you mean that I should change all of the qWarnings to be shown in connectionTestLabel instead (and adjust the text of course if needed)?

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -148,6 +152,27 @@ void MerHardwareDeviceWizardSelectionPage::handleTestConnectionClicked()
...
+        goto end;
+    }
+    if (!m_sdkClientToolsInstalled) {
+        m_ui->connectionTestLabel->setText(tr("SDK tools are not installed"));
+        QString password;
+        if (askInstallSdkClientTools(&password)) {
+            m_ui->connectionTestLabel->setText(tr("Installing SDK tools..."));
+            m_sdkClientToolsInstalled = installSdkClientTools(sshParams, password);
+            if (!m_sdkClientToolsInstalled) {
+                m_ui->connectionTestLabel->setText(tr("SDK tools install failed"));
```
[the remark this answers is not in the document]
Tomin1: That could be done if I change those methods to not to use static (like @vigejolla suggested earlier) and then set the messages inside them.

src/plugins/mer/merhardwaredevicewizardpages.cpp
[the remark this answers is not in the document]
martyone: I agree with the space constraints but I need to stress again than we do not really need a term, and if we do not need it, I would prefer not introducing it - it's not for for free.

src/plugins/mer/merhardwaredevicewizardpages.cpp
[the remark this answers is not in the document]
martyone: yes please. Feel free to improve the UI as well, as there is the space constraint right now.

src/plugins/mer/merhardwaredevicewizardpages.cpp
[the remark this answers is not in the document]
martyone: Or you can pass `QString *errorString`. We may be moving these pieces to Sfdk lib soon, so keeping the functions static may be more convenient.

src/plugins/mer/merhardwaredevicewizardpages.cpp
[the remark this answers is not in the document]
Tomin1: So do you want it to be changed or not? I didn't quite understand.

We could also have a more descriptive text above the test button if needed. Something like "You may be prompted to install software packages required for SDK operation such as deploying packages on device" and then just call it SDK tools or SDK packages elsewhere.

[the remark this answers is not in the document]
martyone: Sorry :) I prefer saying "{Detecting|Installing|Failed to install|...} software packages required for proper SDK operation...", including additional information in case of failure, and adjust the UI to remove the space constraint - moving the label below the button and using multiple lines?

src/plugins/mer/merhardwaredevicewizardpages.cpp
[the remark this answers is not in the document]
Tomin1: With error dialog (QMessageBox::critical) I could fit quite a bit more information there.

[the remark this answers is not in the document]
martyone: Yup

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -138,34 +139,40 @@ void MerHardwareDeviceWizardSelectionPage::handleTestConnectionClicked()
...
     m_ui->connectionTestLabel->setText(MerConnectionManager::testConnection(sshParams,
                 &m_connectionTestOk));
     if (!m_connectionTestOk)
         goto end;
 
     m_ui->connectionTestLabel->setText(tr("Detecting device properties..."));
-    m_architecture = detectArchitecture(sshParams, &m_connectionTestOk);
-    m_deviceName = detectDeviceName(sshParams, &m_connectionTestOk);
+    m_architecture = detectArchitecture(sshParams, &m_connectionTestOk, &error);
+    if (m_connectionTestOk) {
```
martyone: Not needed, the functions already check this to allow chaining without checking each result separately

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -138,34 +139,40 @@ void MerHardwareDeviceWizardSelectionPage::handleTestConnectionClicked()
     sshParams.timeout = timeout();
     sshParams.authenticationType = SshConnectionParameters::AuthenticationTypeAll;
 
+    QString error;
```
martyone: Nitpick, more often the name `errorMessage` is used.

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -138,34 +139,40 @@ void MerHardwareDeviceWizardSelectionPage::handleTestConnectionClicked()
...
     m_ui->connectionTestLabel->setText(tr("Detecting device properties..."));
-    m_architecture = detectArchitecture(sshParams, &m_connectionTestOk);
-    m_deviceName = detectDeviceName(sshParams, &m_connectionTestOk);
+    m_architecture = detectArchitecture(sshParams, &m_connectionTestOk, &error);
+    if (m_connectionTestOk) {
+        m_deviceName = detectDeviceName(sshParams, &m_connectionTestOk, &error);
+    }
     if (!m_connectionTestOk) {
         m_ui->connectionTestLabel->setText(tr("Could not autodetect device properties"));
+        QMessageBox::critical(this, tr("Device Error"), error);
```
martyone: I think the "Could not autodetect device properties" message should figure in the message box too, and it should be the main text while the content of `error` should be the `QMessageBox::informativeText`.

martyone: The "Device Error" message sounds a bit drastic :) How about "Connection test failed"?

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -138,34 +139,40 @@ void MerHardwareDeviceWizardSelectionPage::handleTestConnectionClicked()
...
         goto end;
     }
     if (!m_sdkClientToolsInstalled) {
-        m_ui->connectionTestLabel->setText(tr("SDK tools are not installed"));
+        m_ui->connectionTestLabel->setText(tr("Software packages required for proper SDK operation missing"));
         QString password;
         if (askInstallSdkClientTools(&password)) {
-            m_ui->connectionTestLabel->setText(tr("Installing SDK tools..."));
-            m_sdkClientToolsInstalled = installSdkClientTools(sshParams, password);
+            m_ui->connectionTestLabel->setText(tr("Installing software packages..."));
```
martyone: I wouldn't worry about repeating the full "software packages required for proper SDK operation" phrase in every message. There is no context permanently available that could help the user guess the missing part from.

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -201,14 +208,14 @@ Abi::Architecture MerHardwareDeviceWizardSelectionPage::detectArchitecture(
...
-        qCWarning(Log::mer) << "Failed to execute uname on target";
+        *error = tr("Failed to detect architecture: Command 'uname' could not be executed");
         return Abi::UnknownArchitecture;
     }
 
     const QString output = QString::fromLatin1(runner.readAllStandardOutput()).trimmed();
     if (output.isEmpty()) {
         *ok = false;
-        qCWarning(Log::mer) << "Empty output from uname executed on target";
+        *error = tr("Failed to detect architecture: Empty output from uname");
```
martyone: I would prefer "...from the 'uname' command" or so. Similar applies on other places.

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -263,12 +270,8 @@ QString MerHardwareDeviceWizardSelectionPage::detectDeviceName(
 }
 
 bool MerHardwareDeviceWizardSelectionPage::detectSdkClientTools(
-        const SshConnectionParameters &sshParams, bool *ok)
+        const SshConnectionParameters &sshParams, bool *ok, QString *error)
 {
-    if (!*ok) {
-        return false;
-    }
-
```
martyone: That's what enabled chaining without checking each result separately.

src/plugins/mer/merhardwaredevicewizardpages.cpp
Tomin1: Yeah, you are right. At first I thought that I would need to do this so I don't overwrite *error but actually it works like that just fine.

src/plugins/mer/merhardwaredevicewizardpages.cpp
Tomin1: I think in this case user must have just approved installation of those packages so they must be aware that it's "software packages required for proper SDK operation" they were installing. That's why I decided to leave it out and it's a bit too long to fit on one line so that's why I tried to shorten it.

martyone: True :) Please check if the other cases would profit from full wording (I didn't)

src/plugins/mer/merhardwaredevicewizardpages.cpp
Tomin1: That's better than errorString suggested above. :wink: (QString errorString felt a bit redundant.)

src/plugins/mer/merhardwaredevicewizardpages.cpp
Tomin1: I think it's now quite clear for the user. Error messages and the install approval dialog contain the full wording.

src/plugins/mer/merhardwaredevicewizardpages.cpp
martyone: yeah I am a bit messing with these two here and there, but since now I will remember I am sure :)

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -135,20 +139,49 @@ void MerHardwareDeviceWizardSelectionPage::handleTestConnectionClicked()
...
         goto end;
     }
 
+    m_ui->connectionTestLabel->setText(tr(
+        "Detecting software packages required for proper SDK operation..."));
+    m_sdkClientToolsInstalled = detectSdkClientTools(sshParams, &m_connectionTestOk, &errorMessage);
+    if (!m_connectionTestOk) {
+        m_ui->connectionTestLabel->setText(tr("Could not autodetect installed software packages"));
+        showErrorMessageDialog(tr(
+            "Could not autodetect installed software packages for proper SDK operation."), errorMessage);
```
martyone: I think the word "required" is missing here.

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -135,20 +139,49 @@ void MerHardwareDeviceWizardSelectionPage::handleTestConnectionClicked()
...
+        goto end;
+    }
+    if (!m_sdkClientToolsInstalled) {
+        m_ui->connectionTestLabel->setText(tr(
+            "Software packages required for proper SDK operation missing"));
```
martyone: I would write "*are* missing" in this case, what do you think?

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -135,20 +139,49 @@ void MerHardwareDeviceWizardSelectionPage::handleTestConnectionClicked()
...
+        QString password;
+        if (askInstallSdkClientTools(&password)) {
+            m_ui->connectionTestLabel->setText(tr("Installing software packages..."));
+            m_sdkClientToolsInstalled = installSdkClientTools(sshParams, password, &errorMessage);
+            if (!m_sdkClientToolsInstalled) {
+                m_ui->connectionTestLabel->setText(tr("Software package installation failed"));
+                showErrorMessageDialog(
+                    tr("Installation of software packages for proper SDK operation failed."), errorMessage);
```
martyone: +required?

src/plugins/mer/merhardwaredevicewizardpages.cpp
Tomin1: Well spotted.

src/plugins/mer/merhardwaredevicewizardpages.cpp
Tomin1: Yeah, that's better and fits still. (At least on my computer!)

PR #235, opened 2019-12-04
Jb48335

martyone:


martyone:

martyone:
1 is the minimum value enforced by the spin box

PR #239, opened 2019-12-11
Sfdk: Document path conversion issues under MSYS2

martyone:


vigejolla:
I wonder if it would be possible to detect when the path conversion is happening (e.g. if parameters have "C:\" in them) and warn users in this case, possibly hint about setting the environment variable?

martyone:
> I wonder if it would be possible to detect when the path conversion is happening (e.g. if parameters have "C:\" in them) and warn users in this case, possibly hint about setting the environment variable?

I though about it too but I am afraid it couldn't be simply applied to all arguments, so I would rather not do that. Another idea was we could grep the output of a remote process for windows paths - this would be possible to keep on one place but again it wouldn't work in every case.

PR #240, opened 2019-12-11
Docker build engine on Windows

vigejolla:
Note: Only the last commit belongs to this PR

### DIFF
--- a/src/libs/sfdk/dockervirtualmachine.cpp
+++ b/src/libs/sfdk/dockervirtualmachine.cpp
@@ -28,6 +28,7 @@
 #include "sfdkconstants.h"
 #include "usersettings_p.h"
 
+#include <utils/environment.h>
 #include <utils/fileutils.h>
 #include <utils/optional.h>
 #include <utils/qtcassert.h>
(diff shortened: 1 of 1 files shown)


src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -83,12 +84,20 @@ QString dockerPath()
     if (!path.isNull())
         return path;
 
+    QString dockerBinary = DOCKER;
+    QString pathVariable = "PATH";
+    QChar pathSeparator = ':';
```
martyone: QDir::listSeparator()

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -83,12 +84,20 @@ QString dockerPath()
...
+    if (Utils::HostOsInfo::isWindowsHost()) {
+        dockerBinary.append(".exe");
```
martyone: QTC_HOST_EXE_SUFFIX

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -83,12 +84,20 @@ QString dockerPath()
...
+        pathVariable = "Path";
```
martyone: I think this is not needed. The docs for QProcessEnvironment says "On Windows, the variable names are case-insensitive, but case-preserving. QProcessEnvironment behaves accordingly." and looging at QProcessEnvKey in qproess_p.h it really seems to be so.

martyone:
Now I noticed `Utils::Environment::systemEnvironment().searchInPath(executable)` is used elsewhere that does all the necessary tricks.

vigejolla:
> Now I noticed `Utils::Environment::systemEnvironment().searchInPath(executable)` is used elsewhere that does all the necessary tricks.

nice

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -28,7 +28,9 @@
 #include "sfdkconstants.h"
 #include "usersettings_p.h"
 
+#include <utils/environment.h>
 #include <utils/fileutils.h>
+#include <utils/hostosinfo.h>
```
martyone: Is this still needed?

vigejolla: nope. removed.

PR #244, opened 2019-12-19
Jb48045

martyone:


martyone:
Added a merge commit following github's message conventions, so that the build does not record sha1 of a commit from a side branch.

vigejolla:
So.. You merged this already, but have the PR still open? How does this work?

martyone:
The merge commit is in the source branch. It will be then merged by fast-forwarding master, so the merge commit will appear directly on master :)

PR #248, opened 2020-01-21
Jb45576

martyone:


martyone:

src/tools/sfdk/task.cpp
```diff
@@ -406,6 +420,11 @@ BOOL WINAPI SignalHandler::ctrlHandler(DWORD fdwCtrlType)
     switch (fdwCtrlType)
     {
     case CTRL_C_EVENT:
+        if (TaskManager::ctrlCFilter() && !TaskManager::ctrlCFilter()()) {
+            qCDebug(sfdk) << "Ctrl-C event filtered out";
+            return TRUE;
+        }
+        Q_FALLTHROUGH();
```
vigejolla: I was able to find only one place where ctrclCFilter was set, and ther it was set to IgnoreAllCtrlCEvents. This seems like way over complicated way of checking whether ctrl c events should be filtered or not - please consider using just a simple bool.

martyone: Yup, initially it seemed that more would be needed and after all it turned to be this easy. This is an area where issues are to be expected, so I prefer keeping the api this flexible.

PR #249, opened 2020-01-22
Jb45990

jlehtoranta:


src/tools/merssh/cmakecommand.cpp
```diff
@@ -0,0 +1,95 @@
...
+{
+    return QLatin1String("cmake");
+}
+
+int CMakeCommand::execute()
+{
+    if (arguments().contains(QLatin1String("--version"))
+        || arguments().contains(QLatin1String("--help"))
+        || arguments().contains(QLatin1String("-E capabilities")))
+        m_cacheFile = QLatin1String(Sfdk::Constants::CMAKE_QUERY_CACHE);
```
vigejolla: You forgot to add sfdkconstants.h to the commit

src/tools/merssh/cmakecommand.cpp
```diff
@@ -0,0 +1,95 @@
...
+                    sharedHomePath() +
+                    QLatin1String("\" ") +
+                    QLatin1String("MER_SSH_HOST_SHARED_SRC=\"") +
+                    sharedSourcePath() +
+                    QLatin1String("\"; "));
+
+    MerRemoteProcess process;
+    process.setSshParameters(sshParameters());
+    process.setCommand(command);
+    return process.executeAndWait();
```
vigejolla: Quite a lot of this stuff is identical to QMakeCommand. Please consider refactoring so that the identical parts would be implemented in Command instead.

PR #250, opened 2020-01-22
Jb48360

jlehtoranta:
Only the last 3 commits belong to this PR

### DIFF
--- a/src/plugins/mer/mersdkmanager.cpp
+++ b/src/plugins/mer/mersdkmanager.cpp
@@ -40,6 +40,9 @@
 #include <sfdk/virtualmachine.h>
 
 #include <app/app_version.h>
+#include <cmakeprojectmanager/cmakekitinformation.h>
+#include <cmakeprojectmanager/cmaketool.h>
+#include <cmakeprojectmanager/cmaketoolmanager.h>
 #include <coreplugin/coreconstants.h>
 #include <coreplugin/icore.h>
 #include <coreplugin/messagemanager.h>
@@ -66,6 +69,7 @@
 #include <QMenu>
 #include <QProcess>
 
+using namespace CMakeProjectManager;
 using namespace Debugger;
 using namespace ExtensionSystem;
 using namespace Mer::Constants;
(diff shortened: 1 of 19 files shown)


src/plugins/mer/mersdkmanager.cpp
```diff
@@ -135,11 +139,13 @@ void MerSdkManager::initialize()
...
 
-        // If debugger became available
+        // If debugger and/or cmake became available
         for (BuildEngine *const engine : Sdk::buildEngines()) {
             for (const QString &targetName : engine->buildTargetNames()) {
-                for (Kit *const kit : kitsForTarget(targetName))
+                for (Kit *const kit : kitsForTarget(targetName)) {
                     ensureDebuggerIsSet(kit, engine, targetName);
+                    ensureCmakeToolIsSet(kit, engine, targetName);
+                }
```
martyone: I wonder what makes cmake more similar to debugger than qmake w.r.t the way it is handled here?

src/plugins/mer/mersdkmanager.cpp
```diff
@@ -43,6 +43,9 @@
 #include <coreplugin/coreconstants.h>
 #include <coreplugin/icore.h>
 #include <coreplugin/messagemanager.h>
+#include <cmakeprojectmanager/cmakekitinformation.h>
+#include <cmakeprojectmanager/cmaketool.h>
+#include <cmakeprojectmanager/cmaketoolmanager.h>
```
vigejolla: Nitpick: Shouldn't these be before coreplugin includes?

src/tools/merssh/command.cpp
```diff
@@ -44,6 +44,18 @@ int Command::executeRemoteCommand(const QString &command)
     // Convert to build engine paths
     QString mappedCommand = remotePathMapping(command);
 
+    // Share host mount points
+    mappedCommand.prepend(QLatin1String("export ") +
+                    QLatin1String("MER_SSH_HOST_SHARED_HOME=") +
```
vigejolla: IIRC, we should no longer use QLatin1String

src/tools/merssh/cmakecommand.cpp
```diff
@@ -0,0 +1,80 @@
...
+{
+    return QLatin1String("cmake");
+}
+
+int CMakeCommand::execute()
+{
+    if (arguments().contains(QLatin1String("--version"))
+            || arguments().contains(QLatin1String("--help"))
+            || arguments().contains(QLatin1String("-E capabilities"))) {
+        m_cacheFile = QLatin1String(Sfdk::Constants::CMAKE_QUERY_CACHE);
```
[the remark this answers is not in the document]
martyone: This is gone now

src/tools/merssh/command.cpp
martyone: This is gone now.

PR #252, opened 2020-01-22
Jb48660

jlehtoranta:
Only the last 2 commits belong to this PR

### DIFF
--- a/src/plugins/mer/merdeployconfiguration.cpp
+++ b/src/plugins/mer/merdeployconfiguration.cpp
@@ -1,7 +1,7 @@
 /****************************************************************************
 **
 ** Copyright (C) 2012-2015,2017-2019 Jolla Ltd.
-** Copyright (C) 2019 Open Mobile Platform LLC.
+** Copyright (C) 2020 Open Mobile Platform LLC.
 ** Contact: http://jolla.com/
 **
 ** This file is part of Qt Creator.
@@ -69,7 +69,7 @@ QString MerRpmDeployConfigurationFactory::displayName()
 
 Core::Id MerRpmDeployConfigurationFactory::configurationId()
 {
-    return Core::Id("QmakeProjectManager.MerRpmDeployConfiguration");
+    return Core::Id("Mer.MerRpmDeployConfiguration");
 }
 
 ////////////////////////////////////////////////////////////////////////////////////////////
@@ -87,7 +87,7 @@ QString MerRsyncDeployConfigurationFactory::displayName()
 
 Core::Id MerRsyncDeployConfigurationFactory::configurationId()
 {
-    return Core::Id("QmakeProjectManager.MerRSyncDeployConfiguration");
+    return Core::Id("Mer.MerRSyncDeployConfiguration");
 }
 
 /////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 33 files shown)


src/plugins/mer/merdeployconfiguration.cpp
```diff
@@ -178,11 +177,16 @@ void MerAddRemoveSpecialDeployStepsProjectListener::updateProject(QmakeProject *
...
 }
 
-bool MerAddRemoveSpecialDeployStepsProjectListener::isAmbienceProject(QmakeProject *project)
+bool MerAddRemoveSpecialDeployStepsProjectListener::isAmbienceProject(Project *project)
 {
-    QmakeProFile *root = project->rootProFile();
-    return root->projectType() == ProjectType::AuxTemplate &&
-        root->variableValue(Variable::Config).contains(QLatin1String(SAILFISH_AMBIENCE_CONFIG));
+    if (qobject_cast<QmakeProject*>(project)) {
+        QmakeProFile *root = qobject_cast<QmakeProject*>(project)->rootProFile();
```
martyone: Possibly
```
if (auto *qmakeProject = qobject_cast<QmakeProject*>(project)) {
    QmakeProFile *root = qmakeProject->rootProFile();
```

src/plugins/mer/merdeploysteps.cpp
```diff
@@ -223,9 +225,17 @@ bool MerProcessStep::init()
...
+    if (qobject_cast<CMakeBuildConfiguration*>(buildConfiguration())) {
+        bc = qobject_cast<CMakeBuildConfiguration*>(buildConfiguration());
+        if (!bc)
+            bc =  qobject_cast<CMakeBuildConfiguration*>(target()->activeBuildConfiguration());
+    } else {
+        // Default to QmakeBuildConfiguration
+        bc = qobject_cast<QmakeBuildConfiguration*>(buildConfiguration());
+        if (!bc)
+            bc =  qobject_cast<QmakeBuildConfiguration*>(target()->activeBuildConfiguration());
+    }
```
martyone: This line is unreachable I think:

`bc =  qobject_cast<CMakeBuildConfiguration*>(target()->activeBuildConfiguration());`

It would be also wort to think if resorting to `target()->activeBuildConfiguration()` is really needed here as `buildConfiguration` already returns that as a fallback, it just doesn't do the casting. Maybe this changed at some point and we can simplify it now here.

src/plugins/mer/merdeployconfiguration.cpp
```diff
@@ -69,7 +69,7 @@ QString MerRpmDeployConfigurationFactory::displayName()
 
 Core::Id MerRpmDeployConfigurationFactory::configurationId()
 {
-    return Core::Id("QmakeProjectManager.MerRpmDeployConfiguration");
+    return Core::Id("Mer.MerRpmDeployConfiguration");
```
martyone: I would consider moving this renames into a separate commit.

src/plugins/mer/merdeployconfiguration.cpp
```diff
@@ -69,7 +69,7 @@ QString MerRpmDeployConfigurationFactory::displayName()
```
vigejolla: Remember to update the copyright year :)

src/plugins/mer/merdeployconfiguration.cpp
vigejolla: I would not recommend assignment inside if condition. 

```
auto *qmakeProject = qobject_cast<QmakeProject*>(project);
if (qmakeProject) ...
```

should be preferred.

PR #253, opened 2020-01-29
Jb48741

martyone:


martyone:
Just the very last 4 commits belong to this PR

src/libs/sfdk/buildengine.cpp
```diff
@@ -914,6 +915,9 @@ BuildEngineManager::BuildEngineManager(QObject *parent)
     Q_ASSERT(!s_instance);
     s_instance = this;
 
+    QHostInfo::lookupHost(QHostInfo::localHostName(), this,
+            &BuildEngineManager::completeHostNameLookup);
```
vigejolla: QHostInfo::localHostName already returns the local hostname, and we are not interested in the IP address, so what's the point of the lookup?

src/plugins/mer/mergeneraloptionswidget.cpp
```diff
@@ -45,9 +47,13 @@ MerGeneralOptionsWidget::MerGeneralOptionsWidget(QWidget *parent)
...
-    m_ui->environmentFilterOverriddenLabel->setVisible(MerSettings::isEnvironmentFilterFromEnvironment());
-    m_ui->environmentFilterOverriddenLabel->setText(m_ui->environmentFilterOverriddenLabel->text()
-            .arg(Constants::SAILFISH_SDK_ENVIRONMENT_FILTER));
+    m_ui->environmentFilterInfoLabel->setVisible(MerSettings::isEnvironmentFilterFromEnvironment());
+    m_ui->environmentFilterInfoLabel->setPixmap(Utils::Icons::WARNING.pixmap());
+    m_ui->environmentFilterInfoLabel->setToolTip(
+            QLatin1String("<font color=\"red\">")
+            + tr("This option is currently overridden with the %1 environment variable.")
+                .arg(Constants::SAILFISH_SDK_ENVIRONMENT_FILTER)
+            + QLatin1String("</font>"));
```
vigejolla: It looks like you did a bit more than just shuffle widgets - perhaps the commit message could be improved?

src/tools/sfdk/command.cpp
```diff
@@ -79,9 +79,16 @@ class PropertiesAccessor
 {
     Q_DECLARE_TR_FUNCTIONS(Sfdk::PropertiesAccessor)
 
+public:
+    enum PrepareResult {
+        Ignored = 0,
+        Prepared,
+        Failed,
+    };
+
```
vigejolla: Again I feel the commit message is misleading: All you did here was change the return value of prepareSet from simple bool to enum.

src/plugins/mer/mergeneraloptionswidget.cpp
martyone: I would not complicate it, shuffling is the main point and shuffling one label into icon tooltip is just small part of it :)

src/libs/sfdk/buildengine.cpp
martyone: localHostName returns just the unqualified host name, lookup needs to be done to get the possible FQDN.

src/tools/sfdk/command.cpp
martyone: It was really done in order to allow use of multiple properties accessors in a chain.

src/libs/sfdk/buildengine.cpp
vigejolla: I'm not sure getting the fully qualified name is worth the extra hassle, but ok... :)

src/tools/sfdk/command.cpp
vigejolla: A good commit message would explain _what_ was changed and _why_. You have only explained the _why_ part, and not _what_.

martyone: More precisely it should be *what*, *why* and *how*. The commit title actually explains the *what*, not *why*. The *why* in this particular case is sufficiently explained by the fact that the commit precedes another commit(s) in a pull request. A good commit message should only try to explain *how* if the change is not easy to understand, which, according to your original comment, is not the case. I suggest to keep it as is.  :)

vigejolla: No, it does not (explain _what_). The _what_ in this case would be "Change the type of prepareSet return value". _why_ is "... in order to allow to chain PropertiesAccessor results". I agree that _how_ in this case is easy enough to understand.

martyone: - what: Allow to chain PropertiesAccessor results
- why: Because something later in the PR needs that (clear)
- how: Change the type of prepareSet return type... (clear)

Let's keep the bikes outside ;)

vigejolla: Eh. Sorry, but the answer to the question _what was changed_ is really "The type of prepareSet return value".

martyone: I wonder why is the title of this commit https://github.com/sailfishos/sailfish-qtcreator/pull/223/commits/023c3d57d0662073ea0830cbc97539992aa9d584 "Mer: Support targets on other build engines besides vbox", instead of something like "Fix MerSdkManager::kit by ...".

src/libs/sfdk/buildengine.cpp
```diff
@@ -950,6 +954,47 @@ QString BuildEngineManager::installDir()
...
 
+QString BuildEngineManager::defaultBuildHostName()
+{
+    if (!s_instance->m_defaultBuildHostName.isNull())
+        return s_instance->m_defaultBuildHostName;
+
+    // A non-blocking lookup was initiated during initialization, but it takes
+    // longer than anticipated - let's be patient.
+
+    QTime time;
```
vigejolla: time is only used for debug message. I wonder if it's possible to have it instantiated only if debug messages are enabled?

src/tools/sfdk/command.cpp
```diff
@@ -453,27 +455,73 @@ class BuildEnginePropertiesAccessor : public PropertiesAccessor
...
     }
 
     bool set() override
     {
-        return m_vmAccessor->set();
+        return m_vmAccessor->set() && setOther();
+    }
+
+private:
+    QMap<QString, QString> getOther() const
```
vigejolla: Now this is a weird name for a method. What other? Looks like m_hostName is the only "other" that this methods gets? What's the whole purpose of this method?

src/libs/sfdk/buildengine.cpp
martyone: That would be clearly a premature optimization :)

src/tools/sfdk/command.cpp
martyone: Really, it should be "others" - there is none now but more to be added in future. Will fix.

martyone: s/none/one/

PR #256, opened 2020-02-07
Sfdk: Use explicit cast to QString

vigejolla:


vigejolla:
Jenkins refused to build because there was the implicit cast from const char* to QString. On my laptop the original version compiled nicely though

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -579,15 +579,15 @@ QString DockerVirtualMachinePrivate::labelFor(VirtualMachinePrivate::SharedPath
 {
     switch (which) {
     case SharedSrc:
-        return Constants::BUILD_ENGINE_SHARED_SRC;
+        return QString(Constants::BUILD_ENGINE_SHARED_SRC);
```
martyone: Please try with just `{...}`.

vigejolla: Yup, works.

PR #258, opened 2020-02-20
Sfdk: Hardcode IPv4 address for build engine and emulator

vigejolla:
Using just the name localhost caused at least git ssh to try to
connect first via IPv6, causing unwanted delays.

### DIFF
--- a/src/libs/sfdk/buildengine.cpp
+++ b/src/libs/sfdk/buildengine.cpp
@@ -542,6 +542,7 @@ void BuildEnginePrivate::syncWwwProxy()
 
     QStringList args;
     const QString eq("=");
+    args.append(Constants::MER_SSH_HOST + eq + sshParameters.host());
     args.append(Constants::MER_SSH_PORT + eq + QString::number(sshParameters.port()));
     args.append(Constants::MER_SSH_PRIVATE_KEY + eq + sshParameters.privateKeyFile);
     args.append(Constants::MER_SSH_SHARED_HOME + eq + sharedHomePath.toString());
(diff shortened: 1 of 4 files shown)


vigejolla:
In my testing, this improved the build time for the template project from over 3 minutes to just 43 seconds.

martyone:
Nice. The two constants here are just the fallback values common to all build engine and emulator implementations, some of which could even use other than loopback addresses. The source of the actual address is the installer, passing it via sdktool, which is broken now, and that is the bug that needs to be fixed in first line.

src/libs/sfdk/buildengine.h
```diff
@@ -87,6 +87,8 @@ class SFDK_EXPORT BuildEngine : public QObject
     void setSharedSrcPath(const Utils::FileName &sharedSrcPath, const QObject *context,
             const Functor<bool> &functor);
 
+    QString sshHost() const;
```
martyone: hm, there is still a big question mark hanging above the properties that primarily exists on VirtualMachine side but are exposed either on BuildEngine or both VirtualMachine and BuildEngine. There is e.g. "FIXME internal" on VirtualMachine::sshParameters suggesting that exposing it on BuildEngine would be more correct but this is still to be investigated.

So for now I would prefer keeping the existing API as is and use `engine->virtualMachine()->sshParameters().host()` where we need the host.

vigejolla:
> How about the address displayed in options? I didn't try with these changes but the version I run displays "localhost" for the build engine. And how about emulator? Does it use what installer passes (if it does)?

The text "localhost" in options is just text - it has nothing to do with what is really used. That is a separate bug and I'd like to handle it as such.
As far as I understand, emulator uses the values passed from installer.

src/libs/sfdk/buildengine.h
vigejolla: I somewhat disagree, but fine, I'll change it

vigejolla: And to be clear about what I disagree with you about: I think we should either expose port and host or expose neither. And I think the best option would be what you already suggested: expose sshParameters.

martyone: Yup, it's just a bit disrupted by the fact that there is (still) the wwwPort too, so I am not totally sure yet the final state will be with sshParameters exposed here.

PR #259, opened 2020-02-20
Jb44416

martyone:


martyone:
The clang code model needs to know the additional qmake args pretty accurately.

src/plugins/mer/merbuildconfiguration.cpp
```diff
@@ -89,9 +90,15 @@ MerBuildConfiguration::MerBuildConfiguration(Target *target, Core::Id id)
         if (!isReallyActive())
             return;
         QTC_ASSERT(project(), return);
-        if (!project()->files(Project::AllFiles).contains(document->filePath())
-                || !document->filePath().toString().contains(QRegExp("\\.spec$|\\.yaml$")))
+        if (!document->filePath().toString().contains(QRegExp("\\.spec$|\\.yaml$")))
             return;
+        if (!project()->files(Project::AllFiles).contains(document->filePath())) {
+            if (document->filePath().isChildOf(project()->rootProjectDirectory())) {
+                MessageManager::write(tr("Warning: RPM *.spec (or *.yaml) file not registered to the project."),
```
vigejolla: I don't understand, in which situation do we end up in here?

martyone: When the .spec/.yaml file is not referred by the .pro file.

vigejolla: In that case, wouldn't we already return from either !isReallyActive() or project() checks?

martyone: Nope. In this context `this` is `MerBuildConfiguration`, not `IDocument` - I guess that is what confused you.

src/plugins/mer/merbuildconfiguration.cpp
```diff
@@ -56,6 +56,9 @@ namespace {
 
 const char MER_VARIABLES_CACHE_FILENAME[] = ".mb2/qmake_variables.cache";
 
+// Avoid actions on temporary changes and asking questions too hastily
+const int UPDATE_EXTRA_PARSER_ARGUMENTS_DELAY_MS = 3000;
```
vigejolla: I think there are some cases (e.g. when opening a project) when we should not wait. It's rather annoying when the dialog interrupts you when you had already started e.g. opening another project or something.

martyone: Unfortunately there are  too many scenarios to consider. Opening a project is actualy one of those that need some sort of care - be it the simple delay or something more complicated. When a project is being opened there are temporary states when it is unconfigured or configured using the default kit or so.

Ideally we should avoid the necessity for asking anything here, but I don't have a better idea how to reliably collect the additional args now. We should think about this more a propose some future work.

vigejolla: I agree. I'll write it down on my own notes so that I can bring it up again in the future :)

PR #260, opened 2020-02-21
Implement JSON-based wizard to create QML-files for Sailfish OS project

yarfruct:
We implemented the JSON-based wizard to create the QML-files for the Sailfish OS projects. 

This wizard allows to:

1. Set up file name.
2. Add imports with local files and directories to the QML-file.
3. Add imports with external libraries (Qt Positioning and Qt Sensors) to the QML-file.
4. Set up the root QML-type for the file: `Item` or one of imported (like `PositionSource` from Qt Positioning or `Sensor` from Qt Sensors).

Pages for adding local and libraries imports are custom pages with buttons, dialogs and list widgets. To implement the pages we developed a separate Qt Creator plugin with name `SailfishOSWizards` that initializes and registers new pages with their types ids to use them inside `wizard.json`. 

Also we edit the `ComboBoxField` class. By default `ComboBoxField` does not allow to set list of items as a string, only as the JSON array. So, the field does not allow to pass the wizard variable to the `items` field. 

For example, following code is not working:

```json
{
    "name": "QmlBaseType",
    "trDisplayName": "Base type of QML document:",
    "type": "ComboBox",
    "data":
    {
           "items": "%{QmlLibrariesTypes}"
    }
}
```

Variable `QmlLibrariesTypes` containt list of QML-types in depend on the imported libraries and "%{QmlLibrariesTypes}" is a string for the `items` field. So, default `ComboBoxField` cannot be initialized in this case and whole wizard cannot be created also.

This is a reason why we customize `ComboBoxField` to use a string value as the `items` field.

### DIFF
--- a/share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
+++ b/share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
@@ -22,6 +22,11 @@
             "typeId": "File"
         },
         {
+            "trDisplayName": "Local Imports",
+            "trShortTitle": "Local Imports",
+            "typeId": "QmlLocalImports"
+        },
+        {
             "trDisplayName": "QML File Options",
             "trShortTitle": "QML File Options",
             "typeId": "Fields",
--- a/share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
+++ b/share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
@@ -5,7 +5,7 @@
     "category": "Sailfish OS",
     "trDescription": "Creates a QML file for Sailfish OS application project.",
     "trDisplayName": "QML File",
-    "trDisplayCategory": "Sailfish OS",
+    "trDisplayCategory": "%{SailfishOSFileWizardCategoryName}",
     "iconText": "QML",
     "enabled": "true",
 
--- a/share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
+++ b/share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
@@ -43,7 +43,7 @@
                    "type": "ComboBox",
                    "data":
                    {
-                       "items": [ "Item" ]
+                       "items": "%{QmlLibrariesTypes}"
                    }
                },
                {
--- a/share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
+++ b/share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
@@ -27,6 +27,11 @@
             "typeId": "QmlLocalImports"
         },
         {
+            "trDisplayName": "Import Libraries",
+            "trShortTitle": "Import Libraries",
+            "typeId": "ImportExternalLibraries"
+        },
+        {
             "trDisplayName": "QML File Options",
             "trShortTitle": "QML File Options",
             "typeId": "Fields",
--- a/share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
+++ b/share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
@@ -1,0 +1,71 @@
+{
+    "version": 1,
+    "supportedProjectTypes": [ ],
+    "id": "E.QmlFile",
+    "category": "Sailfish OS",
+    "trDescription": "Creates a QML file for Sailfish OS application project.",
+    "trDisplayName": "QML File",
+    "trDisplayCategory": "Sailfish OS",
+    "iconText": "QML",
+    "enabled": "true",
+
+    "options" : [
+        { "key": "TargetPath", "value": "%{Path}" },
+        { "key": "QmlFile", "value": "%{JS: Util.fileName('%{TargetPath}', 'qml')}" }
+    ],
+
+    "pages":
+    [
+        {
+            "trDisplayName": "Location",
+            "trShortTitle": "Location",
+            "typeId": "File"
+        },
+        {
+            "trDisplayName": "QML File Options",
+            "trShortTitle": "QML File Options",
+            "typeId": "Fields",
+            "data" :
+            [
+               {
+                   "name": "QmlBaseType",
+                   "trDisplayName": "Base type of QML document:",
+                   "type": "ComboBox",
+                   "data":
+                   {
+                       "items": [ "Item" ]
+                   }
+               },
+               {
+                   "name": "QtQuickVersion",
+                   "trDisplayName": "Qt Quick version:",
+                   "type": "ComboBox",
+                   "data":
+                   {
+                       "items": [ "2.6", "2.5", "2.4", "2.3", "2.2", "2.1", "2.0", "1.0" ]
+                   }
+               }
+            ]
+        },
+        {
+            "trDisplayName": "Project Management",
+            "trShortTitle": "Summary",
+            "typeId": "Summary"
+        }
+    ],
+
+    "generators":
+    [
+        {
+            "typeId": "File",
+            "data":
+            [
+                {
+                    "source": "template.qml",
+                    "target": "%{QmlFile}",
+                    "openInEditor": true
+                }
+            ]
+        }
+    ]
+}
--- a/src/plugins/projectexplorer/jsonwizard/jsonfieldpage.cpp
+++ b/src/plugins/projectexplorer/jsonwizard/jsonfieldpage.cpp
@@ -50,6 +50,8 @@
 #include <QStandardItem>
 #include <QItemSelectionModel>
 #include <QDir>
+#include <QJsonDocument>
+#include <QJsonArray>
 
 using namespace Utils;
 
@@ -869,17 +871,19 @@ bool ListField::parseData(const QVariant &data, QString *errorMessage)
                 .arg(type(), name());
         return false;
     }
-    if (value.type() != QVariant::List) {
+    if (value.type() == QVariant::List) {
+        for (const QVariant &i : value.toList()) {
+            std::unique_ptr<QStandardItem> item = createStandardItemFromListItem(i, errorMessage);
+            QTC_ASSERT(!item || !item->text().isEmpty(), continue);
+            m_itemList.emplace_back(std::move(item));
+        }
+    } else if (value.type() == QVariant::String) {
+        m_itemsString = value.toString();
+    } else {
         *errorMessage = QCoreApplication::translate("ProjectExplorer::JsonFieldPage",
-                                                    "%1 (\"%2\") \"items\" is not a JSON list.")
+                                                    "%1 (\"%2\") \"items\" is not a JSON list or string.")
                 .arg(type(), name());
         return false;
-    }
-
-    for (const QVariant &i : value.toList()) {
-        std::unique_ptr<QStandardItem> item = createStandardItemFromListItem(i, errorMessage);
-        QTC_ASSERT(!item || !item->text().isEmpty(), continue);
-        m_itemList.emplace_back(std::move(item));
     }
 
     warnAboutUnsupportedKeys(tmp, name(), type());
@@ -907,45 +911,18 @@ void ListField::initializeData(MacroExpander *expander)
         m_index = -1;
     }
 
-    QStandardItem *currentItem = m_index >= 0 ? m_itemList[uint(m_index)].get() : nullptr;
     QList<QStandardItem*> expandedValuesItems;
-    expandedValuesItems.reserve(int(m_itemList.size()));
 
-    for (const std::unique_ptr<QStandardItem> &item : m_itemList) {
-        bool condition = JsonWizard::boolFromVariant(item->data(ConditionRole), expander);
-        if (!condition)
-            continue;
-        QStandardItem *expandedValuesItem = item->clone();
-        if (item.get() == currentItem)
-            currentItem = expandedValuesItem;
-        expandedValuesItem->setText(expander->expand(item->text()));
-        expandedValuesItem->setData(expander->expand(item->data(ValueRole).toString()), ValueRole);
-        expandedValuesItem->setData(expander->expand(item->data(IconStringRole).toString()), IconStringRole);
-        expandedValuesItem->setData(condition, ConditionRole);
-
-        QString iconPath = expandedValuesItem->data(IconStringRole).toString();
-        if (!iconPath.isEmpty()) {
-            if (auto *page = qobject_cast<JsonFieldPage*>(widget()->parentWidget())) {
-                const QString wizardDirectory = page->value("WizardDir").toString();
-                iconPath = QDir::cleanPath(QDir(wizardDirectory).absoluteFilePath(iconPath));
-                if (QFileInfo::exists(iconPath)) {
-                    QIcon icon(iconPath);
-                    expandedValuesItem->setIcon(icon);
-                    addPossibleIconSize(icon);
-                } else {
-                    qWarning().noquote() << QString("Icon file \"%1\" not found.").arg(QDir::toNativeSeparators(iconPath));
-                }
-            } else {
-                qWarning().noquote() <<  QString("%1 (\"%2\") has no parentWidget JsonFieldPage to get the icon path.").arg(type(), name());
-            }
-        }
-        expandedValuesItems.append(expandedValuesItem);
+    if (m_itemsString.isEmpty()) {
+        expandedValuesItems = createStandardItemListFromItemList(expander);
+    } else {
+        expandedValuesItems = createStandardItemListFromItemsString(expander);
     }
 
     itemModel()->clear();
     itemModel()->appendColumn(expandedValuesItems); // inserts the first column
 
-    selectionModel()->setCurrentIndex(itemModel()->indexFromItem(currentItem), QItemSelectionModel::ClearAndSelect);
+    selectionModel()->setCurrentIndex(itemModel()->index(0, 0), QItemSelectionModel::ClearAndSelect);
 
     updateIndex();
 }
@@ -988,6 +965,57 @@ void ListField::updateIndex()
         selectionModel()->setCurrentIndex(itemModel()->index(m_savedIndex, 0), QItemSelectionModel::ClearAndSelect);
         m_savedIndex = -1;
     }
+}
+
+QList<QStandardItem *> ListField::createStandardItemListFromItemList(MacroExpander *expander)
+{
+    QList<QStandardItem*> expandedValuesItems;
+    expandedValuesItems.reserve(int(m_itemList.size()));
+
+    for (const std::unique_ptr<QStandardItem> &item : m_itemList) {
+        bool condition = JsonWizard::boolFromVariant(item->data(ConditionRole), expander);
+        if (!condition)
+            continue;
+        QStandardItem *expandedValuesItem = item->clone();
+        expandedValuesItem->setText(expander->expand(item->text()));
+        expandedValuesItem->setData(expander->expand(item->data(ValueRole).toString()), ValueRole);
+        expandedValuesItem->setData(expander->expand(item->data(IconStringRole).toString()), IconStringRole);
+        expandedValuesItem->setData(condition, ConditionRole);
+
+        QString iconPath = expandedValuesItem->data(IconStringRole).toString();
+        if (!iconPath.isEmpty()) {
+            if (auto *page = qobject_cast<JsonFieldPage*>(widget()->parentWidget())) {
+                const QString wizardDirectory = page->value("WizardDir").toString();
+                iconPath = QDir::cleanPath(QDir(wizardDirectory).absoluteFilePath(iconPath));
+                if (QFileInfo::exists(iconPath)) {
+                    QIcon icon(iconPath);
+                    expandedValuesItem->setIcon(icon);
+                    addPossibleIconSize(icon);
+                } else {
+                    qWarning().noquote() << QString("Icon file \"%1\" not found.").arg(QDir::toNativeSeparators(iconPath));
+                }
+            } else {
+                qWarning().noquote() <<  QString("%1 (\"%2\") has no parentWidget JsonFieldPage to get the icon path.").arg(type(), name());
+            }
+        }
+        expandedValuesItems.append(expandedValuesItem);
+    }
+
+    return expandedValuesItems;
+}
+
+QList<QStandardItem *> ListField::createStandardItemListFromItemsString(MacroExpander *expander)
+{
+    QList<QStandardItem *> standardItemList;
+    QVariantList items = QJsonDocument::fromJson(expander->expand(m_itemsString).toUtf8())
+                         .array().toVariantList();
+    for (QVariant item : items) {
+        QStandardItem *standardItem = new QStandardItem();
+        standardItem->setText(item.toString());
+        standardItem->setData(item, ValueRole);
+        standardItemList << standardItem;
+    }
+    return standardItemList;
 }
 
 void ComboBoxField::setup(JsonFieldPage *page, const QString &name)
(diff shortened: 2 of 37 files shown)


share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
```diff
@@ -0,0 +1,81 @@
+{
+    "version": 1,
+    "supportedProjectTypes": [ ],
+    "id": "E.QmlFile",
+    "category": "Sailfish OS",
+    "trDescription": "Creates a QML file for Sailfish OS application project.",
```
etaishev: It would be a good to think about using a parameterized OS name from `Sfdk::Sdk::osVariant()`

yarfruct: Do you mean for "category"?

etaishev: For all string `Sailfish OS`, `Sailfish SDK`, `Sailfish IDE` which are displayed in the interface. OS name can now be configured.

share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
```diff
@@ -40,10 +40,10 @@
                {
                    "name": "QmlBaseType",
                    "trDisplayName": "Base type of QML document:",
-                   "type": "ComboBox",
+                   "type": "StringComboBox",
                    "data":
                    {
-                       "items": [ "Item" ]
+                       "items": "%{QmlLibrariesTypes}"
```
martyone: I would consider changing upstream ComboBox so that it accepts string as well. IIUC now it already accepts list of strings and list of objects at least.

src/plugins/sailfishoswizards/data/qml-modules.json
```diff
@@ -0,0 +1,22 @@
...
+        "yamlheader": "Requires",
+        "specheader": "Requires",
+        "dependencies": ["sailfishsilica-qt5 >= 0.10.9", "qt5-qtdeclarative-import-sensors"],
+        "types": ["Accelerometer", "AccelerometerReading", "Altimeter", "AltimeterReading", "AmbientLightReading", "AmbientLightSensor", "AmbientTemperatureReading",
+        "AmbientTemperatureSensor", "Compass", "CompassReading", "DistanceReading", "DistanceSensor", "Gyroscope", "GyroscopeReading", "HolsterReading",
+        "HolsterSensor", "HumidityReading", "HumiditySensor", "IRProximityReading", "IRProximitySensor", "LidReading", "LidSensor", "LightReading", "LightSensor",
+        "Magnetometer", "MagnetometerReading", "OrientationReading", "OrientationSensor", "PressureReading", "PressureSensor", "ProximityReading", "ProximitySensor",
+        "RotationReading", "RotationSensor", "Sensor", "SensorGesture", "SensorGlobal", "SensorReading", "TapReading", "TapSensor", "TiltReading", "TiltSensor"]
+    }
+}
```
martyone: Available modules, their dependencies and types they provide depend on target OS version. We should allow to choose the target OS version and adjust the offered modules data based on that. Preferably with per-target configuration.

src/plugins/sailfishoswizards/include/models/externallibrary.h
```diff
@@ -0,0 +1,56 @@
...
+    void setYamlHeader(const QString &yamlHeader);
+    void setSpecHeader(const QString &specHeader);
+
+    QString getName() const;
+    QString getLayoutName() const;
+    QStringList getQtList() const;
+    QStringList getDependencyList() const;
+    QStringList getTypesList() const;
+    QString getYamlHeader() const;
+    QString getSpecHeader() const;
```
martyone: The part of updating .spec/.yaml seems to not be implemented yet. This would be also the most challenging part to abstract from to allow upstreaming. Ideas?

share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
yarfruct: Do you mean that we should change the standard `ComboBoxField` for using strings as items instead of creating custom `StringComboBoxField`?
And also what means IIUC?

src/plugins/sailfishoswizards/include/models/externallibrary.h
yarfruct: Methods for spec- and yaml-files are used inside other wizards for creating Yaml and Spec files. Here I forgot to remove them from the file. Do I should to remove them from this pull request?

share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
martyone: Yes. https://www.google.com/search?q=IIUC%20meaning

src/plugins/sailfishoswizards/include/models/externallibrary.h
martyone: Yes, if the code is only needed for future PRs then I would prefer only including it in the future PRs, i.e., removing it from this PR.

But my understanding was the .spec/.yaml would be updated when adding a new QML file that brings new dependencies. Hence my question if you have any idea how to abstract that part for upstreaming. Note on that, that it should be enough to only care about YAML and leave .spec untouched, as that one would be updated from the YAML on next build anyway.

vigejolla: Qt Creator can be used also on projects which only use .spec (and no .yaml), so it's not good to leave .spec untouched. Maybe there could be some logic, that if the project has .yaml file, then .spec could be left alone. But then again, that would just add unnecessary complications to the code, so I wouldn't do even that.

martyone: Leaving .spec untouched is perfectly OK, because the yaml-spec cooperation it is designed for that. The .spec is not completely overwritten based on the yaml. The yaml packaging format cover only subset of .spec. In advanced cases, the generated .spec is to be hand-edited and `specify` will preserve those modifications next time it updates it from yaml.

YAML packaging format exists exactly for the advantage of being more easily editable programmatically compared to .spec. It has been always the plan to develop wizards and similar tools that would manipulate the YAML only.

yarfruct:
@martyone @vigejolla @etaishev Sorry for offtop, but what version of Qt you use for building `sailfish-qtcreator`? Now we are using Qt 5.12.7 but we have a problem: app is crashed when we click "Preferences".
Host OS: macOS 10.13.6.

martyone:
We are using 5.12.5 https://github.com/sailfishos/sdk-build-tools/blob/master/defaults.sh#L53 . It may be compiler specific too - feel free to include a commit to add an #include there in your PR.

yarfruct:
Ok.

yarfruct:
@martyone Sorry, but a problem with opening "Preferences" is still here. App crashes without any error messages.
Maybe you have an idea about what can be wrong?

share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
yarfruct: @etaishev We checked possibility of setting the `"category"` and `"trDisplayCategory"` parameters as the wizard variables like `"%{SailfishOSCategory}"` that initialized inside C++ code using `Sfdk::Sdk::osVariant()` and it's impossible.

The `"category"` and `"trDisplayCategory"` parameters are used exactly the string that is passed instead of the variable value. So, in case you suggest user will see only "%{SailfishOSCategory}" instead of "Sailfish OS" as a category.

martyone: Check this https://github.com/sailfishos/sailfish-qtcreator/blob/master/share/qtcreator/templates/wizards/projects/vcs/git/wizard.json#L27 - I guess the indirection with "options" is needed.

martyone:
> @martyone Sorry, but a problem with opening "Preferences" is still here. App crashes without any error messages.
> Maybe you have an idea what can be wrong?

Hm, try with clean configuration ((re)move `~/.config/SailfishSDK` and also the pre-3.0 path `~/.config/SailfishOS-SDK` if it exists). If it does not help, open a new issue here to further discuss more. It will be nice if you attach a backtrace in that case.

yarfruct:
> > @martyone Sorry, but a problem with opening "Preferences" is still here. App crashes without any error messages.
> > Maybe you have an idea what can be wrong?
> 
> Hm, try with clean configuration ((re)move `~/.config/SailfishSDK` and also the pre-3.0 path `~/.config/SailfishOS-SDK` if it exists). If it does not help, open a new issue here to further discuss more. It will be nice if you attach a backtrace in that case.

Thanks. Removing `~/.config/SailfishOS-SDK` helps.

share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
yarfruct: The "options" object doesn't help too.
Such things as the "category" and "trDisplayCategory" fields are defined inside the [`IWizardFactory`](https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/plugins/coreplugin/iwizardfactory.h#L45) class in the `coreplugin` Qt Creator plugin. 
The class contains following declarations for the fields methods:

```cpp
QString category() const { return m_category; }
QString displayCategory() const { return m_displayCategory; }
...
void setDisplayName(const QString &displayName) { m_displayName = displayName; }
void setCategory(const QString &category) { m_category = category; }
```

Fields are set by exactly the string that is passed ("%{SailfishOSCategory}" in our case). There is no the [`MacroExpander`](https://github.com/sailfishos/sailfish-qtcreator/blob/188bf6777d3df8d4389c687119887f18fdecc564/src/libs/utils/macroexpander.h) instance that is used inside other wizards pages to retrieve the wizard variable value, for example, inside [`ListField::initializeData()`](https://github.com/sailfishos/sailfish-qtcreator/blob/d5ef6a3b84666e082556aa69b2afb73e4886f312/src/plugins/projectexplorer/jsonwizard/jsonfieldpage.cpp#L901) method.

src/plugins/sailfishoswizards/include/models/externallibrary.h
yarfruct: @martyone @vigejolla What if we will add dependencies to the YAML and/or SPEC files if a project is opened in Sailfish Qt Creator? But if a project is not opened this wizard will only create a new QML-file.

share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
martyone: We should try to improve upstream to support that then.

src/plugins/sailfishoswizards/include/models/externallibrary.h
martyone: Something like that makes sense, but it shouldn't be controlled by the fact that a project is open. There should be the option to (not) register the new file to a project and/or VCS on the last page. At least when I checked with some existing wizard there is such an option.

share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
yarfruct: So, I check this conversation as "resolved". Ok?

martyone: Not really - please read my previous message as "You should try..." :)

src/plugins/sailfishoswizards/include/models/externallibrary.h
yarfruct: @martyone Can you tell how we can write a code for editing the YAML (and SPEC) file after going through the whole wizard? 

We think that we should add dependencies when the user clicks "Ok" on the "Summary" page. Class `JsonSummaryPage` contains inside the `projectexplorer` plugin and we think that our code is too specific to add this into the standard `JsonSummaryPage`. Maybe we should to inherit the `JsonSummaryPage` class, implement the YAML file editing and use our class in the wizard? Or maybe do you have some another idea?

src/plugins/sailfishoswizards/data/qml-modules.json
yarfruct: Do you mean Sailfish OS version? But what a problem with different Sailfish OS versions?

martyone: The list of supported modules is not stable - some new are added, some old are dropped as times go. So every Sailfish OS version (and their variants) may be different in this respect.

yarfruct: So, where we can take an information about modules and Sailfish OS versions where they are available? And how to update this information when Sailfish OS will be updated?

src/plugins/sailfishoswizards/include/models/externallibrary.h
martyone: Nope, I don't have a better idea. Subclassing `JsonSummaryPage` seems to be the only option. (I would expect some generic mechanism being available for that, similar to that one for generating files, but I guess they had no use case for that and so the registration to project is done from `JsonSummaryPage`.)

src/plugins/sailfishoswizards/data/qml-modules.json
martyone: At this level it is not necessary to deal with that question. Just a mechanism needs to be implemented that will enable adding that information.

So e.g. instead of using single hardcoded `qml-modules.json` we could load all json files found under a well known directory (likely a subdirectory of the wizard directory)  and extend the json file format so that the  information about the OS version can be recorded there. Then, in the page where the modules are selected, there would be one more combox box to choose the desired OS version.

Later we would think how to collect the information about available modules for each supported OS version and supply that.

src/plugins/sailfishoswizards/include/models/externallibrary.h
martyone: Don't forget just YAML needs to be updated. Spec will receive the modifications automatically.

yarfruct: Another option is to create the whole wizard with all pages manually inside the `sailfishoswizards` plugin without using the `wizard.json` file. What about this idea?

> Don't forget just YAML needs to be updated. Spec will receive the modifications automatically.

But what about projects that use only SPEC files? Maybe they cannot be created by default, but users can remove YAML manually and use only SPEC.

martyone: We should definitely do json based wizards - upstream moved from C++ based wizards long time ago.

As I wrote earlier, one reason for supporting YAML packaging is to avoid the need to edit .spec programmatically. In that sense we should only care about packages that use YAML.

src/plugins/sailfishoswizards/data/qml-modules.json
vigejolla: In order to be both backwards and forwards compatible, we actually need two mechanisms: 1) For old OS versions, we need the functionality that martyone described above. To be futureproof, we should also have another mechanism, which fetches the information directly from the build target. But I think you shouldn't even try to implement it here and now.

share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
yarfruct: @martyone @etaishev We researched a possibility of improving upstream to support setting the wizard categry with wizard variable and it is impossible too.

Class `IWizardFactory` with methods to set up the category contains inside the `coreplugin` Qt Creator plugin. To set up new wizard variables we need the `JsonWizard` class instance and this instance should be initialized before initializing instance of `IWizardFactory` that is impossible because wizard factory are created earlier than any wizard. 

Also the `IWizardFactory` class and classes of JSON-based wizards are placed inside different plugins: `coreplugn` for `IWizardFactory` and `projectexplorer` for JSON-based wizards. The `projectexplorer` plugin already depends on `coreplugin` and we cannot make a loop of dependencies.

So, we cannot set up the wizard category with a variable. And we think it is not a problem. All standard wizard categories are set as simple translatable strings.

src/plugins/sailfishoswizards/include/models/externallibrary.h
yarfruct: > We should definitely do json based wizards - upstream moved from C++ based wizards long time ago.

So, in this case we need to add `PROJECTEXPLORER_EXPORT` macros to the `JsonSummaryPage` class declaration to have a possibility to inherit this class inside `sailfishoswizards` plugin. Can we make this change inside upstream?

src/plugins/sailfishoswizards/data/qml-modules.json
martyone: Yes, by "Later we would think how to collect the information about available modules for each supported OS version and supply that." I meant that another mechanism.

src/plugins/sailfishoswizards/include/models/externallibrary.h
martyone: The best is to ask upstream. Possibly getting some better suggestions from them :)

src/plugins/sailfishoswizards/data/qml-modules.json
yarfruct: Ok. We undestood about setting the Sailfish OS version and display available QML-types in depend on the version. But what versions should we display for user right now? And where we can find information about types differences between OS versions?

martyone: In scope of this PR it is not expected to supply any data about available modules. For testing you can use any fake data. Real data with be supplied by the "another mechanism". (The wizard should fail gracefully if no data is supplied)

share/qtcreator/templates/wizards/sailfishos-qml-file/wizard.json
martyone: "Impossible" is too strong word :) The display name and category from the json is processed here https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/plugins/projectexplorer/jsonwizard/jsonwizardfactory.cpp#L602 and that is the place where additional processing could happen I think. Maybe simply by processing it with a MacroExpander instantiated right there, right for that purpose. Macros for Sailfish SDK branding could be then registered with Utils::globalMacroExpander() if they are not yet.

src/plugins/sailfishoswizards/sailfishoswizardsplugin.cpp
```diff
@@ -30,6 +31,9 @@ bool SailfishOSWizardsPlugin::initialize(const QStringList &arguments, QString *
 {
     Q_UNUSED(arguments)
     Q_UNUSED(errorString)
+    Utils::globalMacroExpander()->registerVariable("SailfishOSFileWizardCategoryName",
+                                                   tr("Category of a wizard for Sailfish OS files"),
+                                                   []() { return Sfdk::Sdk::osVariant(); });
```
martyone: Yes, something like that, but
1. done from the SailfishOS plugin
2. only exporting the xxxVariant string
3. using the identifiers as here https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/tools/sfdk/dispatch.cpp#L66 but prefixed with "Sailfish" (e.g. "SailfishOsVariant") and then
4. having something like "Category of a wizard for %{SailfishOsVariant} files" in the json.

martyone: And 5. changing upstream code with separate commit to be sent upstream :)

yarfruct: 1 — So, where should we put the variable registration instead of the `sailfishoswizards` plugin?

3 — Do you mean to use `tr("Category of a wizard for %{SailfishOsVariant} files")` instead of `tr("Category of a wizard for Sailfish OS files")`?

5 — Do you mean a separate pull request? Because commit is already separate.

martyone: > 1 — So, where should we put the variable registration instead of the `sailfishoswizards` plugin?

Done from `MerPlugin::initialize()` should be OK

> 4 — Do you mean to use `tr("Category of a wizard for %{SailfishOsVariant} files")` instead of `tr("Category of a wizard for Sailfish OS files")`?

Sorry, my example was a bit confusing because I confused that message for the macro value, while it is actually the description of the macro. So the JSON side is OK (except the macro name), and the description argument to the `registerVariable` call should be change to as `tr("Category of a wizard for %1 files").arg(Sdk::osVariant())`.

> 5 — Do you mean a separate pull request? Because commit is already separate. Also, do you mean that it should be [only one change inside the `JsonWizardFactory` class](https://github.com/sailfishos/sailfish-qtcreator/pull/260/commits/b483f21d10cb3d223a18050cdb8c1af7809e5fdc#diff-dbc3af13b462fd690c75ef256107c694L614)? Or something else?

I mean separate commit with changes to upstream code, so that you can cherry-pick the one commit and submit it for upstream review. Also the registration of the macros should be done with a separate commit.

src/plugins/projectexplorer/jsonwizard/jsonfieldpage.cpp
```diff
@@ -869,19 +871,21 @@ bool ListField::parseData(const QVariant &data, QString *errorMessage)
                 .arg(type(), name());
         return false;
     }
-    if (value.type() != QVariant::List) {
```
martyone: All changes to upstream code should be done with separate commits and submitted for upstream review.

src/plugins/projectexplorer/jsonwizard/jsonwizardfactory.cpp
```diff
@@ -611,7 +611,7 @@ bool JsonWizardFactory::initialize(const QVariantMap &data, const QDir &baseDir,
         *errorMessage = tr("No displayCategory set.");
         return false;
     }
-    setDisplayCategory(strVal);
+    setDisplayCategory(Utils::globalMacroExpander()->expand(strVal));
```
martyone: I think we should do the same for displayName and description at least.

src/plugins/sailfishoswizards/sailfishoswizardsplugin.cpp
yarfruct: We added PR about setting a wizard info with variables: https://github.com/sailfishos/sailfish-qtcreator/pull/264

src/plugins/projectexplorer/jsonwizard/jsonwizardfactory.cpp
yarfruct: Done inside PR: https://github.com/sailfishos/sailfish-qtcreator/pull/264

src/plugins/sailfishoswizards/include/models/externallibrary.h
yarfruct: We added PR about adding the `PROJECTEXPLORER_EXPORT` macros to the `JsonSummaryPage` class: https://github.com/sailfishos/sailfish-qtcreator/pull/265

src/plugins/projectexplorer/jsonwizard/jsonfieldpage.cpp
yarfruct: Done inside PR: https://github.com/sailfishos/sailfish-qtcreator/pull/266

martyone: @yarfruct Note for your convenience, that I do not require that you create separate PRs here. I am OK with creating more PRs if you really find it more convenient, but I believe it would be less bothering for you having single PR here, with changes clearly separated into commits and submit selected commits for upstream adoption. When accepted by upstream, you would update the PR here with possible changes required by upstream and then the PR would be merged at once.

martyone: @yarfruct One more note that may not be obvious: your PRs are *not* going to to be delayed until we upgrade to the newer Qt Creator release that will contain your commits accepted by upstream. We can merge your PRs with the original commits immediately when those commits appear in an upstream branch.

yarfruct:

PR #261, opened 2020-02-24
Mer: Fix crash on KitManager initialization

stanislavmaksimov-omp:


src/plugins/mer/mersdkkitinformation.cpp
```diff
@@ -260,9 +260,6 @@ MerSdkKitInformationWidget::MerSdkKitInformationWidget(Kit *kit,
         const MerSdkKitInformation *kitInformation)
     : KitConfigWidget(kit, kitInformation)
 {
-    if (!visibleInKit())
-        return;
```
martyone: IIUC it would be more correct to replace this with `QTC_ASSERT(visibleInKit(), return)` and ensure the it does not fail.

martyone: It might be also good to instead use `DeviceTypeKitInformation::deviceTypeId(m_kit) == Constants::MER_DEVICE_TYPE` literally instead of calling `visibleInKit` to make it more clear what is asserted.

stanislavmaksimov-omp: Other parts of QtC expects fully constructed objects (editors/widgets for each configuration setting). I mean **this object has to be initialized in any case** and widgets will be hidden later according visibleInKit() return value. BTW any partial initialization of objects it is a bad idea (especially in complex solutions with a lots of public interfaces and private implementations)

martyone: I can agree on that, but we cannot fix it just by removing the barriers. We need to ensure that the rest of the code is not going to fail with incompatible kits, which is why the conditional return was implemented I believe.

stanislavmaksimov-omp: I had investigated code and there's nothing dangerous in private implementation of MerSdkKitInformationWidget. I am sure there is no reason to add such barrier and conditional return was added just in case.

martyone: Hm, it is somehow suspicious to me, because now I ran git-blame on it and noticed that the code was added by me and it was quite recently during big refactorization, so it's likely I added it to really deal with some issue. But checking the rest of the code now it all seems safe no matter which kit it is used with as you say. So OK, let's drop it. Thank you!

martyone:
@stanislavmaksimov-omp please rebase onto current master

stanislavmaksimov-omp:
@martyone, branch rebased

PR #264, opened 2020-02-28
Allow to set up JSON-based wizard information with variables

yarfruct:
This pull request adds possibility to set up the JSON-based wizard information like `displayName`, `displayCategory` and `description` with variables like `"%{VariableName}"`.

### DIFF
--- a/src/plugins/mer/merplugin.cpp
+++ b/src/plugins/mer/merplugin.cpp
@@ -154,6 +154,16 @@ bool MerPlugin::initialize(const QStringList &arguments, QString *errorString)
     RunControl::registerWorker<RemoteLinuxQmlProfilerSupport>(QML_PROFILER_RUN_MODE, constraint);
     //RunControl::registerWorker<RemoteLinuxPerfSupport>(PERFPROFILER_RUN_MODE, constraint);
 
+    Utils::globalMacroExpander()->registerVariable("SailfishOSVariant",
+                                                   tr("Expands to \"Sailfish OS\" or its variant name").arg(Sdk::osVariant()),
+                                                   []() { return Sdk::osVariant(); });
+    Utils::globalMacroExpander()->registerVariable("SailfishIDEVariant",
+                                                   tr("Expands to \"Sailfish IDE\" or its variant name").arg(Sdk::ideVariant()),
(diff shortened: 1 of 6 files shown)


vigejolla:
Please prefix the commit messages with the plugin/library you are modifying e.g. "Sfdk: Add missing includes"

src/plugins/mer/merplugin.cpp
```diff
@@ -154,6 +154,10 @@ bool MerPlugin::initialize(const QStringList &arguments, QString *errorString)
     RunControl::registerWorker<RemoteLinuxQmlProfilerSupport>(QML_PROFILER_RUN_MODE, constraint);
     //RunControl::registerWorker<RemoteLinuxPerfSupport>(PERFPROFILER_RUN_MODE, constraint);
 
+    Utils::globalMacroExpander()->registerVariable("SailfishOSVariant",
+                                                   tr("Category of a wizard for %1 files").arg(Sdk::osVariant()),
+                                                   []() { return Sdk::osVariant(); });
```
martyone: I would immediately register all the variables os/ide/sdk variant (similarly to how it is done under sfdk)

src/plugins/mer/merplugin.cpp
```diff
@@ -154,6 +154,16 @@ bool MerPlugin::initialize(const QStringList &arguments, QString *errorString)
     RunControl::registerWorker<RemoteLinuxQmlProfilerSupport>(QML_PROFILER_RUN_MODE, constraint);
     //RunControl::registerWorker<RemoteLinuxPerfSupport>(PERFPROFILER_RUN_MODE, constraint);
 
+    Utils::globalMacroExpander()->registerVariable("SailfishOSVariant",
+                                                   tr("Category of a wizard for %1 files").arg(Sdk::osVariant()),
```
martyone: Actually the possible use is not limited for wizard category. So e.g. `"Expands to \"Sailfish OS\" or its variant name"`, `"Expands to \"Sailfish IDE\" or its variant name"`... would be more appropriate.

The variables should be SailfishOsVariant, SailfishIdeVariant and SailfishSdkVariant (notice the case) unless I missed something.

yarfruct:

PR #265, opened 2020-02-28
ProjectExplorer: Add the export macros to the JsonSummaryPage declaration

yarfruct:
This pull request adds the `PROJECTEXPLORER_EXPORT` macros to the `JsonSummaryPage` class declaration.

We implement the JSON-based wizards for creating different files for Sailfish OS projects. For example, there is a wizard for creating QML-file with possibility to import external libraries like Qt Sensors and Qt Positioning. To add these libraries to the Sailfish OS project it have to add the dependencies to the project YAML-file by editing the file after going through the all wizard pages.

So, we have to implement our own "Summary" page to use it inside our `wizard.json` file. And for inherit the `JsonSummaryPage` class we add the `PROJECTEXPLORER_EXPORT` macros to the class declaration.

More conversation inside PR about implementation of the QML-file wizard: https://github.com/sailfishos/sailfish-qtcreator/pull/260#discussion_r382553827.

### DIFF
--- a/src/plugins/projectexplorer/jsonwizard/jsonsummarypage.h
+++ b/src/plugins/projectexplorer/jsonwizard/jsonsummarypage.h
@@ -36,7 +36,7 @@ class FolderNode;
 class Node;
 
 // Documentation inside.
-class JsonSummaryPage : public Internal::ProjectWizardPage
+class PROJECTEXPLORER_EXPORT JsonSummaryPage : public Internal::ProjectWizardPage
 {
     Q_OBJECT
 


martyone:
Yup, now submit that for upstream review (I would consider shortening the commit title a bit - up to 50 chars is the ideal length, not always possible though)

yarfruct:
@martyone 

> Yup, now submit that for upstream review (I would consider shortening the commit title a bit - up to 50 chars is the ideal length, not always possible though)

We made the commit message shorter. Now, how we should submit this PR for upstream review?

martyone:
UTFG :) https://www.google.com/search?q=qt%20creator%20contributing

yarfruct:
> UTFG :) https://www.google.com/search?q=qt%20creator%20contributing

Great, but maybe you already have some way to do this? Something like list of steps or tutorial? We don't want to make an error that can slow down the work.

Also, wouldn't it be a problem that we (yarfruct) make a contribution into Qt? There is so much context about Sailfish OS, Sailfish SDK, creating own JSON-based wizards. Will the whole process drag on for months?

martyone:
> > UTFG :) https://www.google.com/search?q=qt%20creator%20contributing
> 
> Great, but maybe you already have some way to do this? Something like list of steps or tutorial? We don't want to make an error that can slow down the work.

There are no other instructions known to me than what google finds.
 
> Also, wouldn't it be a problem that we (yarfruct) make a contribution into Qt? There is so much context about Sailfish OS, Sailfish SDK, creating own JSON-based wizards. Will the whole process drag on for months?

You are supposed to only submit changes that are not Sailfish OS specific. That's why you use separate commits for those. Just those commits that implement generic changes are to be submitted. Sailfish OS specific bit will remain on our side.

yarfruct:

PR #266, opened 2020-03-02
Allow to set the wizard list field items with a string

yarfruct:
This pull request adds possibility to set up the JSON-based wizard `ListField` items as a string.

The problem is tha by default `ComboBoxField` (child of `ListField`) does not allow to set list of items as a string, only as the JSON array. So, the field does not allow to pass the wizard variable to the `items` field. 

For example, following code is not working:

```json
{
    "name": "QmlBaseType",
    "trDisplayName": "Base type of QML document:",
    "type": "ComboBox",
    "data":
    {
           "items": "%{ItemListAsString}"
    }
}
```

Variable `ItemListAsString` contains list of items as a string in the JSON-array format. So, default `ComboBoxField` cannot be initialized in this case and whole wizard cannot be created also.

This is a reason why we customize `ListField` to use a string value as the `items` field.

### DIFF
--- a/src/plugins/projectexplorer/jsonwizard/jsonfieldpage.cpp
+++ b/src/plugins/projectexplorer/jsonwizard/jsonfieldpage.cpp
@@ -50,6 +50,8 @@
 #include <QStandardItem>
 #include <QItemSelectionModel>
 #include <QDir>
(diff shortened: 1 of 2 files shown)


yarfruct:

PR #267, opened 2020-03-04
Sfdk: Do not require that every VM has a snapshot

martyone:


src/libs/sfdk/vboxvirtualmachine.cpp
```diff
@@ -312,10 +312,14 @@ void VBoxVirtualMachinePrivate::fetchInfo(VirtualMachineInfo::ExtraInfos extraIn
         arguments.append(QLatin1String(LIST));
         arguments.append(QLatin1String(MACHINE_READABLE));
 
-        enqueue(arguments, batch, [=](VBoxManageRunner *runner) {
+        auto *runner = enqueue(arguments, batch, [=](VBoxManageRunner *runner) {
```
vigejolla: Here we have to variables called runner on a single line. Please consider renaming either of them.

martyone: they both refer the same instance at different point of time

vigejolla: Sure, but they are still different variables. Might become confusing, especially if the enqueue implementation is changed in the future so that it no longer returns the same pointer as it uses for calling onSuccess.

martyone: They are different variables at different scopes pointing to the same object.

vigejolla: You only know that they point to the same object because you know the implementation details of enqueue. And in any case, having the same name for variables in nested scopes is forbidden by pretty much any sane c++ coding convention.

martyone: 1. There is no runnerA and runnerB. There is just THE runner. In order to make life easier for the reader I am naming both variables "runner" to express exactly that fact.

2. If the callback was implemented as another member function, the argument would receive exactly the same name "runner" and you would never ask.

3. With some other API you could have

```
Runner runner = ...;
runner->enqueue(x, y, z, [](Runner *runner) {
  ....
});
```

would you expect something else than having the same name used? No.

4. Another example

```
Runner *runner = ...;
enqueue(runner, x, y, [runner]() {
 ....
});
```

would you expect something else than having the same name used? No.

Not all different scopes are different. Some scopes are different at implementation language level, some scopes are different at design level, some are different at both.

vigejolla: In your point 2., as the callback would be a different function, you wouldn't be hiding the outer scope variable, and I wouldn't have any issues with that.

In your point 3, you are not capturing the outer scope variable, so you are not hiding it there either, but I would still argue that the reader could confuse the variables.

I don't really see the connection between your example in point 4 and the case here, as there you only have the captured variable and no parameter variable at all, so there is no possibility for confusion.

Here you have now captured all variables, including runner, and then you are hiding the runner by having another variable with the same name in the inner scope.

If you wrote code like this:
Runner *runner = ... ;
{
    Runner *runner = ...;
}
and compiled it with -Wall, the compiler would give you a warning about hiding a variable. I suppose the reason why it doesn't warn in our case is that the = capture only captures variables which are actually used, and since you have hidden the variable, it's not used, and therefore it's not captured. But I wouldn't be surprised if some compiler in the future decided to warn also about hiding captured variables.

Your first point does have some merit - it is hard to make good names for both variables. I don't believe it's impossible though.

jlehtoranta: I see two solutions here:
a) Keep the current implementation, which more or less explains itself without a separate comment
or b) Use a different name for the variable and add a comment about its relation with "runner"

vigejolla: To be honest, I don't really like the design of enqueue here. It pretends to do the right thing, using std::unique_ptr for managing the pointer, and moving it around using std::move. But then it does the nasty thing of storing the pointer in another vareiable, which means that the pointer is no longer unique. I think the need to even have two variables rises from this.

So, instead of having "just" the problem of potential confusion to the reader, there might be something else as well: What quarantees do we have that the object pointed to by the runner variable really exists when it's used in line 322? And if the runner has actually been already executed, isn't it a bit late to set the expected exit codes?

vigejolla: Thinkin about this a bit more, I have a suggestion: modify enqueue (I'm talking about the lambda in line 269) so that it takes the expected exit codes as argument, and set them before calling commandQueue()->enqueue(). That makes sure the exit codes are set early enough, and also gets rid of the need for the second runner variable.

martyone: enqueue() is a factory method, i.e., kind of highlevel, named constructor. As such it 1) conveniently supports the common use case by accepting the mandatory/commonly used properties and nothing more (like normal constructors usually do; properties that need to be rarely changed from their default value can be set by different means after object construction), 2) takes care of ownership (passed that to CommandQueue) and, naturally, 3) enables access to the created object :)

unique_ptr exists as a means of managing exclusive ownership. It is important to not confuse ownership (lifecycle) management and access management. Those are different problems, though not definitely dissociated. Being exclusively owned does not imply being exclusively accessible - we know we have those compositions, aggregations, associations, dependencies
, ...

Qt is an event driven framework. No guessing is needed at which point an operation enqueued on the same thread may heave been already executed. This is rigorously known.

Nitpick: I believe -Wall wouldn't, even -Wpedantic wouldn't. One would need to use -Wshadow explicitly. But shadowing is really an unrelated topic.

vigejolla: It's perfectly fine for constructors to take parameters in order to support convenient constructions. I'd say partial construction should be considered as the exception instead of the rule.

What you write about ownership is exactly on point - You've given away the ownership, so you no longer know whether the object exists or not. 

You are making assumptions about execution order and implementation details in an asynchronous system. This can only lead to hard to find bugs later on. As soon as someone adds an event loop somewhere down the line of enqueue() implementations your assumptions go out of the window. The developers of mer plugin did the same assumption, and we saw crashes which were hard to figure out because of that.

I really don't understand why do you prefer code which has possibility of confusion to reader, and which is prone to bugs later on (read: code which is hard to maintain) instead of code which is easy to read and understand and which works regardless of the implementation details of underlying components.

Your nitpicking is correct: -Wshadow is needed to generate warning from this:
https://etherpad.jollamobile.com/pastebin/1583472384/

martyone: Sfdk is an asynchronous API that guarantees that 1) every asynchronous operation is truly asynchronous and 2) all asynchronous operations execute in order, without overlapping. So, no one will be ever allowed to request event processing somewhere down the line of enqueue() implementations. Then, since it is guaranteed that the callback 1) will be called, 2) will be called asynchronously and 3) with the runner alive as an argument, it is also guaranteed that the runner will exist and will not be dequeued before the caller scope exits. I don't need to make any assumptions, I work with well defined and stable semantics.

> I really don't understand why do you prefer code which has possibility of confusion to reader, and which is prone to bugs later on (read: code which is hard to maintain) instead of code which is easy to read and understand and which works regardless of the implementation details of underlying components.

Honestly, you claimed to be a beginner programmer https://github.com/sailfishos/sailfish-qtcreator/pull/221#discussion_r336852623. From that position, I would be more cautious about criticizing what I do not understand to (yet), and more open to ideas new to me. Cheer up! :)

vigejolla: How can you guarantee that no one will be ever allowed to modify the enqueue() implementation (or the implementation of something that enqueue() uses) to not do things that you wouldn't make it do? The person who will maintain and modify your code might be someone else than you.

About me claiming to be a beginner programmer - sorry, I lied, or at least exaggerated my beginnerness. I've been writing C++ for living for the last 22 years, about 10 years with Qt. So, even though I'm probably not on the same level as you are, I'm probably not exactly a beginner. I was just trying to make you understand that the person who has to maintain your code might not have as long history with C++ development as you have. I hope you accept my apology, I wish I hadn't misled you like that.

During my career I have seen enough bugs being caused by exactly this kind of assumptions that you are making here that I think I can say that I know what I'm talking about. I have also seen that clean, easy to read and maintain, code should always be preferred over code that's not.

martyone: I can guarantee that if someone breaks any of those two properties (strictly asynchronous and ordered) then it all collapses immediately and with fireworks! That's not an assumption that is a fact. All the code (outside or inside) relies on this strictly defined semantics, which allows composing asynchronous operations in a way that is both concise and easy to reason about (read: clean, easy to read and maintain). Changing anything on this is like swapping red and green on a semaphore..

vigejolla: It might collapse immediately or it might not, it really depends on luck. It might just work, because the order didn't happen to change after the first modification, but then someone tries it on another machine and it crashes. Exactly like we saw with the mer plugin, when the developer who wrote the code didn't expect anyone to change the implementation to have event loops, but then you moved the implementation to sfdk library and added event loops. And we didn't notice it until it crashed in the dialogs, seemingly randomly. 

And in any case, if you modified to code according to my suggestion, the reader of the code would not have to wonder when the object might be deleted because it's not used anymore after the call.

vigejolla: see commit ee33fe37890bd3cfb265bdbbcd395ef3c2ec9fb5 for how this could be done without holding on to the runner pointer after giving it away, and without making assumptions about the execution order, or future modifications to the code

PR #268, opened 2020-03-04
Jb45810

martyone:


martyone:
The first commit belongs to other PR

martyone:
Rebased, updated after VirtualMachine::Features addition + some improvements/fixes

src/libs/sfdk/virtualmachine.h
```diff
@@ -84,6 +84,7 @@ class SFDK_EXPORT VirtualMachine : public QObject
         ShrinkStorageSize = 0x08,
         OptionalHeadless = 0x10,
         Snapshots = 0x20,
+        SwapMemory = 0x40,
```
vigejolla, line 87: I don't quite follow here. We have VirtualMachine::Features::SwapMemory, and we have VirtualMachineInfo::swapSupported. They are almost the same but not quite - what's the difference?

martyone: One is part of public API, the other is internal.

vigejolla:
LGTM

PR #269, opened 2020-03-06
Mer: Favor the "Deploy as RPM Package" method

martyone:


martyone:
The order of registering factories defines the preference, which happens automatically during factory instantiation.

vigejolla:
> The order of registering factories defines the preference, which happens automatically during factory instantiation.

I understood that. I was just saying that it's not obvious and I don't like it because it could be done in a more obvious way.

PR #270, opened 2020-03-10
Jb47507

martyone:


martyone:
All commits prior to "Mer: Drop "Authorize" button from HW device widget" belong to other PRs

src/plugins/remotelinux/genericlinuxdeviceconfigurationwizardpages.cpp
```diff
@@ -232,12 +232,16 @@ bool GenericLinuxDeviceConfigurationWizardKeyDeploymentPage::isComplete() const
 
 bool GenericLinuxDeviceConfigurationWizardKeyDeploymentPage::validatePage()
 {
+    SshConnectionParameters sshParams = d->device->sshParameters();
```
vigejolla, line 235: Isn't this issue in upstream qtc as well? So shouldn't the fix be upstreamed?

martyone: Yes.

vigejolla: Then I would suggest you remove this commit from the PR so we can merge the other commits already, and cherry-pick the fix from upstream separately

martyone: We have more convenient workflow for our own contributions: merge to master, ch-p to mer-X.Y, try to get upstream, drop from mer-X.Y when accepted upstream.

vigejolla: How is that more convenient? If you need to modify the commit to get it accepted to upstream, you would have to modify commits already in master, which I doubt anyone would think is a good idea.

martyone: When upgrading to newer upstream, the commit is dropped from mer-X.Y, and the upstreamed version of the commit is merged from upstream. No modification to commits is done.

vigejolla: What happens to the commit which was merged to master?

martyone: Nothing happens to commits back in history :)

vigejolla: So, after we have upgraded master, we have two commits for the same change in the history. The commits may be identical or they might differ a lot, depending on how much the commit was changed before accepted to upstream. How is this not confusing?

martyone: What can I tell you. I do not find anything surprising about it. It is nothing special, and for bigger projects rather inevitable, that multiple branches exist (by any means), changes are replicated in branches (by any means), some apply cleanly, some not, because branches diverged meanwhile, then branches get merged at some point (by any means). Fortunately we have tools that support as very well in this area.

vigejolla: Sure, such things happen, and might be inevitable in some cases. But this is not one of those cases, as we have the option to just upstream the commit first and then cherry-pick it to master and avoid conflicts that way.

PR #272, opened 2020-03-12
Jb49300

vigejolla:
Disable editing VM Properties if underlying VM does not support changing them

### DIFF
--- a/src/libs/sfdk/dockervirtualmachine.cpp
+++ b/src/libs/sfdk/dockervirtualmachine.cpp
@@ -134,7 +134,8 @@ CommandQueue *commandQueue()
  */
 
 DockerVirtualMachine::DockerVirtualMachine(const QString &name, QObject *parent)
-    : VirtualMachine(std::make_unique<DockerVirtualMachinePrivate>(this), staticType(), name, parent)
+    : VirtualMachine(std::make_unique<DockerVirtualMachinePrivate>(this), staticType(),
+            staticFeatures(), name, parent)
 {
     Q_D(DockerVirtualMachine);
     d->setDisplayType(staticDisplayType());
@@ -157,6 +158,11 @@ QString DockerVirtualMachine::staticType()
 QString DockerVirtualMachine::staticDisplayType()
 {
     return tr("Docker");
+}
+
+VirtualMachine::Features DockerVirtualMachine::staticFeatures()
+{
+    return VirtualMachine::NoFeatures;
 }
 
 void DockerVirtualMachine::fetchRegisteredVirtualMachines(const QObject *context,
--- a/src/libs/sfdk/vboxvirtualmachine.cpp
+++ b/src/libs/sfdk/vboxvirtualmachine.cpp
@@ -202,7 +202,8 @@ CommandQueue *commandQueue()
  */
 
 VBoxVirtualMachine::VBoxVirtualMachine(const QString &name, QObject *parent)
-    : VirtualMachine(std::make_unique<VBoxVirtualMachinePrivate>(this), staticType(), name, parent)
+    : VirtualMachine(std::make_unique<VBoxVirtualMachinePrivate>(this), staticType(),
+            staticFeatures(), name, parent)
 {
     Q_D(VBoxVirtualMachine);
     d->setDisplayType(staticDisplayType());
@@ -225,6 +226,13 @@ QString VBoxVirtualMachine::staticType()
 QString VBoxVirtualMachine::staticDisplayType()
 {
     return tr("VirtualBox");
+}
+
+VirtualMachine::Features VBoxVirtualMachine::staticFeatures()
+{
+    return VirtualMachine::LimitMemorySize | VirtualMachine::LimitCpuCount
+        | VirtualMachine::GrowStorageSize | VirtualMachine::OptionalHeadless
+        | VirtualMachine::Snapshots;
 }
 
 void VBoxVirtualMachine::fetchRegisteredVirtualMachines(const QObject *context,
--- a/src/libs/sfdk/virtualmachine.cpp
+++ b/src/libs/sfdk/virtualmachine.cpp
@@ -38,6 +38,7 @@
 #include <QCoreApplication>
 #include <QDir>
 #include <QLockFile>
+#include <QTimer>
 #include <QThread>
 
 #ifdef Q_OS_WIN
@@ -113,12 +114,13 @@ private:
 VirtualMachine::ConnectionUiCreator VirtualMachine::s_connectionUiCreator;
 
 VirtualMachine::VirtualMachine(std::unique_ptr<VirtualMachinePrivate> &&dd, const QString &type,
-        const QString &name, QObject *parent)
+        const Features &features, const QString &name, QObject *parent)
     : QObject(parent)
     , d_ptr(std::move(dd))
 {
     Q_D(VirtualMachine);
     d->type = type;
+    d->features = features;
     d->name = name;
     QTC_ASSERT(s_connectionUiCreator, return);
     d->connectionUi_ = s_connectionUiCreator(this);
@@ -155,6 +157,11 @@ QString VirtualMachine::type() const
 QString VirtualMachine::displayType() const
 {
     return d_func()->displayType;
+}
+
+VirtualMachine::Features VirtualMachine::features() const
+{
+    return d_func()->features;
 }
 
 QString VirtualMachine::name() const
@@ -194,6 +201,7 @@ bool VirtualMachine::isHeadless() const
 void VirtualMachine::setHeadless(bool headless)
 {
     Q_D(VirtualMachine);
+    QTC_ASSERT(d->features & OptionalHeadless, return);
     if (d->headless == headless)
         return;
     d->headless = headless;
@@ -240,6 +248,8 @@ void VirtualMachine::setMemorySizeMb(int memorySizeMb, const QObject *context,
         const Functor<bool> &functor)
 {
     Q_D(VirtualMachine);
+    QTC_ASSERT(d->features & LimitMemorySize,
+               QTimer::singleShot(0, context, std::bind(functor, false)); return);
     QTC_CHECK(isLockedDown());
 
     const QPointer<const QObject> context_{context};
@@ -278,6 +288,8 @@ int VirtualMachine::cpuCount() const
 void VirtualMachine::setCpuCount(int cpuCount, const QObject *context, const Functor<bool> &functor)
 {
     Q_D(VirtualMachine);
+    QTC_ASSERT(d->features & LimitCpuCount,
+               QTimer::singleShot(0, context, std::bind(functor, false)); return);
     QTC_CHECK(isLockedDown());
 
     const QPointer<const QObject> context_{context};
@@ -308,6 +320,12 @@ void VirtualMachine::setStorageSizeMb(int storageSizeMb, const QObject *context,
         const Functor<bool> &functor)
 {
     Q_D(VirtualMachine);
+    QTC_ASSERT(d->features & (GrowStorageSize | ShrinkStorageSize),
+            QTimer::singleShot(0, context, std::bind(functor, false)); return);
+    QTC_ASSERT(!(storageSizeMb > d->virtualMachineInfo.storageSizeMb) || d->features & GrowStorageSize,
+            QTimer::singleShot(0, context, std::bind(functor, false)); return);
+    QTC_ASSERT(!(storageSizeMb < d->virtualMachineInfo.storageSizeMb) || d->features & ShrinkStorageSize,
+            QTimer::singleShot(0, context, std::bind(functor, false)); return);
     QTC_CHECK(isLockedDown());
 
     const QPointer<const QObject> context_{context};
--- a/src/libs/sfdk/virtualmachine.h
+++ b/src/libs/sfdk/virtualmachine.h
@@ -76,11 +76,23 @@ public:
         Synchronous
     };
 
+    enum Feature {
+        NoFeatures = 0x00,
+        LimitMemorySize = 0x01,
+        LimitCpuCount = 0x02,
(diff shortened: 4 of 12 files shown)


src/libs/sfdk/virtualmachine.h
```diff
@@ -104,15 +104,18 @@ class SFDK_EXPORT VirtualMachine : public QObject
...
     int cpuCount() const;
     void setCpuCount(int cpuCount, const QObject *context,
             const Functor<bool> &functor);
     static int availableCpuCount();
+    virtual bool cpuCountEditable() const { return true; }
 
     int storageSizeMb() const;
     void setStorageSizeMb(int storageSizeMb, const QObject *context,
             const Functor<bool> &functor);
+    virtual bool storageSizeEditable() const { return true; }
```
martyone: I would prefer adding flags instead

```
enum Feature {
    NoFeatures = 0x00,
    LimitMemorySize = 0x01,
    LimitCpuCount = 0x02,
    GrowStorageSize = 0x04,
    ShrinkStorageSize = 0x08,
    OptionalHeadless = 0x10,
    Snapshots = 0x11,
};
Q_DECLARE_FLAGS(Features, Feature)
...
QString displayType() const;
Features features() const;
QString name() const;
...
    VirtualMachine(std::unique_ptr<VirtualMachinePrivate> &&dd, const QString &type,
            const QString &name, const Features &features, QObject *parent);
```

and QTC_ASSERT their availability in this class already, e.g.:

```
void VirtualMachine::setMemorySizeMb(int memorySizeMb, const QObject *context,
        const Functor<bool> &functor)
{
    Q_D(VirtualMachine);
    QTC_ASSERT(d->features & LimitMemorySize,
            QTimer::singleShot(0, context, std::bind(functor, false)); return);
    ...
}
```

src/plugins/mer/mervirtualmachinesettingswidget.h
```diff
@@ -47,7 +52,7 @@ class MerVirtualMachineSettingsWidget : public QWidget
     void setMemorySizeMb(int sizeMb);
     void setCpuCount(int count);
     void setStorageSizeMb(int storageSizeMb);
-    void setVmOff(bool vmOff);
+    void setVmOff(Sfdk::VirtualMachine * const virtualMachine, bool vmOff);
```
martyone: This would not be changed and instead we would add `void setVmFeatures(...)` likely as the very first setter.

src/plugins/mer/mervirtualmachinesettingswidget.cpp
```diff
@@ -112,22 +120,20 @@ void MerVirtualMachineSettingsWidget::initGui()
     });
 
     ui->memoryInfoLabel->setPixmap(Utils::Icons::INFO.pixmap());
-    ui->memoryInfoLabel->setToolTip(
-            QLatin1String("<font color=\"red\">")
-            + tr("Stop the virtual machine to unlock this field for editing.")
-            + QLatin1String("</font>"));
+    setToolTip(ui->memoryInfoLabel, tr("Stop the virtual machine to unlock this field for editing."));
```
martyone: Nice, we could also use a variable and get rid of the repeated tr(message) finally.

src/plugins/mer/mervirtualmachinesettingswidget.h
vigejolla: The signature isn't changed, but the implementation still needs some changes to make sure that the controls aren't enabled

src/libs/sfdk/virtualmachine.cpp
```diff
@@ -308,6 +320,13 @@ void VirtualMachine::setStorageSizeMb(int storageSizeMb, const QObject *context,
         const Functor<bool> &functor)
 {
     Q_D(VirtualMachine);
+    if (storageSizeMb > d->virtualMachineInfo.storageSizeMb) {
+        QTC_ASSERT(d->features & GrowStorageSize,
+                   QTimer::singleShot(0, context, std::bind(functor, false)); return);
+    } else if (storageSizeMb < d->virtualMachineInfo.storageSizeMb) {
+        QTC_ASSERT(d->features & ShrinkStorageSize,
+                   QTimer::singleShot(0, context, std::bind(functor, false)); return);
+    }
```
martyone: I would like to handle also the case when no change is possible

```
QTC_ASSERT(d->features & (GrowStorageSize | ShrinkStorageSize),
        QTimer::singleShot(0, context, std::bind(functor, false)); return);
QTC_ASSERT(!(storageSizeMb > d->virtualMachineInfo.storageSizeMb) || d->features & GrowStorageSize,
        QTimer::singleShot(0, context, std::bind(functor, false)); return);
QTC_ASSERT(!(storageSizeMb < d->virtualMachineInfo.storageSizeMb) || d->features & ShrinkStorageSize,
        QTimer::singleShot(0, context, std::bind(functor, false)); return);
```

src/libs/sfdk/virtualmachine.h
```diff
@@ -104,15 +116,18 @@ class SFDK_EXPORT VirtualMachine : public QObject
     void setMemorySizeMb(int memorySizeMb, const QObject *context,
             const Functor<bool> &functor);
     static int availableMemorySizeMb();
+    virtual bool memorySizeEditable() const { return true; }
```
martyone: remnant

src/libs/sfdk/virtualmachine_p.h
```diff
@@ -192,6 +192,7 @@ class VirtualMachinePrivate
     QSsh::SshConnectionParameters sshParameters;
     bool headless = false;
     bool autoConnectEnabled = true;
+    VirtualMachine::Features features;
```
martyone: I would prefer moving this up before `name`. I would consider applying that order in the ctor too.

src/libs/sfdk/vboxvirtualmachine.cpp
```diff
@@ -202,7 +202,10 @@ CommandQueue *commandQueue()
  */
 
 VBoxVirtualMachine::VBoxVirtualMachine(const QString &name, QObject *parent)
-    : VirtualMachine(std::make_unique<VBoxVirtualMachinePrivate>(this), staticType(), name, parent)
+    : VirtualMachine(std::make_unique<VBoxVirtualMachinePrivate>(this), staticType(), name,
+            VirtualMachine::LimitMemorySize | VirtualMachine::LimitCpuCount
+            | VirtualMachine::GrowStorageSize | VirtualMachine::ShrinkStorageSize
+            | VirtualMachine::OptionalHeadless | VirtualMachine::Snapshots, parent)
```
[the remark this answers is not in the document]
vigejolla: Yup, I also thought it looked a bit ugly :)

src/libs/sfdk/virtualmachine.cpp
vigejolla: Sure, if you want.

src/libs/sfdk/virtualmachine.cpp
```diff
@@ -196,6 +203,7 @@ void VirtualMachine::setHeadless(bool headless)
     Q_D(VirtualMachine);
     if (d->headless == headless)
         return;
+    QTC_ASSERT(d->features & OptionalHeadless, return);
```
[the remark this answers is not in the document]
vigejolla: No it shouldn't. If it goes higher, we get always the assert failures because BuildEnginePrivate::fromMap() always calls setHeadless(). And there's nothing wrong with that, as it never changes the setting.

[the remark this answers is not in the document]
vigejolla: hmm. I suppose I should just change the fromMap() to check the feature flags...

src/libs/sfdk/virtualmachine.cpp
vigejolla: But may I ask, why do you prefer !(storageSizeMb > d->virtualMachineInfo.storageSizeMb) over storageSizeMb <= d->virtualMachineInfo.storageSizeMb?

src/libs/sfdk/virtualmachine.h
vigejolla: removed

src/libs/sfdk/virtualmachine_p.h
vigejolla: moved

src/libs/sfdk/virtualmachine.cpp
martyone: yup, because (A -> B) == (!A || B). So whenever you see (!A || B) you know you can read it simply as "A implies B".

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -134,7 +134,8 @@ CommandQueue *commandQueue()
  */
 
 DockerVirtualMachine::DockerVirtualMachine(const QString &name, QObject *parent)
-    : VirtualMachine(std::make_unique<DockerVirtualMachinePrivate>(this), staticType(), name, parent)
+    : VirtualMachine(std::make_unique<DockerVirtualMachinePrivate>(this), staticType(),
+            VirtualMachine::NoFeatures, name, parent)
```
martyone: Please apply the pattern with staticFeatures here too.

src/libs/sfdk/vboxvirtualmachine.cpp
```diff
@@ -201,8 +201,16 @@ CommandQueue *commandQueue()
  * \class VBoxVirtualMachine
  */
 
+static VirtualMachine::Features staticFeatures()
+{
+    return VirtualMachine::LimitMemorySize | VirtualMachine::LimitCpuCount
+        | VirtualMachine::GrowStorageSize | VirtualMachine::ShrinkStorageSize
+        | VirtualMachine::OptionalHeadless | VirtualMachine::Snapshots;
+}
+
```
martyone: Ah, I meant static member function, similar to `staticType` and `staticDisplayType`, ordered right after those two.

src/plugins/mer/mervirtualmachinesettingswidget.cpp
```diff
@@ -57,6 +57,24 @@ MerVirtualMachineSettingsWidget::~MerVirtualMachineSettingsWidget()
...
+    QString stopVmText = tr("Stop the virtual machine to unlock this field for editing.");
+    if (features & VirtualMachine::LimitMemorySize)
+        setToolTip(ui->memoryInfoLabel, stopVmText);
+    else
+        setToolTip(ui->memoryInfoLabel, tr("Virtual Machine does not allow changing memory size"));
+    if (features & VirtualMachine::LimitCpuCount)
+        setToolTip(ui->cpuInfoLabel, stopVmText);
+    else
+        setToolTip(ui->cpuInfoLabel, tr("Virtual Machine does not allow changing cpu count"));
+    if (features & (VirtualMachine::GrowStorageSize | VirtualMachine::ShrinkStorageSize))
```
martyone: We should also update the min/max limits based on that or at least add an assertion that the ShrinkStorageSize feature is not enabled.

vigejolla: We are getting quite far from the bug that I'm fixing here...

src/plugins/mer/mervirtualmachinesettingswidget.cpp
```diff
@@ -112,22 +130,27 @@ void MerVirtualMachineSettingsWidget::initGui()
...
+
+void MerVirtualMachineSettingsWidget::setStorageSizeLimits()
+{
+    // Prohibit adding more than MAX_STORAGE_SIZE_INCREMENT_GB in one step, because there is no way to go back
+    // with VirtualBox
+    const double currentValue = ui->storageSizeGbSpinBox->value();
+    const double maximumSizeGb = (m_features & VirtualMachine::GrowStorageSize)
+        ? currentValue + MAX_STORAGE_SIZE_INCREMENT_GB : currentValue;
+    const double minimumSizeGb = (m_features & VirtualMachine::ShrinkStorageSize)
+        ? 0 : currentValue;
```
martyone: great but the ugly extra parentheses :) I would prefer wrapping each operand of `?:` onto a separate line

```
    const double maximumSizeGb = m_features & VirtualMachine::GrowStorageSize
        ? currentValue + MAX_STORAGE_SIZE_INCREMENT_GB
        : currentValue;
    const double minimumSizeGb = m_features & VirtualMachine::ShrinkStorageSize
        ? 0
        : currentValue;
```

vigejolla: Just following Qt Creator coding convention with the parenthesis: https://doc-snapshots.qt.io/qtcreator-extending/coding-style.html#parentheses

martyone: That's a bit different case :) check `grep -e '^[[:space:]]*?' -C1 -r src/plugins/` for reference. Slightly related is also what I pasted before under some review https://pastebin.com/raw/PRWJsWip

vigejolla: Ok, let's then go against the coding convention. Sure, they've done it themselves in quite a few places so I guess it's understandable. Or maybe they have updated the convention but haven't updated the code everywhere yet?

vigejolla: Done. I still want to emphasize that:
1) I prefer using the parenthesis
2) So does the official Qt Creator coding style.
But as this bug fix needs to go in, I'm not going to waste time arguing about this any further.

martyone: My understanding is that the coding convention deals with two issues: code readability and error prevention. When comes to ternary operator use, the former is more conveniently solved by line wrapping and the latter is not likely to be relevant given that ternary operator has as low precedence as the assignment operators have. Of course the situation would be different at the level of its operands, so e.g.

```
return (A && B) || C
    ? (a + b) & c
    : 0;
```

would be prefered.

PR #274, opened 2020-03-13
Sfdk: Fix help on root access need for pkcon (un)deploy

martyone:


share/qtcreator/sfdk/modules/50-testing-mb2.json
```diff
@@ -19,14 +19,14 @@
...
...ns are allowed.  The '-*-debug@(info|source)' pattern is implied unless the '--debug' option is used.  (Note: '--debug' is not needed for successfull debugging with the 'debug' command as that command uses the local debug information.)  Options must precede patterns if any.\n\nThe following deployment methods are available, corresponding to the mandatory option:\n\n--pkcon\n    Uses 'pkcon' to install RPM package(s).  Depending on particular device operating system version and configuration it might not require root access to the device.\n\n--rsync\n    This method is special.  Uses 'rsync' to copy the installed files under '/opt/sdk/<name>' without building/installing RPM package(s).  Does not require root access to the device.  Incompatible with '--all'.\n\n--sdk\n    Uses the method native to the SDK to install RPM package(s). Requires developer mode enabled on the device, does not require root access.\n\n--zypper\n    Uses 'zypper' to install RPM package(s).  Requires 'zypper' installed on the device, requires root access to the device.\n\n--manual\n    Only transfers the RPM package(s) to the device for manual installation. Uses '~/RPMS' as the target directory on the device."
```
vigejolla: Isn't the "Allow untrusted software" required anymore?

martyone: More importantly it is ineffective w.r.t to plain pkcon usage in 3.2.1 already and iiuc it's not going to become effective in future, so I rather used excluded this information from the built-in help.

vigejolla: Is it really? Have you tested this, i.e. if you have root access, can you install untrusted packages without having enabled the "Allow untrusted software" setting?

martyone: > [...] if you have root access, can you install untrusted packages without having enabled the "Allow untrusted software" setting?

That's not what I am saying but yes, of course, if you have root access pkcon does what you want.

But only now I found that mb2 does not obey the user name that a device was configured with, so it is still inaccurate :\ I will return to this later

martyone:
Updated the message

PR #276, opened 2020-03-13
Jb49305

martyone:


martyone:

src/tools/sdktool/addsfdkbuildengineoperation.h
```diff
@@ -59,6 +59,7 @@ class AddSfdkBuildEngineOperation : public Operation
            const QUrl &vmUri,
            const QDateTime &creationTime,
            bool autodetected,
+           const QString &sharedInstallPath,
```
vigejolla: Nothing wrong with this, but now RmSfdkBuildEngineOperation::test() won't compile anymore, as it doesn't have this parameter

vigejolla:
Nice, just fix the tests :)

src/tools/sdktool/addsfdkbuildengineoperation.h
martyone: oops, fixed.

PR #277, opened 2020-03-16
Sfdk: Check virtualmachine features before modifying properties

vigejolla:


src/tools/sfdk/command.cpp
```diff
@@ -262,10 +281,16 @@ class VirtualMachinePropertiesAccessor : public PropertiesAccessor
         } else if (name == VM_STORAGE_SIZE_MB) {
             if (!parsePositiveInt(&m_storageSizeMb, value, errorString))
                 return Failed;
-            if (m_storageSizeMb < m_vm->storageSizeMb()) {
+            if ((m_storageSizeMb < m_vm->storageSizeMb())
+                && !(m_vm->features() & VirtualMachine::ShrinkStorageSize)) {
```
martyone: Should be indented once more. I would prefer omitting braces on the left side of `&&`. Similar below.

src/tools/sfdk/command.cpp
```diff
@@ -111,15 +111,26 @@ class PropertiesAccessor
...
+    }
+
     static QString unknownPropertyMessage(const QString &name)
     {
         return tr("Unrecognized property \"%1\"").arg(name);
     }
+
+    static QString modifyNotSupportedMessage(const QString &name)
+    {
+        return tr("Virtual machine does not support modifying property \"%1\"").arg(name);
```
martyone: I would prefer using more general (reusable) message, e.g.

```
static QString readOnlyPropertyMessage()
{
    return tr("Read-only property");
}
```

Could you please also fix the error reporting by adding property name prefix where it is written to stderr in `BuiltinWorker::setProperties`?

```
qerr() << property << ": " << errorString << endl;
```
...and removing the `name` argument from `unknownPropertyMessage`.

(at least I think that would work well after brief check)

src/tools/sfdk/command.cpp
vigejolla: fixed

src/tools/sfdk/command.cpp
vigejolla: fixed. Seems to work.

PR #284, opened 2020-03-30
Sfdk: Remove port forwarding rule synchronously during addition

yarfruct:
**Problem:** Adding the port forwarding rule is a task of the `VBoxVirtualMachinePrivate::doAddPortForwarding()` method. Actually, this method updates a port forwarding rule. It removes a port forwarding rule that already exists and adds a new one with the same name. For removing existing rules this method uses the `doRemovePortForwarding()` call that is asynchronous. Thus, we get a case when addition is done earlier than removing and in the end the `doAddPortForwarding()` removes the rule that was just added.

**Solution:** As a solution we propose to make the `doRemovePortForwarding()` call inside the `VBoxVirtualMachinePrivate::doAddPortForwarding()` method synchronous using `Sfdk::execAsynchronous()`. Thus, removing of the rule always will be performed before adding a new one.

### DIFF
--- a/src/libs/sfdk/vboxvirtualmachine.cpp
+++ b/src/libs/sfdk/vboxvirtualmachine.cpp
@@ -634,7 +634,10 @@ void VBoxVirtualMachinePrivate::doAddPortForwarding(const QString &ruleName,
     Q_ASSERT(context);
     Q_ASSERT(functor);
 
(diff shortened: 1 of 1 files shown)


martyone:
Thank you. The proposed solution doesn't address the root cause. See https://github.com/sailfishos/sailfish-qtcreator/pull/285 instead.

PR #288, opened 2020-04-07
Jb49238

martyone:


src/libs/sfdk/virtualmachine.h
```diff
@@ -106,7 +106,7 @@ class SFDK_EXPORT VirtualMachine : public QObject
     void setHeadless(bool headless);
 
     bool isAutoConnectEnabled() const; // FIXME internal?
-    void setAutoConnectEnabled(bool autoConnectEnabled);
+    bool setAutoConnectEnabled(bool autoConnectEnabled);
```
vigejolla, line 109: I know writing comments or documenting interfaces isn't exactly the QtC style, but I think a single line comment telling that the return value is the previous value would be nice. Having to always read the implementation before using a method adds unnecessary burden to the user of the interface.

martyone: I promise everything will get documented when I am at the api docs task :)

PR #291, opened 2020-04-09
Add SYS_ADMIN capability

vigejolla:
unshare needs SYS_ADMIN capability. On some systems also security option
apparmor=unconfined is needed.


martyone:
Something obsolete with wrong bug number I believe

PR #292, opened 2020-04-09
Jb49303

martyone:


martyone:
> LGTM. About fixing the proxy settings on Docker (I think we have already an open bug about that): Maybe this syncWwwProxy thing could be moved to virtualmachine implementations? For docker we could just add the *PROXY enviroment variables to the container.

Hopefully, but there may be also some difference in setting up proxy under emulator (not supported yet) which would require differentiating at the engine-emulator level.

PR #296, opened 2020-04-22
Jb48661

vigejolla:


vigejolla:
Only the last commit (Mer: Fix executable path on cmake projects) belongs to this PR

martyone:
To be fixed on project template side instead, like in https://github.com/sailfishos/cmakesample/pull/5

PR #306, opened 2020-05-07
MerSSH: Map also tool paths in CMake files

vigejolla:
This ensures that Qt from PATH is never displayed when configuring
projects.

This commit is not planned to be sent upstream.

### DIFF
--- a/src/tools/merssh/command.cpp
+++ b/src/tools/merssh/command.cpp
@@ -32,6 +32,7 @@
 #include <QDir>
 #include <QDirIterator>
 #include <QFile>
+#include <QRegularExpression>
 
 namespace {
 const char BACKUP_SUFFIX[] = ".qtccache";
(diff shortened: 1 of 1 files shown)


vigejolla:
When the cmake and gcc paths match, projectimporter no longer creates temporary kits

src/tools/merssh/command.cpp
```diff
@@ -253,6 +253,8 @@ void Command::maybeDoCMakePathMapping()
         QString data = QString::fromUtf8(reader.data());
         data.replace(Sfdk::Constants::BUILD_ENGINE_SHARED_HOME_MOUNT_POINT, sharedHomePath());
         data.replace(Sfdk::Constants::BUILD_ENGINE_SHARED_SRC_MOUNT_POINT, sharedSourcePath());
+        data.replace("/usr/bin/gcc", sdkToolsPath() + "/gcc");
+        data.replace("/usr/bin/cmake", sdkToolsPath() + "/cmake");
```
martyone, line 257: Beautiful!

I think it's worth being more strict here, using "=/usr/bin/cmake$" instead and adding some QTC_CHECK here that it matched.

I checked CMakeCache.txt in one of test projects I built before and it containe /usr/bin/c++ but no /usr/bin/gcc.

vigejolla: Doing what you suggest would require using regexp replace instead of simple string compare. But while making that change I could actually change it so that it replaces all CMAKE_CXX_COMPILER and CMAKE_C_COMPILER values, regardless of the original value. That should work even if the original cache file has /usr/bin/c++ or /usr/bin/cc.

martyone: good idea, same could be done with cmake.

PR #310, opened 2020-05-11
Jb49846

martyone:


martyone:

src/plugins/mer/mersdkmanager.cpp
```diff
@@ -580,17 +580,24 @@ void MerSdkManager::ensureCmakeToolIsSet(Kit *k, const BuildEngine *buildEngine,
...
+            cmakeTool->setAutorun(true);
+            cmakeTool->setAutoCreateBuildDirectory(true);
+            const Core::Id id = cmakeTool->id();
+            const bool registeredOk = CMakeToolManager::registerCMakeTool(std::move(cmakeTool));
+            QTC_ASSERT(registeredOk, return);
+            CMakeKitInformation::setCMakeTool(k, id);
+            QStringList cmakeConf = { "CMAKE_CXX_COMPILER:STRING=/usr/bin/gcc",
+                                      "CMAKE_C_COMPILER:STRING=/usr/bin/gcc",
+                                      "CMAKE_PREFIX_PATH:STRING=%{Qt:QT_INSTALL_PREFIX}",
+                                      "QT_QMAKE_EXECUTABLE:STRING=%{Qt:qmakeExecutable}" };
```
vigejolla, line 598: I have a feeling that these should point to the wrapper scripts

src/plugins/cmakeprojectmanager/cmaketool.cpp
```diff
@@ -48,6 +48,7 @@ const char CMAKE_INFORMATION_DISPLAYNAME[] = "DisplayName";
 const char CMAKE_INFORMATION_AUTORUN[] = "AutoRun";
 const char CMAKE_INFORMATION_AUTO_CREATE_BUILD_DIRECTORY[] = "AutoCreateBuildDirectory";
 const char CMAKE_INFORMATION_AUTODETECTED[] = "AutoDetected";
+const char CMAKE_INFORMATION_AUTODETECTED_BY_PLUGIN[] = "AutoDetectedByPlugin";
```
vigejolla, line 51: Why is it necessary to distinguish CMake tools which are autodetected by other plugins? Couldn't we just mark them as AutoDetected?

src/plugins/mer/mersdkmanager.cpp
martyone: To me it looks suspicious too but I couldn't fully understand the effect quickly. Definitely something to keep in mind and possibly verify, but out of scope here.

src/plugins/cmakeprojectmanager/cmaketool.cpp
martyone: Do you really think I would do it just for fun? :) :) Commit message explains:

> Other plugins may autodetect CMake tools on custom paths. Those would be
> dropped by CMakeToolSettingsAccessor on startup.

vigejolla: No I don't :) 
I misread the CMakeToolSettingsAccessor change and was a bit confused by it. Now reading it again I understand the purpose.

martyone: Yeah it's a bit ugly hack there :) Pity they did not implement it similarly how it is done under Debugger plugin which allows for plugin provided autodetection :\

PR #311, opened 2020-05-12
MerSsh: Map paths from target root with shared target paths

vigejolla:


src/tools/merssh/command.cpp
```diff
@@ -223,11 +226,25 @@ bool Command::isValid() const
...
 
+
+QString Command::rootFromDir(const QString &dir) const
+{
+    QString result = dir;
+    int treeDepth = dir.count('/');
+    for (int i = 0; i < treeDepth; i++) {
+        result += "/..";
+    }
+    return result;
```
martyone, line 237:
```suggestion
    return dir + '/' + QDir(dir).relativeFilePath("/");
```

src/tools/merssh/command.h
```diff
@@ -64,6 +64,7 @@ class Command : public QObject
     virtual int execute() = 0;
 
 private:
+    QString rootFromDir(const QString& dir) const;
```
martyone, line 67: spacing around &, I would prefer static.

src/tools/merssh/command.cpp
```diff
@@ -252,6 +269,14 @@ void Command::maybeDoCMakePathMapping()
...
 
         QString data = QString::fromUtf8(reader.data());
+        if (it.fileName() == "CMakeCache.txt") {
+            QRegularExpression cmakeHomeRegexp("CMAKE_HOME_DIRECTORY:INTERNAL=(.*)");
+            QRegularExpressionMatch cmakeHomeMatch=cmakeHomeRegexp.match(data);
+            if (cmakeHomeMatch.hasMatch())
+                relativeRoot = rootFromDir(cmakeHomeMatch.captured(1));
+        }
+        if (!relativeRoot.isEmpty())
+            data.replace(relativeRoot, sharedTargetRoot);
```
martyone, line 279: .cbp files on my machine contain paths with backslashes too `<Option virtualFolder="CMake Files\..\..\..\..\..\usr\lib\cmake\Qt5Core\"/>` - shouldn't these be replaced?

src/tools/merssh/command.cpp
```diff
@@ -98,12 +98,12 @@ void Command::setSharedSourcePath(const QString& path)
...
 QString Command::sharedTargetPath() const
 {
-    return m_sharedSourcePath;
+    return m_sharedTargetPath;
 }
 
 void Command::setSharedTargetPath(const QString& path)
 {
-    m_sharedSourcePath = QDir::fromNativeSeparators(path);
+    m_sharedTargetPath = QDir::fromNativeSeparators(path);
```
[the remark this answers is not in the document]
vigejolla: moved to separate commit

src/tools/merssh/command.cpp
```diff
@@ -252,6 +269,14 @@ void Command::maybeDoCMakePathMapping()
         }
 
         QString data = QString::fromUtf8(reader.data());
+        if (it.fileName() == "CMakeCache.txt") {
+            QRegularExpression cmakeHomeRegexp("CMAKE_HOME_DIRECTORY:INTERNAL=(.*)");
+            QRegularExpressionMatch cmakeHomeMatch=cmakeHomeRegexp.match(data);
```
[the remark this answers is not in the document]
vigejolla: fixed

src/tools/merssh/command.cpp
```diff
@@ -223,11 +226,25 @@ bool Command::isValid() const
...
+
 void Command::maybeDoCMakePathMapping()
 {
     if (!QFile::exists("CMakeCache.txt"))
         return;
 
+    QString sharedTargetRoot = sharedTargetPath() + "/" + targetName();
+    QString relativeRoot;
+
     QDirIterator it(".", {"CMakeCache.txt", "*.cbp", "QtCreatorDeployment.txt"}, QDir::Files,
```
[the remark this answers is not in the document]
vigejolla: Your assessment is correct. I moved reading the cmake home variable to a separate step before the loop now.

src/tools/merssh/command.cpp
vigejolla: Beautiful, thanks! :)

src/tools/merssh/command.cpp
vigejolla: To me it seems QtC does not use those

src/tools/merssh/command.cpp
```diff
@@ -223,11 +226,31 @@ bool Command::isValid() const
     return true;
 }
 
+static QString readRelativeRoot()
```
martyone, line 229: I meant static member, sorry for that.

It would be nice to flatten the body using this style:

```
if (!cond)
    return;

....
```

src/tools/merssh/command.cpp
```diff
@@ -223,11 +226,31 @@ bool Command::isValid() const
...
+}
+
 void Command::maybeDoCMakePathMapping()
 {
     if (!QFile::exists("CMakeCache.txt"))
         return;
 
+    QString sharedTargetRoot = sharedTargetPath() + "/" + targetName();
+    QString relativeRoot = readRelativeRoot();
+    Q_ASSERT(!relativeRoot.isEmpty());
```
martyone, line 252: Inactive in release build, use QTC_CHECK please.

src/tools/merssh/command.cpp
martyone: I would consider replacing these too so that we do not need to ponder about in future.

vigejolla: What would you replace them with? "/usr/lib/cmake/Qt5Core/" or "\usr\lib\cmake\Qt5Core\" or "CMake Files\usr\lib\cmake\Qt5Core" or "CMake Files/usr/lib/cmake/Qt5Core/" or somethine else?

martyone: Hm, I think `sharedTargetRoot` should look more like the original `relativeRoot`, i.e. it should be a concatenation of source path and target root path relative to source path.

Better expressed by example:

```diff
-<Unit filename="/home/mersdk/share/tmp/foobar/../../../../../usr/lib/cmake/Qt5Core/Qt5CoreMacros.cmake">
+<Unit filename="/home/user/tmp/foobar/../../SailfishOS/mersdk/targets/SailfishOS-3.3.0.16-i486/usr/lib/cmake/Qt5Core/Qt5CoreMacros.cmake">
```

(Given that `/home/user/` is user home on host and the SDK is installed under `/home/user/SailfishOS`.)

This is then more strightforward to be extend to the `Option` tag:

```diff
-    <Option virtualFolder="CMake Files\..\..\..\..\..\usr\lib\cmake\Qt5Core\"/>
+    <Option virtualFolder="CMake Files\..\..\SailfishOS\mersdk\targets\SailfishOS-3.3.0.16-i486\usr\lib\cmake\Qt5Core\"/>
```

vigejolla: And what if SDK is installed in C:\SailfishOS, user home directory is D:\Users\myuser and he is using alternative source directory in E:\src?

vigejolla: The more I think about this, the more it seems to me that it doesn't make any sense to modify the virtualFolder at all. There are always cases when the value wouldn't make any sense, and as long as it is not used anywhere (QtC only checks that the value exists), it does not matter what it is. Modifying it is just wasted time.

src/tools/merssh/command.cpp
vigejolla: fixed

src/tools/merssh/command.cpp
vigejolla: fixed

src/tools/merssh/command.cpp
martyone: yeah, lovely :\

vigejolla:
> Consider replacing `return QString()` with `return {}` yet :)

Replaced.

PR #312, opened 2020-05-12
Jb49844

martyone:


src/tools/merssh/cmakecommand.cpp
```diff
@@ -68,10 +71,30 @@ int CMakeCommand::execute()
         return 1;
     }
 
+    // FIXME If merssh is passed "'Foo Bar'", Command::arguments() returns ["'Foo", "Bar'"].
+    const bool abortOnMeta = false;
```
vigejolla, line 75: abortOnMeta is used only once, and it's passed by value. I would just omit the whole variable.

martyone: Compare `foo(false)` with `foo(abortOnMeta=false)` from reader's point of view :)

PR #313, opened 2020-05-13
Jb49791

martyone:


vigejolla:
> I haven't seen those FILEPATHs, but as long as we don't change the existing STRINGs your changes LGTM

Actually, now I see them as well. Well, the main point is that now it works :)

martyone:
Extended CMAKE_SYSROOT mapping to both PATH and STRING types, added QT_QMAKE_EXECUTABLE handling.

src/tools/merssh/command.cpp
```diff
@@ -276,18 +276,35 @@ void Command::maybeDoCMakePathMapping()
         }
 
         QString data = QString::fromUtf8(reader.data());
+
```
vigejolla, line 279: I believe whitespace changes on unrelated lines in a same commit with functional changes are against Qt commit policy. Just saying, even though I'm not sure the Qt commit policy is really relevant here.

martyone: Roughly speaking, I believe it makes crucial difference whether it is a change to existing non-blank lines or not. The reason for splitting changes into smaller portions is to allow determining the origin/purpose of code, and vertical space is rarely interesting in this respect. So I am OK with changes like this :)

PR #318, opened 2020-05-28
Jb49043

martyone:


src/tools/sfdk/remoteprocess.h
```diff
@@ -124,7 +124,7 @@ private slots:
     QString m_workingDirectory;
     QSsh::SshConnectionParameters m_sshConnectionParams;
     QProcessEnvironment m_extraEnvironment;
-    bool m_runInTerminal = true;
+    bool m_runInTerminal = false;
```
vigejolla, line 127: m_runInTerminal is now initialized in constructor, so initialization here has no effect.

martyone: OTOH if I saw single member uninitialized, it would make me ponder _why_ unnecessarily, so I prefer initializing all primitive members using some safe defaults here :)

PR #320, opened 2020-05-28
Jb50035

martyone:


martyone:

src/plugins/compilationdatabaseprojectmanager/compilationdatabaseutils.cpp
```diff
@@ -148,9 +148,7 @@ void filteredFlags(const QString &fileName,
         }
 
         if (flag == "-o" || flag == "-MF" || flag == "-c" || flag == "-pedantic"
-                || flag.startsWith("-O") || flag.startsWith("-W") || flag.startsWith("-w")
-                || QString::compare(flag, "-fpic", Qt::CaseInsensitive) == 0
-                || QString::compare(flag, "-fpie", Qt::CaseInsensitive) == 0) {
+                || flag.startsWith("-O") || flag.startsWith("-W") || flag.startsWith("-w")) {
```
vigejolla, line 151: This is a bit strange, why were they filtered out in the first place? Have you tried upstreaming this change already?

martyone: Unfortunately that's unclear. It was added with one big commit enabling use with clang code model IIUC. Maybe something changed later on the clang code model that it makes not harm now. Not tried upstreaming yet.

martyone:
Upstream bug filed https://bugreports.qt.io/browse/QTCREATORBUG-24106

PR #323, opened 2020-06-03
Jb45818

martyone:


qtcreator.pri
```diff
@@ -169,10 +169,12 @@ gcc:!clang: QMAKE_CXXFLAGS += -Wno-noexcept-type
 
 RELATIVE_PLUGIN_PATH = $$relative_path($$IDE_PLUGIN_PATH, $$IDE_BIN_PATH)
 RELATIVE_LIBEXEC_PATH = $$relative_path($$IDE_LIBEXEC_PATH, $$IDE_BIN_PATH)
+RELATIVE_REVERSE_LIBEXEC_PATH = $$relative_path($$IDE_BIN_PATH, $$IDE_LIBEXEC_PATH)
 RELATIVE_DATA_PATH = $$relative_path($$IDE_DATA_PATH, $$IDE_BIN_PATH)
 RELATIVE_DOC_PATH = $$relative_path($$IDE_DOC_PATH, $$IDE_BIN_PATH)
 DEFINES += $$shell_quote(RELATIVE_PLUGIN_PATH=\"$$RELATIVE_PLUGIN_PATH\")
 DEFINES += $$shell_quote(RELATIVE_LIBEXEC_PATH=\"$$RELATIVE_LIBEXEC_PATH\")
+DEFINES += $$shell_quote(RELATIVE_REVERSE_LIBEXEC_PATH=\"$$RELATIVE_REVERSE_LIBEXEC_PATH\")
```
vigejolla, line 177: These are only used in merssh. Please consider having the definitions in merssh.pro only.

src/tools/merssh/command.cpp
```diff
@@ -67,6 +69,47 @@ int Command::executeRemoteCommand(const QString &command)
...
+    maybeUndoCMakePathMapping();
+
+    int rc{};
+    process.start();
+    if (!process.waitForFinished(-1) || process.error() == QProcess::FailedToStart)
+        rc = -2; // like QProcess::execute does
+    else
+        rc = process.exitStatus() == QProcess::NormalExit ? process.exitCode() : -1;
+
+    maybeDoCMakePathMapping();
```
vigejolla, line 109: To me it looks like the only additional value merssh provides here is the path mapping. I think it would be good to consider adding that functionality to sfdk and calling sfdk directly. I don't mean that we should do that now, but maybe at some convenient point in the future.

martyone: Yup, path mapping is definitely going to be moved to sfdk (at some point) but the rest depends on the exact meaning of "here" and "invoking directly". merssh also provides the cached queries and this is something I am not sure we should taint sfdk side with.

qtcreator.pri
martyone: right, that would be better.

src/tools/merssh/command.cpp
vigejolla: Please interpret my writings as something that makes sense ;)

I was only thinking about these cases where we use merssh as a thin wrapper for sfdk. So some logic would be needed before calling to determine which binary to call. And by "invoking directly", I was thinking about modifying the wrapper scripts to call sfdk instead of merssh, but of course other possibilities do exist. 

It's really not that straight forward, that's why I wrote that we should not do it now.

martyone: :)

PR #324, opened 2020-06-05
TeaLeafReader: Don't forcefully stop CMake when restarting

vigejolla:
Trying forcefully stop cmake is not a good idea with merssh, for two reasons:
1) It doesn't work, cmake keeps on running on the build engine
2) It kills merssh, which results in cmake files not having their paths mapped.
Therefore, it's better to just let cmake run to finish, ignore the results, and start again.

This might not be perfect solution, but it's going to be a temporary one anyway: upstream is getting rid of the whole TeaLeafReader in 4.13. This probably means that we are going to have to figure out how to run cmake in server mode inside build targets, translating paths on the fly...

### DIFF
--- a/src/plugins/cmakeprojectmanager/tealeafreader.cpp
+++ b/src/plugins/cmakeprojectmanager/tealeafreader.cpp
@@ -213,6 +213,10 @@ void TeaLeafReader::parse(bool forceConfiguration)
 
 void TeaLeafReader::stop()
 {
+    if (isParsing()) {
+        m_stopping = true;
+        return;
+    }
     cleanUpProcess();
 
     if (m_future) {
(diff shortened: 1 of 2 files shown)


vigejolla:
Also, because TeaLeadReader is going away, it doesn't make sense to try to upstream this

martyone:
After upgrade to 4.12 merssh will be using sfdk which does signal handling and kills the remote cmake process - maybe wait for that?

vigejolla:
> After upgrade to 4.12 merssh will be using sfdk which does signal handling and kills the remote cmake process - maybe wait for that?

Upgrade to 4.12 is not going to be in our 3.2 release, and I'd rather have this fix in 3.2. I suggest I now create a branch for the 3.2 release, and merge this to that branch instead of master. How does that sound?

martyone:
Release branches in this repo are merged back to master, so OK to simply merge that to master if you want it in 3.2

vigejolla:
> Release branches in this repo are merged back to master, so OK to simply merge that to master if you want it in 3.2

Ok. We just need to remember to revert the commit after the change from merssh to sfdk has been done then.

PR #325, opened 2020-06-09
Jb50188

etaishev:
Added the "rpm sign" build stage for signing rpm after build

The last commit with exports needed to redefine the RPM sign build step for customization.

### DIFF
--- a/share/qtcreator/sfdk/modules/20-building-mb2.json
+++ b/share/qtcreator/sfdk/modules/20-building-mb2.json
@@ -36,10 +36,11 @@
         },
         {
             "name": "build",
-            "trSynopsis": "[-p|--prepare] [--no-check] [-d|--enable-debug] [-j <n>] [<project-dir-or-file>] [[--] <rpmbuild-extra-arg>...]",
+            "trSynopsis": "[-p|--prepare] [--no-check] [-d|--enable-debug] [-j <n>] [--sign] [<project-dir-or-file>] [[--] <rpmbuild-extra-arg>...]",
             "trBrief": "Execute all build steps at once",
             "configOptions": ["*", "target!"],
-            "trDescription": "Run rpmbuild. Execute all sections of the .spec file except for %prep unless told otherwise.\n\nIf path to a <project-dir-or-file> is passed, shadow build will be done. Otherwise the current working directory is treated as the project directory. (The file name part is ignored - project file selection is done by qmake call inside the .spec file.)\n\n--prepare         Run %prep section as well\n--no-check        Skip the %check section\n--enable-debug    Enable debug build\n-j <n>            Run up to <n> jobs simultaneously"
+            "trDescription": "Run rpmbuild. Execute all sections of the .spec file except for %prep unless told otherwise.\n\nIf path to a <project-dir-or-file> is passed, shadow build will be done. Otherwise the current working directory is treated as the project directory. (The file name part is ignored - project file selection is done by qmake call inside the .spec file.)\n\n--prepare         Run %prep section as well\n--no-check        Skip the %check section\n--enable-debug    Enable debug build\n-j <n>            Run up to <n> jobs simultaneously\n--sign            Sign packages",
+            "preRun": "module.maybeImportSigningKey"
         },
         {
             "name": "qmake",
@@ -70,10 +71,11 @@
         },
         {
             "name": "package",
-            "trSynopsis": "[--no-check]",
+            "trSynopsis": "[--no-check] [--sign]",
             "trBrief": "Execute package creation build step",
             "configOptions": ["*", "target!"],
-            "trDescription": "Run rpmbuild. Execute just the %install section of the .spec file, then build binary packages. Unless '--no-check' is passed, execute also the %check section if present."
+            "trDescription": "Run rpmbuild. Execute just the %install section of the .spec file, then build binary packages and optionally sign those. Unless '--no-check' is passed, execute also the %check section if present.",
+            "preRun": "module.maybeImportSigningKey"
         }
     ],
 
@@ -144,6 +146,21 @@
         {
             "name": "no-pull-build-requires",
             "trDescription": "Do not update build-time dependencies unless required explicitly with 'build-requires pull'."
+        },
+        {
+            "name": "package.signing-user",
+            "trArgument": "<name>",
+            "trDescription": "Select the GPG key to sign packages with. Accepts the same values\n as the '--local-user' option of the gpg command does. See GPG\n documentation for more details.\n\nDepending on your SDK variant, different semantics may apply."
+        },
+        {
+            "name": "package.signing-passphrase",
+            "trArgument": "<passphrase>",
+            "trDescription": "In case a passphrase-protected signing key is selected for signing\n packages, use <string> as the passphrase."
+        },
+        {
+            "name": "package.signing-passphrase-file",
+            "trArgument": "<file>",
+            "trDescription": "In case a passphrase-protected signing key is selected for signing\n packages, read the passphrase from the <file>. Only the first line will\n be read."
         }
     ]
 }
--- a/share/qtcreator/sfdk/modules/20-building-mb2.mjs
+++ b/share/qtcreator/sfdk/modules/20-building-mb2.mjs
@@ -35,3 +35,18 @@ export function validateSearchOutputDirOption(value) {
 
     return [false, qsTr("Invalid keyword used")];
 }
+
+
+export function maybeImportSigningKey() {
+    if (!configuration.isOptionSet("package.signing-user"))
+        return [true, ""];
+
+    var signingUser = configuration.optionArgument("package.signing-user");
+    var signingPassphraseFile = configuration.isOptionSet("package.signing-passphrase-file")
+        ? configuration.optionArgument("package.signing-passphrase-file")
+        : "";
+    if (!buildEngine.importGpgKey(signingUser, signingPassphraseFile))
+        return [false, qsTr("Failed to share GnuPG key with the build engine")];
+
+    return [true, ""];
+}
--- a/src/libs/sfdk/signingutils.cpp
+++ b/src/libs/sfdk/signingutils.cpp
@@ -1,0 +1,319 @@
+/****************************************************************************
+**
+** Copyright (C) 2021 Open Mobile Platform LLC.
+** Contact: http://jolla.com/
+**
+** This file is part of Qt Creator.
+**
+** GNU Lesser General Public License Usage
+** Alternatively, this file may be used under the terms of the GNU Lesser
+** General Public License version 2.1 as published by the Free Software
+** Foundation and appearing in the file LICENSE.LGPL included in the
+** packaging of this file.  Please review the following information to
+** ensure the GNU Lesser General Public License version 2.1 requirements
+** will be met: http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html.
+**
+** Other Usage
+**
+** Alternatively, this file may be used in accordance with the terms and
+** conditions contained in a signed written agreement between you and Digia.
+**
+****************************************************************************/
+
+#include "signingutils_p.h"
+
+#include "asynchronous_p.h"
+#include "sdk_p.h"
+#include "sfdkconstants.h"
+#include "sfdkglobal.h"
+
+#include <app/app_version.h>
+#include <utils/algorithm.h>
+#include <utils/environment.h>
+#include <utils/fileutils.h>
+#include <utils/qtcassert.h>
+
+#include <QDir>
+#include <QRegularExpression>
+#include <QSettings>
+#include <QTimer>
+
+using namespace Utils;
+
+namespace Sfdk {
+
+namespace {
+
+const char SIGNING_KEY_EXTENSION[] = ".key";
+
+} // namespace anonymous
+
+namespace {
+
+QString gpgPath()
+{
+    // See ProjectExplorerPlugin::extensionsInitialized()
+    QSettings qtcSettings{QLatin1String(Core::Constants::IDE_SETTINGSVARIANT_STR),
+                          QLatin1String(Core::Constants::IDE_CASED_ID)};
+    const QString gitBinary = qtcSettings.value("Git/BinaryPath", "git").toString();
+    const QStringList rawGitSearchPaths
+        = qtcSettings.value("Git/Path").toString().split(':', QString::SkipEmptyParts);
+
+    const auto searchPathRetriever = [=] {
+        Utils::FilePaths searchPaths;
+        const QString libexecPath = QDir::cleanPath(QCoreApplication::applicationDirPath() + '/'
+                + RELATIVE_LIBEXEC_PATH);
+        searchPaths << Utils::FilePath::fromString(libexecPath);
+        if (Utils::HostOsInfo::isWindowsHost()) {
+            const Utils::FilePaths gitSearchPaths
+                = Utils::transform(rawGitSearchPaths, [](const QString &rawPath) {
+                      return Utils::FilePath::fromString(rawPath);
+                  });
+            const Utils::FilePath fullGitPath = Utils::Environment::systemEnvironment()
+                                                    .searchInPath(gitBinary, gitSearchPaths);
+            if (!fullGitPath.isEmpty()) {
+                searchPaths << fullGitPath.parentDir()
+                            << fullGitPath.parentDir().parentDir() + "/usr/bin";
+            }
+        }
+        return searchPaths;
+    };
+
+    //By default, gpg on windows is put together with git, so we look for it there
+    const QList<FilePath> additionalSearchPaths = searchPathRetriever();
+
+    const FilePath filePath = Environment::systemEnvironment()
+            .searchInPath("gpg", additionalSearchPaths);
+    if (!filePath.isEmpty())
+        return filePath.toString();
+    return {};
+}
+
+class GpgRunner : public ProcessRunner
+{
+    Q_OBJECT
+public:
+    explicit GpgRunner(const QStringList &arguments, QObject *parent = 0)
+        : ProcessRunner(path(), arguments, parent)
+    {}
+
+private:
+    static QString path()
+    {
+        QString path = gpgPath();
+
+        if (!SdkPrivate::customGpgPath().isEmpty())
+            path = SdkPrivate::customGpgPath();
+
+        return path;
+    }
+};
+
+CommandQueue *commandQueue()
+{
+    return Sfdk::SdkPrivate::commandQueue();
+}
+
+} // namespace anonymous
+
+/*!
+ * \class SigningUtils
+ * \internal
+ */
+
+SigningUtils::SigningUtils(QObject *parent)
+    : QObject(parent)
+{}
+
+SigningUtils::~SigningUtils() = default;
+
+void SigningUtils::listSecretKeys(const QObject *context,
+        const Functor<bool, const QList<GpgKeyInfo> &, QString> &functor)
+{
+    QStringList arguments;
+    arguments.append("--with-colons");
+    arguments.append("--list-secret-keys");
+
+    auto runner = std::make_unique<GpgRunner>(arguments);
+
+    QObject::connect(runner.get(), &GpgRunner::done, context, [=, process = runner->process()](bool ok) {
+        if (!ok) {
+            functor(false, {}, QString::fromUtf8(process->readAllStandardError()));
+            return;
+        }
+        const QList<GpgKeyInfo> keysInfo = gpgKeyInfoListFromOutput(
+            QString::fromUtf8(process->readAllStandardOutput().trimmed()));
+        functor(true, keysInfo, {});
+     });
+    commandQueue()->enqueue(std::move(runner));
+}
+
+void SigningUtils::exportSecretKey(const QString &id, const Utils::FilePath &passphraseFile,
+        const Utils::FilePath &outputDir, const QObject *context, const Functor<bool, QString> &functor)
+{
+    const QPointer<const QObject> context_{context};
+
+    QTC_ASSERT(!id.isEmpty(),
+            QTimer::singleShot(0, context, std::bind(functor, false, QString()));
+            return);
+    if (!passphraseFile.isEmpty() && !passphraseFile.exists()) {
+        QTimer::singleShot(0, context, std::bind(
+                functor, false,
+                tr("Passphrase file set but doesn't exit")));
+        return;
+    }
+
+    verifyPassphrase(id, passphraseFile.toString(), Sdk::instance(), [=](bool passphraseValid,
+            const QString &errorString) {
+        if (!passphraseValid) {
+            callIf(context_, functor, false, errorString);
+            return;
+        }
+
+        if (!outputDir.exists()) {
+            QDir dir(outputDir.toString());
+            if (!dir.mkdir(dir.absolutePath())) {
+                callIf(context_, functor, false, tr("Failed to create folder for keys: %1").arg(outputDir.toString()));
+                return;
+            }
+        }
+
+        const QString keyFilePath = outputDir.pathAppended(id + SIGNING_KEY_EXTENSION).toString();
+
+        QStringList arguments;
+        arguments.append("--batch");
+        arguments.append("--yes"); //need for overwrite key file if it exists
+        arguments.append("--output");
+        arguments.append(keyFilePath);
+        arguments.append("--pinentry-mode=loopback");
+
+        if (!passphraseFile.isEmpty()) {
+            arguments.append("--passphrase-file");
+            arguments.append(passphraseFile.toString());
+        }
+
+        arguments.append("--export-secret-keys");
+        arguments.append(id);
+
+        auto runner = std::make_unique<GpgRunner>(arguments);
+        QObject::connect(runner.get(), &GpgRunner::done, context, [=, process = runner->process()](bool ok) {
+             if (!ok) {
+                 callIf(context_, functor, false, tr("Failed to export secret \"%1\" key: %2")
+                         .arg(id)
+                         .arg(QString::fromUtf8(process->readAllStandardError())));
+                 return;
+             }
+
+             callIf(context_, functor, true, QString());
+        });
+        commandQueue()->enqueueImmediate(std::move(runner));
+    });
+}
+
+void SigningUtils::exportPublicKey(const QString &id, const QObject *context,
+        const Functor<bool, const QByteArray &, QString> &functor)
+{
+    QTC_ASSERT(!id.isEmpty(),
+            QTimer::singleShot(0, context, std::bind(functor, false, QByteArray(), QString()));
+            return);
+
+    QStringList arguments;
+    arguments.append("--batch");
+    arguments.append("--armor");
+    arguments.append("--export");
+    arguments.append(id);
+
+    auto runner = std::make_unique<GpgRunner>(arguments);
+
+    QObject::connect(runner.get(), &GpgRunner::done, context,
+            [id, functor, process = runner->process()](bool ok) {
+        if (!ok) {
+            functor(false, {}, tr("Failed to export public key for \"%1\": %2")
+                    .arg(id, QString::fromUtf8(process->readAllStandardError())));
+            return;
+        }
+        functor(true, process->readAllStandardOutput(), {});
+    });
+    commandQueue()->enqueue(std::move(runner));
+}
+
+bool SigningUtils::isGpgAvailable(QString *errorString)
+{
+    const FilePath gpg = FilePath::fromString(gpgPath());
+    if (!gpg.exists()) {
+        *errorString = tr("GPG utility not found");
+        return false;
+    }
+    return true;
+}
+
+QList<GpgKeyInfo> SigningUtils::gpgKeyInfoListFromOutput(const QString &output)
+{
+    if (output.isEmpty())
+        return {};
+
+    QList<GpgKeyInfo> keys;
+    QRegularExpression regex("(fpr:*(.*)):\ngrp(.*)\n(uid:u(.*))");
+    QRegularExpressionMatchIterator i = regex.globalMatch(output);
+
+    while (i.hasNext()) {
+        QRegularExpressionMatch m = i.next();
+        GpgKeyInfo gpgKeyInfo;
+        for (int i = 1; i <= m.lastCapturedIndex(); ++i) {
+            QString captured = m.captured(i);
+
+            if (captured.startsWith("fpr", Qt::CaseInsensitive))
+                gpgKeyInfo.fingerprint = m.captured(i + 1);
+            if (captured.startsWith("uid", Qt::CaseInsensitive)) {
+                QStringList keyInfo = m.captured(i + 1)
+                                          .split("::", QString::SplitBehavior::SkipEmptyParts);
+                gpgKeyInfo.name = keyInfo.at(2);
+            }
+        }
+        keys << gpgKeyInfo;
+    }
+    return keys;
+}
+
+void SigningUtils::verifyPassphrase(const QString &id, const QString &passphraseFile,
+        const QObject *context, const Functor<bool, QString> &functor)
+{
+    QStringList arguments;
+    arguments.append("--batch");
+    arguments.append("--status-fd");
+    arguments.append("1");
+    arguments.append("--sign");
+    arguments.append("--local-user");
+    arguments.append(id);
+    arguments.append("--output");
+    arguments.append("/dev/null");
+    arguments.append("--pinentry-mode=loopback");;
+
+    if (!passphraseFile.isEmpty()) {
+        arguments.append("--passphrase-file");
+        arguments.append(passphraseFile);
+    }
+
+    arguments.append("/dev/null");
+
+    auto runner = std::make_unique<GpgRunner>(arguments);
+
+    QObject::connect(runner.get(), &GpgRunner::done, context,
+            [passphraseFile, functor, process = runner->process()](bool ok) {
+        QString errorString;
+        if (!ok) {
+            if (passphraseFile.isEmpty())
+                errorString = tr("The selected GPG key is passphrase protected"
+                              "and no passphrase was specified.");
+            else
+                errorString = tr("The given passphrase is not a correct passphrase"
+                              "for the selected key.");
+        }
+        functor(ok, errorString);
+    });
+    commandQueue()->enqueue(std::move(runner));
+}
+
+} // namespace Sfdk
+
+#include "signingutils.moc"
--- a/src/plugins/mer/mer.pro
+++ b/src/plugins/mer/mer.pro
@@ -80,6 +80,7 @@ SOURCES += \
     mersdkkitaspect.cpp \
     mersdkmanager.cpp \
     mersettings.cpp \
+    mersigninguserselectiondialog.cpp \
     mersshparser.cpp \
     mertargetmanagementdialog.cpp \
     mertoolchain.cpp \
@@ -136,6 +137,7 @@ HEADERS += \
     mersdkkitaspect.h \
     mersdkmanager.h \
     mersettings.h \
+    mersigninguserselectiondialog.h \
     mersshparser.h \
     mertargetmanagementdialog.h \
     mertoolchain.h \
--- a/src/plugins/mer/merbuildconfigurationaspect.cpp
+++ b/src/plugins/mer/merbuildconfigurationaspect.cpp
@@ -202,7 +202,10 @@ const QStringList MerBuildConfigurationAspect::s_allowedSfdkOptions{
     "no-task",
     "fix-version",
     "no-fix-version",
-    "no-pull-build-requires"
+    "no-pull-build-requires",
+    "package.signing-user",
+    "package.signing-passphrase",
+    "package.signing-passphrase-file"
 };
 
 MerBuildConfigurationAspect::MerBuildConfigurationAspect(BuildConfiguration *buildConfiguration)
--- a/src/plugins/mer/merbuildconfigurationaspect.cpp
+++ b/src/plugins/mer/merbuildconfigurationaspect.cpp
@@ -24,22 +24,33 @@
 
 #include "ui_merbuildconfigurationwidget.h"
 #include "merconstants.h"
+#include "mersdkkitaspect.h"
+#include "mersettings.h"
+#include "mersigninguserselectiondialog.h"
 
+#include <sfdk/buildengine.h>
 #include <sfdk/sdk.h>
 #include <sfdk/sfdkconstants.h>
+#include <sfdk/utils.h>
 
+#include <coreplugin/messagemanager.h>
 #include <coreplugin/variablechooser.h>
 #include <projectexplorer/buildconfiguration.h>
+#include <projectexplorer/buildsystem.h>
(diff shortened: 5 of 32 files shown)


martyone:
We are currently in the process of upgrading to Qt Creator 4.12 - conflicts are unavoidable :\ I will ping you when this is done and our master is ready for new features.

martyone:
> We are currently in the process of upgrading to Qt Creator 4.12 - conflicts are unavoidable :\ I will ping you when this is done and our master is ready for new features.

The master branch is now on 4.12, you can rebase.

etaishev:
> > We are currently in the process of upgrading to Qt Creator 4.12 - conflicts are unavoidable :\ I will ping you when this is done and our master is ready for new features.
> 
> The master branch is now on 4.12, you can rebase.

Updated PR

martyone:
Under https://github.com/sailfishos/sdk-setup/pull/179 we are also dropping implication of "make-install" step during "deploy --rsync". So there should be similar split done to "MerMb2RsyncDeployStep" here unless I missed something.

etaishev:
> I would prefer squashing the two commits
> 
> * Mer: Add show dialog options for MerMb2RpmBuildStep
> * Mer: Split the deploy step to build and install rpm
> 
> into single commit: "Mer: Split the deploy step to build and deploy RPM" and mention in the commit message that "The 'deploy' command of mb2 used to imply package creation when ran behind Qt Creator. Nowadays the 'package' command needs to be used explicitly."
> 
> There should be a checkbox in the "RPM Build" step's config widget to control the "Sign" option, and another checkbox under Options > Sailfish OS > General to control the default.
> 
> Why is the export needed more exactly?
> 
> Is it really needed to move MerDeployStepWidget into its own file? At the same time you are exporting another class (MerProcessStep) from the original file.

Move MerDeployStepWidget is small refactoring -> `merdeploysteps.h` is so big and mix UI and logic.

Adding export useful for us, so that we can implement own custom build step, but if you have other good ideas on how to insert a build step into the process I would be happy to see them

martyone:
> Move MerDeployStepWidget is small refactoring -> `merdeploysteps.h` is so big and mix UI and logic.

It's a tiny class, closely related to the others, so I would prefer keeping them together.

> Adding export useful for us, so that we can implement own custom build step, but if you have other good ideas on how to insert a build step into the process I would be happy to see them

Every exported class is a promise of API stability. This is something I am reluctant to do as the API is still awaiting a significant refactor. So please keep the exports on your side for the now.

May I review your custom build steps implemented on top of these exports? Maybe there are parts generic enough for inclusion on our side.

etaishev:
> May I review your custom build steps implemented on top of these exports? Maybe there are parts generic enough for inclusion on our side.

I add you to prive repo with my plugin.

martyone:
> I add you to prive repo with my plugin.

OK, I will check. Consider adding Ville as well so that we can discuss conveniently.

src/plugins/mer/merdeploysteps.cpp
```diff
@@ -93,6 +94,8 @@ namespace {
 const int CONNECTION_TEST_CHECK_FOR_CANCEL_INTERVAL = 2000;
 
 const char MER_RPM_VALIDATION_STEP_SELECTED_SUITES[] = "MerRpmValidationStep.SelectedSuites";
+const char MER_RPM_BUILD_STEP_SIGN[] = "MerRpmBuildStep.SignEnable";
```
martyone, line 97: name and value would be better matching

```suggestion
const char MER_RPM_BUILD_STEP_SIGN_PACKAGES[] = "MerRpmBuildStep.SignPackages";
```

src/plugins/mer/merdeploysteps.cpp
```diff
@@ -732,6 +737,23 @@ void MerMb2RpmBuildStep::doRun()
     emit addOutput(tr("Building RPM package..."), OutputFormat::NormalMessage);
     AbstractProcessStep::doRun();
 }
+
+bool MerMb2RpmBuildStep::fromMap(const QVariantMap &map)
+{
+    if (!MerProcessStep::fromMap(map))
+        return false;
+
+    m_enableSignPackages = map.value(MER_RPM_BUILD_STEP_SIGN, m_enableSignPackages).toBool();
```
martyone, line 748: ditto

```suggestion
    m_signPackages = map.value(MER_RPM_BUILD_STEP_SIGN_PACKAGES, m_signPackages).toBool();
```

src/plugins/mer/merdeploystepwidget.h
```diff
@@ -0,0 +1,47 @@
...
+QT_END_NAMESPACE
+
+namespace Mer {
+namespace Internal {
+
+namespace Ui {
+    class MerDeployStepWidget;
+} // namespace Ui
+
+class MerDeployStepWidget : public ProjectExplorer::BuildStepConfigWidget
```
martyone, line 18: I would prefer moving this back to merdeploysteps.h/cpp.

src/plugins/mer/merdeploystepwidget.h
```diff
@@ -0,0 +1,47 @@
...
+    QString commnadText() const;
+    void setCommandText(const QString& commandText);
+    void formatAndSetSummaryText(const QString &summaryText);
+private slots:
+    void commandArgumentsLineEditTextEdited();
+protected:
+    Ui::MerDeployStepWidget *m_ui;
+};
+
+class MerRpmBuildStepWidget : public MerDeployStepWidget
```
martyone, line 32: I would prefer implementing this in merdeploysteps.h/cpp

src/plugins/mer/merdeploystepwidget.h
```diff
@@ -0,0 +1,47 @@
...
+{
+    Q_OBJECT
+public:
+    MerRpmBuildStepWidget(MerProcessStep *step);
+    void setSign(bool sign);
+signals:
+    void enableSign(bool sign);
+private:
+    QCheckBox *m_signCheckBox;
```
martyone: naming

```suggestion
    void setSignPackages(bool sign);
signals:
    void signPackagesChanged(bool sign);
private:
    QCheckBox *m_signPackagesCheckBox;
```

src/plugins/mer/mer.pro
```diff
@@ -151,3 +151,5 @@ HEADERS += \
     mervirtualmachinesettingswidget.h \
     mercmakebuildconfiguration.h \
     merqmakebuildconfiguration.h
+
+!isEmpty(MER_ENABLE_SIGN): DEFINES += MER_ENABLE_SIGN
```
martyone, line 155: What would be the reason to exclude the signing functionality?

martyone (commented):
Rebase is needed.

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -522,5 +525,117 @@ bool MerHardwareDeviceWizardSetupPage::validatePage()
...
+    setTitle(tr("RPM Key Deployment"));
+    setSubTitle(QLatin1String(" "));
+
+    auto infoLabel = new QLabel;
+    infoLabel->setText(tr("Install the gpg developer key to verify the signature of rpm packages.\n"
+        "The installation confirmation window will not be displayed if packages signed with known keys."));
+
+    m_keyPathChoother = new Utils::PathChooser;
+    m_keyPathChoother->setExpectedKind(PathChooser::Kind::File);
+    m_keyPathChoother->setPromptDialogTitle(tr("Select the public key file for the sign key"));
```
martyone, line 579: "Select a Trusted GnuPG Public Key"

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -522,5 +525,117 @@ bool MerHardwareDeviceWizardSetupPage::validatePage()
...
+    m_keyPathChoother->setToolTip(tr("Public key for verify packages"));
```
martyone, line 580: "Select a trusted GnuPG public key"

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -522,5 +525,117 @@ bool MerHardwareDeviceWizardSetupPage::validatePage()
     return true;
 }
```
martyone, line 527:
```suggestion

/*!
 * \class PackageKeyDeployer
 * \internal
 */

```

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -522,5 +525,117 @@ bool MerHardwareDeviceWizardSetupPage::validatePage()
...
+                "|| mkdir -p %1 && chmod 0700 %1 && echo '"
+                    + QString::fromLocal8Bit(reader.data())
+                + "' >> %1/%2 && sdk-deploy-key %1/%2; rm %1/%2";
+        runner.run(command.arg(dirName, QFileInfo(keyFilePath).fileName()), sshParams);
+        loop.exec();
+    }
+signals:
+    void finished(bool ok, const QString &errorString);
+};
+
```
martyone, line 566:
```suggestion

/*!
 * \class MerHardwareDeviceWizardPackageKeyDeploymentPage
 * \internal
 */

```

share/qtcreator/sfdk/modules/20-building-mb2.json
```diff
@@ -152,6 +152,21 @@
         {
             "name": "no-pull-build-requires",
             "trDescription": "Do not update build-time dependencies unless required explicitly with 'build-requires pull'."
+        },
+        {
+            "name": "package-signing-user",
```
martyone, line 157:
```suggestion
            "name": "package.signing-user",
```

Similarly in the other two cases.

src/plugins/mer/merbuildconfigurationaspect.cpp
```diff
@@ -202,7 +202,10 @@ const QStringList MerBuildConfigurationAspect::s_allowedSfdkOptions{
     "no-task",
     "fix-version",
     "no-fix-version",
-    "no-pull-build-requires"
+    "no-pull-build-requires",
+    "package-signing-user",
+    "package-signing-passphrase",
+    "package-signing-passphrase-file"
```
martyone, line 208: ditto

martyone (commented):
Typo in commit title "Sfdk: Add sign packeges options"

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -522,5 +525,117 @@ bool MerHardwareDeviceWizardSetupPage::validatePage()
...
+                emit finished(false, tr("Deploy key error: %1").arg(QString::fromLatin1(runner.readAllStandardError())));
+            loop.quit();
+        });
+
+        const QString dirName = "~/.sdk";
+        const QString command = "test -d %1 "
+                "|| mkdir -p %1 && chmod 0700 %1 && echo '"
+                    + QString::fromLocal8Bit(reader.data())
+                + "' >> %1/%2 && sdk-deploy-key %1/%2; rm %1/%2";
+        runner.run(command.arg(dirName, QFileInfo(keyFilePath).fileName()), sshParams);
```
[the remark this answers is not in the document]
etaishev: With the "-" option, the sdk-deploy-key must wait for the key to be entered in stdin, right?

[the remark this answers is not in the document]
martyone: yup, that's the common convention that you can pass '-' in place of a file name and the command will read stdin instead.

BTW, you *may* need to also somehow close the input channel after writeDataToProcess (you will see whether it hangs or not)

src/plugins/mer/merdeploysteps.cpp
```diff
@@ -733,6 +745,23 @@ bool MerMb2RpmBuildStep::init()
...
+        QStringList packageUserSigningArgs;
+        if (!m_signPackageKey.isEmpty())
+            packageUserSigningArgs.append(QStringLiteral("-c") + QtcProcess::quoteArgUnix("package-signing-user=%1").arg(m_signPackageKey));
+        if (!m_signPackagePassFile.isEmpty())
+            packageUserSigningArgs.append(QStringLiteral("-c") + QtcProcess::quoteArgUnix("package-signing-passphrase-file=%1").arg(m_signPackagePassFile));
+
+        if (!packageUserSigningArgs.empty()) {
+            auto env = pp->environment();
+            env.appendOrSet(QLatin1String(Sfdk::Constants::MER_SSH_SFDK_OPTIONS), packageUserSigningArgs.join(' '));
+            pp->setEnvironment(env);
```
martyone, line 761: This looks really hacky so I checked a bit deeper and I realized all the signing related configuration should be implemented in MerBuildConfigurationAspect/Widget instead. What should remain here is just something like:

```
    auto aspect = buildConfiguuration()->aspect<MerBuildConfigurationAspect>();
    QTC_ASSERT(aspect, return false);
    if (aspect->isSignPackagesEnabled())
        rpmCommand.addArgs("--sign", CommandLine::Raw);
```

src/plugins/mer/merdeploysteps.h
```diff
@@ -265,10 +277,31 @@ class MerDeployStepWidget : public ProjectExplorer::BuildStepConfigWidget
...
+    void signPackagePassFileChanged(const QString &file);
+private:
+    void initKeyChooser();
+    void initKeyPassphraseChooser();
+    QCheckBox *m_signPackagesCheckBox;
+    QLineEdit *m_signPackageKeyEdit;
+    Utils::PathChooser *m_signPackageKeyPassFileChooser;
+    Sfdk::BuildEngine *m_engine;
+};
+
```
martyone, line 304: ditto, should be implemented in MerBuildConfigurationAspect/Widget instead.

etaishev: For Rsync deploy step, signing is not necessary, but build configuration is common for all methods. Wouldn't that be weird?

martyone: rsync deployment is rather a hack that needs to be used in cooperation with regular RPM deployment. Anyway it is the "package" step, not the "deploy" step what is controlled by those options, so it's all OK.

src/plugins/mer/merbuildconfigurationaspect.cpp
```diff
@@ -120,6 +129,63 @@ class MerBuildConfigurationWidget : public ProjectExplorer::NamedWidget
...
+
+        auto completer = new QCompleter(keys, this);
+        completer->setCaseSensitivity(Qt::CaseInsensitive);
+        m_ui->signPackageKeyLineEdit->setCompleter(completer);
+
+        m_ui->signPackageKeyPassFileChooser->setExpectedKind(PathChooser::Kind::File);
+        m_ui->signPackageKeyPassFileChooser->setPromptDialogTitle(tr("Select the passphrase file for the sign key"));
+        const QString toolTip = tr("Passphrase file for the sign key");
+        m_ui->signPackageKeyPassFileChooser->lineEdit()->setPlaceholderText(toolTip);
+        m_ui->signPackageKeyPassFileChooser->setToolTip(toolTip);
```
martyone, line 165: `signKeysFolder` is a location internal to the SDK. It is not a place the user should use as a primary storage for their keys - it will be removed during SDK uninstallation. The user should keep their keys stored on the regular GPG keyring and then:

> When the package.signing-user <name> option is set, sfdk should ensure that the corresponding private key is exported as ~/SailfishSDK/vmshare/gnupg/<name>.key prior to executing a command on the build engine.

(quoted from sailfishos/sdk-setup#179)

So what we should offer here to choose from is what we get with `gpg --list-secret-keys`. I would start with a combo box with items in formatted as `NAME <EMAIL> [SHORT_ID]`.

I think we have no other option than to pass the ID as `package.signing-user` (which should accept exactly the same values as gpg's `--local-user` option). NAME+EMAIL may not be unique. But it would be more user friendly when seen in build output, so it would be great if all of ID+NAME+EMAIL could be combined in argument to gpg's `--local-user` option somehow. The documentation does not mention the details, the implementation does not seem to allow that after brief check - maybe you spot a possibility there https://github.com/gpg/gnupg/blob/357ad9ae29677c1676b56d2b81282e2f78ec8040/common/userids.c#L41

src/plugins/mer/merbuildconfigurationwidget.ui
```diff
@@ -54,6 +54,62 @@
      </property>
     </widget>
    </item>
+   <item row="2" column="0" colspan="2">
+    <widget class="QGroupBox" name="signPackagesGroupBox">
```
martyone, line 58: It would be more natural if `Additional sfdk configuration:` came at the very end. I would also try to go without (visible) group box. Just something like:

```
RPM spec file:                 [                                            ] [Browse...]
[X] Sign packages
        Signing user:          [                                           v]
        Passphrase file:       [                                            ] [Browse...]
Additional sfdk configuration: [                                            ]

File: src/plugins/mer/merbuildconfigurationaspect.cpp (hunk 16 above)
Comment (etaishev) [reply src/plugins/mer/merbuildconfigurationaspect.cpp:165]: So in sfdk we have to call `gpg --export-secret-keys <key-name>` on the developer host and then use it?

Comment (martyone) [reply src/plugins/mer/merbuildconfigurationaspect.cpp:165]: yup

Comment (etaishev) [reply src/plugins/mer/merbuildconfigurationaspect.cpp:165]: Hm, on Mac OS no `gpg`  by default and on Windows it's will be avaliabel only if running QtCreator under GitBash

Comment (martyone) [reply src/plugins/mer/merbuildconfigurationaspect.cpp:165]: I wouldn't see it as granted that gpg is available even on Linux by default. This is not an issue. Want to gpg sign? - Install gpg first! :)

On Windows you mean it is not in PATH I believe. We can locate it the same way as we do for SSH - sitting in the same dir as git: https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/plugins/projectexplorer/projectexplorer.cpp#L1992

Comment (etaishev) [reply src/plugins/mer/merbuildconfigurationaspect.cpp]: Add search git location because gpg installed with git bash

File: src/tools/sfdk/script.cpp [hunk 18]
```diff
@@ -260,6 +262,32 @@ class JSBuildEngineExtension : public QObject
...
+    {
+        if (!SdkManager::hasEngine()) {
+            qCWarning(sfdk).noquote() << SdkManager::noEngineFoundMessage();
+            return false;
+        }
+
+        QString keyName;
+        QString keyPassFile;
+        // TODO: HACK ??
+        for (const auto &option : Configuration::effectiveState()) {
```
etaishev, line 280: @martyone what do you think about exporting key here and getting the arguments?

src/libs/sfdk/packagesignkeys.cpp
```diff
@@ -0,0 +1,238 @@
...
+using Utils::FilePath;
+using Utils::FilePaths;
+
+namespace {
+const char GPG_NOT_FOUND_ERROR[] = "PackageSign: Gpg not found.";
+}
+
+namespace Sfdk {
+
+class GpgRunner
```
martyone, line 29: This needs to be implemented a non blocking way. Check e.g. how docker or vboxmanage is used in {docker,vbox}virtualmachine.cpp.

src/tools/sfdk/script.cpp
martyone: I think we should instead add

```
void BuildEngine::importGpgKey(const QString &id, const Utils::FilePath &passphraseFile, const QObject *context,
```

expose it on `JSBuildEngineExtension` with help of `execAsynchronous` and implement the rest in javascript:

```javascript
export function maybeImportSigningKey() {
    if (!configuration.isOptionSet("package.signing-user"))
        return [true, ""];

    var signingUser = configuration.optionArgument("package.signing-user");
    var signingPassphraseFile = configuration.isOptionSet("package.signing-passphrase-file")
        ? configuration.optionArgument("package.signing-passphrase-file")
        : "";
    if (!engine.importGpgKey(signingUser, signingPassphraseFile))
        return [false, qsTr("Failed to share GnuPG key with the build engine")];

    return [true, ""];
}
```

We should also think about being more verbose about the reasons for a failure - gpg may print something useful on stderr.

src/libs/sfdk/packagesignkeys.cpp
etaishev: Why is this needed? Why can't we do with a non-blocking completion wait in QEventLoop?
For few gpg calls looks like a excess mechanism.

martyone: Sfdk API is meant to be 100% asynchronous - no blocking operation that could freeze the UI inadvertently. This is well supported internally, so the implementation overhead is too negligible to bother with the question whether this particular external command may or may not ever block our application. Use of nested event loops on library side is a asking for issues and should be avoided when possible.

src/libs/sfdk/signingutils.cpp
```diff
@@ -0,0 +1,374 @@
...
+        arguments << "--passphrase-file" << passphraseFile.toString();
+
+    arguments << "--export-secret-keys" << fingerprint;
+
+    auto runner = std::make_unique<GpgRunner>(arguments);
+    if (runner->process()->program().isEmpty()) {
+        m_lastError = GPG_NOT_FOUND_ERROR;
+        QTimer::singleShot(0, context, std::bind(functor, false));
+        return;
+    }
```
martyone, line 176: This is too late to check for this. We should never enter this function if gpg was not found. All GPG related UIs should be disabled if GPG executable is not available. Simply drop this and the other occurrences.

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -522,5 +529,153 @@ bool MerHardwareDeviceWizardSetupPage::validatePage()
...
+    m_keyComboBox->setToolTip(tr("Select a trusted GnuPG public key"));
+
+    SigningUtils signUtils;
+    bool ok{false};
+    execAsynchronous(std::tie(ok, p->gpgKeyInfoList), std::mem_fn(&SigningUtils::availableKeys), signUtils);
+    if (!ok)
+        Core::MessageManager::write(signUtils.error(), Core::MessageManager::Flash);
+
+    for (const auto &keyInfo : p->gpgKeyInfoList)
+        m_keyComboBox->addItem(keyInfo.name);
```
martyone, line 615: Here I would do similarly as I suggested for build settings: insert items in the form "NAME [FINGERPRINT]" to ensure usability in case of multiple keys for single NAME.

src/libs/sfdk/signingutils.h
```diff
@@ -0,0 +1,49 @@
...
+    QString error() const;
+
+protected:
+    SigningUtils(std::unique_ptr<SigningUtilsPrivate> &&d, QObject *parent);
+    std::unique_ptr<SigningUtilsPrivate> d_ptr;
+
+private:
+    Q_DECLARE_PRIVATE(SigningUtils);
+    QString m_error;
+};
```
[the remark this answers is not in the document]
etaishev: If we take out Sfdk::Device::importGpgKey we lose the ability to set the key in MerHardwareDeviceWizardPackageKeyDeploymentPage when adding a device. I think this is more convenient than a separate step and we need to keep this feature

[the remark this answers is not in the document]
martyone: I am not sure I follow what you mean, but as I see it most of the code that is now part of `PackageKeyDeployerPrivate::deploy` would be moved under `Sfdk::Device::importGpgKey(const QString &id, const QObject *context, const Functor<bool, QString> &functor)` which can use `SigningTools` internally.

`MerHardwareDeviceWizardPackageKeyDeploymentPage::deployKey()` would than directly do

```
const QString id = gpgKeyInfoList.at(m_keyComboBox->currentIndex()).fingerprint;
m_device->importGpgKey(id, this, [&](bool ok, const QString &errorString) {
    /* same as now */
})
```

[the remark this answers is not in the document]
martyone: Now I see that `m_device` is a `MerHardwareDevice` and `Sfdk::HardwareDevice` does not exit yet at that point. - Is that what you meant?

This is not going to change easily :\ OK, let's create `static Sfdk::Device::importGpgKey(const QSsh::SshConnectionParameters &sshParameters, const QString &id, const QObject *context, const Functor<bool, QString> &functor)` as a workaround then and let `MerHardwareDeviceWizardPackageKeyDeploymentPage::deployKey()` do

```
const QString id = gpgKeyInfoList.at(m_keyComboBox->currentIndex()).fingerprint;
HardwareDevice::importGpgKey(m_device->sshParameters(), id, this,
        [&](bool ok, const QString &errorString) {
    /* same as now */
})
```

This will allow us to follow the long term objective of minimizing the amount of code in the Mer plugin and keeping as much as possible inside the Sfdk library. Please mark it with "TODO Make non-static".

src/plugins/mer/mergeneraloptionswidget.cpp
```diff
@@ -86,6 +90,45 @@ MerGeneralOptionsWidget::MerGeneralOptionsWidget(QWidget *parent)
...
+
+    connect(m_ui->signPackagesByDefaultCheckBox, &QCheckBox::toggled, this, [&](bool checked) {
+        setSignElementsEnabled(m_ui, checked);
+    });
+    m_ui->signPackagesByDefaultCheckBox->setChecked(MerSettings::signPackagesByDefault());
+
+    const bool isGpgAvailable = Sfdk::isGpgAvailable();
+    if (!isGpgAvailable) {
+        m_ui->signErrorLabel->setText(tr("GPG utility not found!"));
+        m_ui->signErrorLabel->setStyleSheet(QLatin1String("color: \"yellow\""));
```
martyone, line 107: hm, I don't see any "yellow" usage in src/plugins. Maybe "red"? More occurrences.

src/plugins/mer/merhardwaredevicewizardpages.cpp
```diff
@@ -522,5 +533,194 @@ bool MerHardwareDeviceWizardSetupPage::validatePage()
...
+    void deploy(const MerHardwareDevice::Ptr &device, const QString &id)
+    {
+        QTC_ASSERT(device, emit deployFinished(false, tr("No device set.")); return );
+
+        bool ok;
+        QString error;
+        execAsynchronous(std::tie(ok, error),
+                         &HardwareDevice::importPublicGpgKey,
+                         device->sshParameters(),
+                         id);
```
martyone, line 653: Now I realized maybe we could create a temporary `Sfdk::HardwareDevice` instance just for the wizard purposes. I do not remember the details, so could you please try creating it like in `MerHardwareDeviceManager::onDeviceAddedOrUpdated` (just omit the `Sdk::addDevice` call) make `importPublicGpgKey` non-static and call it on the temporary instance? (If this works, it allows us moving more code from the wizard to the library easily, so it would be very nice! :))

src/plugins/mer/mergeneraloptionswidget.cpp
```diff
@@ -86,6 +90,45 @@ MerGeneralOptionsWidget::MerGeneralOptionsWidget(QWidget *parent)
...
+    bool ok;
+    QString error;
+    execAsynchronous(std::tie(ok, availableKeys, error), Sfdk::availableGpgKeys);
+    if (!ok)
+        Core::MessageManager::write(error, Core::MessageManager::Flash);
+
+    auto signingUsersModel = new SigningUserListModel(availableKeys);
+    m_ui->defaultSigningUserComboBox->setModel(signingUsersModel);
+    const int index = signingUsersModel->getIndexBy(MerSettings::signingUser());
+    m_ui->defaultSigningUserComboBox->setCurrentIndex(index);
```
martyone, line 124: With `GpgKeyInfo::toString/fromString` we could easily fill the combo box with strings and then get current with `GpgKeyIngo::fromString(combo->currentText())`. You know, the most reliable- and easiest to maintain- code is that one we never wrote, so if we can go without model... :)

src/tools/sfdk/script.cpp
```diff
@@ -242,6 +242,19 @@ class JSBuildEngineExtension : public QObject
         });
     }
 
+    Q_INVOKABLE bool importGpgKey(const QString &signingUser, const QString &signingPassphraseFile)
+    {
+        return withEngine([=](BuildEngine *engine) {
+            QString error;
+            bool ok;
+            execAsynchronous(std::tie(ok, error), std::mem_fn(&BuildEngine::importPrivateGpgKey), engine, signingUser, FilePath::fromString(signingPassphraseFile));
```
martyone, line 250: soooo long line :)

src/libs/sfdk/device.cpp
```diff
@@ -115,6 +117,52 @@ Device::MachineType Device::machineType() const
...
+            loop.quit();
+        });
+        connect(&runner, &SshRemoteProcessRunner::processStarted, [&runner, &keyData]() {
+            runner.writeDataToProcess(keyData);
+        });
+        runner.run("sdk-deploy-key -", sshParameters);
+        loop.exec();
+
+        if (context_)
+            functor(ok, error);
```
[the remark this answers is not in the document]
martyone: Edited suggested change - use `enqueueImmediate` instead of `enqueue`.

src/plugins/mer/mergeneraloptionswidget.cpp
etaishev: "red" looks like a fatal error. Without gpg, we can safely continue using SDK. Maybe should hide items if gpg is not available?

vigejolla: Red indeed looks like a fatal error. Hiding is not good either, as the user should be given a hint that it might be a good idea to install gpg. Perhaps something like Theme::IconWarningColor.name(QColor::HexRgb) or something like that could be used?

martyone: Without GPG, we can safely continue using SDK but we cannot use GPG and that's what this part of UI is about :)

Red highlight is what upstream QtC uses consistently. Check e.g. Options > Version Control - try to enter invalid path in any field.

Alternatively, we could also show the label using default color and show the warning icon in front of it.

And we should also think about adding path chooser here, which already provides the desired feedback, but I would prefer to postpone that as we should ideally merge this before the QtC upgrade work is started.

src/plugins/mer/mersigninguserselectiondialog.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+MerSigningUserSelectionDialog::MerSigningUserSelectionDialog(QObject *parent)
+    : QObject(parent)
+{}
+
+GpgKeyInfo MerSigningUserSelectionDialog::selectSigningUser()
+{
+    QList<GpgKeyInfo> availableKeys;
+    QString gpgErrorString;
+    if (Sfdk::isGpgAvailable(gpgErrorString)) {
```
martyone, line 36: Swap condition, return, omit `else`, declare and initialize `const availableKeys` later in one step.

src/plugins/mer/mersigninguserselectiondialog.cpp
```diff
@@ -0,0 +1,73 @@
...
+        availableKeys = getAvailableKeys();
+    } else {
+        QMessageBox::warning(nullptr, tr("No GPG GPG utility!"),
+                             gpgErrorString);
+        return GpgKeyInfo();
+    }
+
+    QStringList items = Utils::transform(availableKeys, [](const GpgKeyInfo &info) {
```
martyone, line 44: ditto, &GpgKeyInfo::toString, const items

src/plugins/mer/mersigninguserselectiondialog.cpp
```diff
@@ -0,0 +1,73 @@
...
+                                             false,
+                                             &ok);
+    if (!ok || selected.isEmpty())
+        return GpgKeyInfo();
+
+    const int index = items.indexOf(selected);
+    if (index > -1 && index < availableKeys.count())
+        return availableKeys.at(index);
+
+    return GpgKeyInfo();
```
martyone, line 69: case for assert

```suggestion
    QTC_ASSERT(index > -1 && index < availableKeys.count(), return {});
    return availableKeys.at(index);
```

src/plugins/mer/mersigninguserselectiondialog.h
```diff
@@ -0,0 +1,23 @@
...
+namespace Mer {
+namespace Internal {
+
+class MerSigningUserSelectionDialog : public QObject
+{
+    Q_OBJECT
+public:
+    explicit MerSigningUserSelectionDialog(QObject *parent = nullptr);
+    static Sfdk::GpgKeyInfo selectSigningUser();
+signals:
```
martyone, line 17: garbage

src/plugins/mer/mersigninguserselectiondialog.h
```diff
@@ -0,0 +1,23 @@
+#ifndef MERSIGNINGUSERSELECTIONDIALOG_H
+#define MERSIGNINGUSERSELECTIONDIALOG_H
+
+#include <sfdk/utils.h>
+#include <QAbstractListModel>
```
martyone, line 5: garbage

martyone (commented):
This looks quite right already! :)

Please try to check for and fix the following style issues everywhere:

- long lines (100 chars soft limit, 120 hard limit, exceptions apply :))
- `return {}` instead of `return FooBar()` wherever possible
- prefer implicit capture in lambdas
- argument list wrapping - prefer simple wrapping without aligning under opening bracket

Typo in commit message "packeges".

I never checked the distribution of changes into commits - is that right?

Great to see the temporary device instance could be used!

src/plugins/mer/merbuildconfigurationwidget.ui
```diff
@@ -41,19 +41,126 @@
...
+        <string>Passphrase file:</string>
+       </property>
+      </widget>
+     </item>
+    </layout>
+   </item>
+   <item row="2" column="1">
+    <layout class="QHBoxLayout" name="signingUserLayout_2">
+     <item>
+      <layout class="QHBoxLayout" name="horizontalLayout">
```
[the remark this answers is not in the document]
martyone: How about this?

src/plugins/mer/mergeneraloptionswidget.cpp
vigejolla: > Alternatively, we could also show the label using default color and show the warning icon in front of it.

Indeed. Now I feel a bit silly as when I looked at what upstream QtC uses in such situations, I saw the warning icon. That's why I wrote about using IconWarningColor. I feel so stupid for not realizing that we could also use the icon!

martyone (commented):
The whitespace code would still profit from some more love :) I will really appreciate if you improve it but I am also not going to delay the review anymore on that.

I would go with the following commit messages for the last three commits:

Sfdk: Allow to sign packages
Mer: Allow to sign packages
Mer: Allow to deploy public GPG keys to devices

Please add "Co-authored-by: Martin Kampas <martin.kampas@jolla.com>" footers to the last three commits.

Please ensure that the changes are properly tested. Given the schedule of upcoming work I am going to test it myself only later after it is integrated.

src/libs/sfdk/device.cpp
```diff
@@ -115,6 +119,40 @@ Device::MachineType Device::machineType() const
...
+        if (!ok) {
+            callIf(context_, functor, ok, exportError);
+            return;
+        }
+
+        QStringList arguments;
+        arguments.append("-");
+
+        auto runner = std::make_unique<RemoteProcessRunner>(sshParameters(),
+                "sdk-deploy-key",
```
etaishev, line 142: I think it would be better to add an argument, such as `-k --key and -f --file` instead of `-`, to use via ssh
@martyone ?

martyone: The acceptance of either file name or '-' as a positional argument follows the conventions used by coreutils, so I would keep it as it. We could make the argument optional though, reading from stdin whether it is '-' or no file is passed at all. And allow to pass multiple keys at once . Not necessarily, just an idea.

martyone (commented):
Looks beautiful now! Thank you!

src/plugins/mer/mergeneraloptionswidget.cpp
martyone: @etaishev it would be nice to fix this one yet

src/plugins/mer/merbuildconfigurationaspect.cpp
```diff
@@ -120,6 +134,84 @@ class MerBuildConfigurationWidget : public ProjectExplorer::NamedWidget
...
+
+        m_ui->signPackagesCheckBox->setText(tr(Constants::MER_SIGN_PACKAGES_OPTION_NAME));
+        m_ui->signingUserClearButton->setIcon(Utils::Icons::EDIT_CLEAR.icon());
+
+        m_ui->signPackagesCheckBox->setChecked(m_aspect->signPackages());
+        QString gpgErrorString;
+        const bool isGpgAvailable = Sfdk::isGpgAvailable(&gpgErrorString);
+        if (!isGpgAvailable) {
+            m_ui->signErrorLabel->setText(gpgErrorString);
+            m_ui->signErrorIconLabel->setPixmap(Icons::CRITICAL.pixmap());
```
martyone, line 153: Wasn't the intention to let it look less critical? :) Maybe Icons::WARNING would be enough?

martyone (commented):
I think we should put the warning into its own row to avoid spacing issues. Localized strings may be much longer.

Great that you took emitDone into use after rebase :)

src/plugins/mer/mergeneraloptionswidget.ui
```diff
@@ -242,42 +242,130 @@
...
+           <widget class="QLabel" name="signErrorLabel">
+            <property name="text">
+             <string/>
+            </property>
+           </widget>
+          </item>
+         </layout>
+        </item>
+        <item>
+         <layout class="QHBoxLayout" name="horizontalLayout_6">
```
martyone, line 277: This hbox seems needless now.

share/qtcreator/sfdk/modules/20-building-mb2.mjs
```diff
@@ -95,3 +95,17 @@ export function mapCompilationDatabasePaths() {
...
+
+    var signingUser = configuration.optionArgument("package.signing-user");
+    var signingPassphraseFile = configuration.isOptionSet("package.signing-passphrase-file")
+        ? configuration.optionArgument("package.signing-passphrase-file")
+        : "";
+    if (!buildEngine.importGpgKey(signingUser, signingPassphraseFile))
+        return [false, qsTr("Failed to share GnuPG key with the build engine")];
+
+    return [true, ""];
+}
```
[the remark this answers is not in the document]
martyone: Hm, somehow this re-appeared.

martyone (approved):
LGTM. Please tell me when you feel it is properly tested (also WRT the changes to build steps) and I can start merging it. Thank you!

etaishev:
Also need to keep in mind the `sdk-deploy-key`

martyone:
> Also need to keep in mind the `sdk-deploy-key`

Hm, I somehow thought this was already fixed. OK, in order to not delay merging the rest of this PR, let's split the "Mer: Allow to deploy public GPG keys to devices" commit out to another PR on top of the current one. It is the last commit, so it should be as easy as in

git branch jb50188-1 jb50188
git branch -f jb50188 jb50188^

Then push and open another PR from jb50188-1

etaishev:
Also put the changes from `Sfdk::Device` into a separate commit in `jb50188-1`

martyone:
OK, great! Just the 'Co-authored-by' footer is now missing from the "Sfdk: Allow to sign packages" commit. Don't forget to open PR for the other branch.

PR #328, opened 2020-06-23
Jb49035

martyone:


martyone:
Just the very last two commits belong to this PR

martyone:
> Just the very last two commits belong to this PR

Seven commits now

martyone:
Yeah, but the other side of the coin is that one would like to spot relations between closely related classes more easily. In Java, one would have classes organized into a tree of packages. In C++ this approach is not that common to see and so it is hard to push through (I was a big fan of that approach before, but gave out; and there were technical obstacles too, esp. on Windows). I remember in early days of Qt Creator development it was a coding convention to have a separate header file for each class. And we can see this rule faded out and they prefer expressing relations between classes by grouping their definitions. Most probably because coding assistance for c++ improved a lot in last decade - the IDE promptly navigates you to the class of your interest no matter where it is defined :)

PR #329, opened 2020-06-26
Sfdk: Map also paths with native separators

vigejolla:
Merssh passes paths to sfdk using native separators, which failed under
Windows

### DIFF
--- a/src/tools/sfdk/sdkmanager.cpp
+++ b/src/tools/sfdk/sdkmanager.cpp
@@ -1316,6 +1316,7 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
         QProcessEnvironment *environment) const
 {
     QTC_ASSERT(hasEngine(), return false);
+    *program = QDir::fromNativeSeparators(*program);
 
     const QString cleanSharedHome = this->cleanSharedHome();
     const QString cleanSharedSrc = this->cleanSharedSrc();
@@ -1348,15 +1349,37 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
         {cleanSharedSrc, Constants::BUILD_ENGINE_SHARED_SRC_MOUNT_POINT, caseInsensitiveOnWindows}
 #endif
     };
+
     for (const Mapping &mapping : mappings) {
         qCDebug(sfdk) << "Mapping" << mapping.hostPath << "as" << mapping.enginePath;
+        auto mapGreedy = [=](QString string) {
+            QRegularExpression::PatternOption reCaseSensitivity = mapping.cs == Qt::CaseInsensitive
+                    ? QRegularExpression::CaseInsensitiveOption
+                    : QRegularExpression::NoPatternOption;
+            QRegularExpression re(QString("(?:%1|%2)([^:;,'\"]*)")
+                                  .arg(QRegularExpression::escape(mapping.hostPath))
+                                  .arg(QRegularExpression::escape(QDir::toNativeSeparators(mapping.hostPath))),
+                                  reCaseSensitivity);
+            // Replace longer matches first! Use map for sorting.
+            std::map<QString, QString> replacements;
+            QRegularExpressionMatchIterator reIterator = re.globalMatch(string);
+            while (reIterator.hasNext()) {
+                QRegularExpressionMatch match = reIterator.next();
+                QString fullNativeHostPath = match.captured(0);
+                QString fullEnginePath = QDir::fromNativeSeparators(mapping.enginePath + match.captured(1));
+                replacements.insert(std::pair<QString, QString>(fullNativeHostPath, fullEnginePath));
+            }
+            for (auto i = replacements.crbegin(); i != replacements.crend(); ++i)
+                string.replace(i->first, i->second);
+            return string;
+        };
+
         program->replace(mapping.hostPath, mapping.enginePath, mapping.cs);
-        arguments->replaceInStrings(mapping.hostPath, mapping.enginePath, mapping.cs);
+        *arguments = Utils::transform(*arguments, mapGreedy);
         workingDirectory->replace(mapping.hostPath, mapping.enginePath, mapping.cs);
         for (const QString &key : environment->keys()) {
             QString value = environment->value(key);
-            value.replace(mapping.hostPath, mapping.enginePath, mapping.cs);
-            environment->insert(key, value);
+            environment->insert(key, mapGreedy(value));
         }
     }
 


src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1352,6 +1352,10 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
         qCDebug(sfdk) << "Mapping" << mapping.hostPath << "as" << mapping.enginePath;
         program->replace(mapping.hostPath, mapping.enginePath, mapping.cs);
         arguments->replaceInStrings(mapping.hostPath, mapping.enginePath, mapping.cs);
+        if (QDir::toNativeSeparators(mapping.hostPath) != mapping.hostPath) {
+            program->replace(QDir::toNativeSeparators(mapping.hostPath), mapping.enginePath, mapping.cs);
+            arguments->replaceInStrings(QDir::toNativeSeparators(mapping.hostPath), mapping.enginePath, mapping.cs);
+        }
```
martyone, line 1358: What if we have e.g. `C:\Users\user\foo\bar` in arguments and `C:\Users\user` is the shared home path. Wouldn't it result to `/home/mersdk/share\foo\bar`?

vigejolla: Indeed it did. Fixed now.

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1348,6 +1348,9 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
         {cleanSharedSrc, Constants::BUILD_ENGINE_SHARED_SRC_MOUNT_POINT, caseInsensitiveOnWindows}
 #endif
     };
+    *program = QDir::fromNativeSeparators(*program);
+    for (QString& argument : *arguments)
+         argument = QDir::fromNativeSeparators(argument);
```
martyone, line 1353: Nice but I am afraid it would be too generic - backslashes may be used in arguments for more purposes, not just as a directory separator.

vigejolla: > Nice but I am afraid it would be too generic - backslashes may be used in arguments for more purposes, not just as a directory separator.

That's what I'm afraid myself as well. But can we really know when the backslash is a directory separator and when it's not?

vigejolla: It looks like merssh used simply checking if there exists a file for the argument. I was afraid this would not work with paramaters like "--foo=C:\bar", but I guess we don't use parameters like that?

martyone: merssh have quite limited set of use cases. sfdk has those 'exec' and 'build-shell' commands, so we do not know what to expect. Maybe we could do something like

```
QRegularExpression re(QRegularExpression::escape(sharedHome) + "([^:;,'\"]*)");
arg.replace(re, sharedHomeMountPoint + "\\1");
```

i.e. replace everything that looks like a file path under sharedHome, up to a well known separators/terminators like `[:;,'"]`.

vigejolla: Now it's done in a similar fashion as was in merssh

vigejolla: > merssh have quite limited set of use cases. sfdk has those 'exec' and 'build-shell' commands, so we do not know what to expect. Maybe we could do something like
> 
> ```
> QRegularExpression re(QRegularExpression::escape(sharedHome) + "([^:;,'\"]*)");
> arg.replace(re, sharedHomeMountPoint + "\\1");
> ```
> 
> i.e. replace everything that looks like a file path under sharedHome, up to a well known separators/terminators like `[:;,'"]`.

Sounds like a good idea. I can do it, but I won't have time until tomorrow (I find it hard to write bug free code while in a meeting...)

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1348,10 +1348,22 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
...
+                ? QRegularExpression::CaseInsensitiveOption : QRegularExpression::NoPatternOption;
+        QRegularExpression re(QRegularExpression::escape(mapping.hostPath) + "([^:;,'\"]*)",
+                              reCaseSensitivity);
+        arguments->replaceInStrings(re, mapping.enginePath + "\\1");
+        QString nativeHostPath = QDir::toNativeSeparators(mapping.hostPath);
+        if (mapping.hostPath != nativeHostPath) {
+            QRegularExpression nativeRe(QRegularExpression::escape(nativeHostPath) + "([^:;,'\"]*)",
+                                        reCaseSensitivity);
+            arguments->replaceInStrings(nativeRe, QDir::fromNativeSeparators(mapping.enginePath + "\\1"));
+        }
```
martyone, line 1366: (Comment written before your last update)

We somehow forgot about replacing the backslashes in the captured group.

The most naive approach would be

```
auto map = [=](QString string) {
    while (string.contains(re, &match)) {
        QString replacement = mapping.enginePath + QDir::fromNativeSeparators(match.captured(1));
        return string.replace(match.captured(), replacement);
    }
};
*arguments = Utils::transform(*arguments, map);
```

And we need to also map in environment variables

```
environment->insert(key, map(environment->value(key)));
```

martyone: For the alternative RE I would consider combining that with

```
QRegularExpression re(QString("(?:%1|%2)([^:;,'\"]*)")
        .arg(QRegularExpression::escape(mapping.hostPath))
        .arg(QRegularExpression::escape(QDir::toNativeSeparators(mapping.hostPath))),
        reCaseSensitivity);
````

File: src/tools/sfdk/sdkmanager.cpp [hunk 4]
```diff
@@ -1348,15 +1348,32 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
...
+                              .arg(QRegularExpression::escape(QDir::toNativeSeparators(mapping.hostPath))),
+                              reCaseSensitivity);
+        auto map = [=](QString string) {
+            QRegularExpressionMatchIterator i = re.globalMatch(string);
+            while (i.hasNext()) {
+                QRegularExpressionMatch match = i.next();
+                QString fullNativeHostPath = match.captured(0);
+                QString fullEnginePath = QDir::fromNativeSeparators(mapping.enginePath + match.captured(2));
+                string.replace(fullNativeHostPath, fullEnginePath);
+            }
```
martyone, line 1369: It would still not work e.g. with `"c:\users\user\foo;c:\users\user\foo\bar\baz"` I think. We need to sort matches by length and replace longer matches first. Maybe build `QMap<QString, QString>` first and backward-iterate that doing the replacements.

The capture number should be 1 instead of 2, the first group is non capturing.

vigejolla: > The capture number should be 1 instead of 2, the first group is non capturing.

These write-only languages are hard :)

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1348,15 +1348,37 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
...
     for (const Mapping &mapping : mappings) {
         qCDebug(sfdk) << "Mapping" << mapping.hostPath << "as" << mapping.enginePath;
         program->replace(mapping.hostPath, mapping.enginePath, mapping.cs);
-        arguments->replaceInStrings(mapping.hostPath, mapping.enginePath, mapping.cs);
+        QRegularExpression::PatternOption reCaseSensitivity = (mapping.cs == Qt::CaseInsensitive)
+                ? QRegularExpression::CaseInsensitiveOption : QRegularExpression::NoPatternOption;
+        QRegularExpression re(QString("(?:%1|%2)([^:;,'\"]*)")
+                              .arg(QRegularExpression::escape(mapping.hostPath))
+                              .arg(QRegularExpression::escape(QDir::toNativeSeparators(mapping.hostPath))),
+                              reCaseSensitivity);
```
martyone, line 1361: I would move this into the lambda and move the lambda up just below the qCDebug to not disturb the higher level flow. I would consider adding some adjective to the name to express its special nature - maybe `mapGreedy`.

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1348,15 +1348,37 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
...
+        auto map = [=](QString string) {
+            QRegularExpressionMatchIterator reIterator = re.globalMatch(string);
+            QMap<QString,QString> replacements;
```
martyone, line 1364: space around operators.

I would swap those two lines and add comment before, e.g.,`Replace longer matches first! Use map for sorting.`

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1348,15 +1348,37 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
...
+            while (reIterator.hasNext()) {
+                QRegularExpressionMatch match = reIterator.next();
+                QString fullNativeHostPath = match.captured(0);
+                QString fullEnginePath = QDir::fromNativeSeparators(mapping.enginePath + match.captured(1));
+                replacements.insert(fullNativeHostPath, fullEnginePath);
+            }
+            for (auto i = replacements.end(); i != replacements.begin();) {
+                --i;
+                string.replace(i.key(), i.value());
+            }
```
martyone, line 1374: const_iterator (cend, cbegin) would be a little better. While at it, I would also consider using std::map instead of QMap and iterating with const_reverse_iterator, so it could be compactized as

```suggestion
            for (auto i = replacements.crbegin(); i != replacements.crend(); ++i)
                string.replace(i.key(), i.value());
```

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1348,15 +1348,37 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
         {cleanSharedSrc, Constants::BUILD_ENGINE_SHARED_SRC_MOUNT_POINT, caseInsensitiveOnWindows}
 #endif
     };
+    *program = QDir::fromNativeSeparators(*program);
+
```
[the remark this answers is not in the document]
vigejolla: done

src/tools/sfdk/sdkmanager.cpp
vigejolla: done

src/tools/sfdk/sdkmanager.cpp
vigejolla: done

src/tools/sfdk/sdkmanager.cpp
vigejolla: done

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1348,15 +1349,37 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
...
+            QRegularExpressionMatchIterator reIterator = re.globalMatch(string);
+            while (reIterator.hasNext()) {
+                QRegularExpressionMatch match = reIterator.next();
+                QString fullNativeHostPath = match.captured(0);
+                QString fullEnginePath = QDir::fromNativeSeparators(mapping.enginePath + match.captured(1));
+                replacements.insert(std::pair<QString, QString>(fullNativeHostPath, fullEnginePath));
+            }
+            for (auto i = replacements.crbegin(); i != replacements.crend(); ++i) {
+                string.replace(i->first, i->second);
+            }
```
martyone, line 1373: Extra braces

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1348,15 +1349,37 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
         {cleanSharedSrc, Constants::BUILD_ENGINE_SHARED_SRC_MOUNT_POINT, caseInsensitiveOnWindows}
 #endif
     };
+
     for (const Mapping &mapping : mappings) {
         qCDebug(sfdk) << "Mapping" << mapping.hostPath << "as" << mapping.enginePath;
+        auto mapGreedy = [=](QString string) {
+            QRegularExpression::PatternOption reCaseSensitivity = (mapping.cs == Qt::CaseInsensitive)
+                    ? QRegularExpression::CaseInsensitiveOption : QRegularExpression::NoPatternOption;
```
martyone, line 1357: Sorry I missed that before

```suggestion
            QRegularExpression::PatternOption reCaseSensitivity = mapping.cs == Qt::CaseInsensitive
                 ? QRegularExpression::CaseInsensitiveOption
                 : QRegularExpression::NoPatternOption;
```

src/tools/sfdk/sdkmanager.cpp
vigejolla: removed

src/tools/sfdk/sdkmanager.cpp
vigejolla: Moved : QRegularExpression:NoPatternOption to separate line. I refuse to remove the parenthesis around the comparison, as that would make the code more difficult to read.

martyone: Please check `grep -e '^[[:space:]]*?' -C1 -r src/plugins/ ` and the discussion we had before here https://github.com/sailfishos/sailfish-qtcreator/pull/272#pullrequestreview-374961612 (you need to click Show resolved)

vigejolla: Fine, just to be consistent with myself, I'll do the same as then: I'll change it, just to get the review accepted. I still think that it is a stupid thing to do. And I probably will fight this every single time.

vigejolla: > Fine, just to be consistent with myself, I'll do the same as then: I'll change it, just to get the review accepted. I still think that it is a stupid thing to do. And I probably will fight this every single time.

I hadn't noticed that you had already accepted the review ;) Well, it's changed anyway now..

PR #331, opened 2020-07-06
Jb50389 fix

vigejolla:
All characters were had "/" added to them when executing e.g. "sfdk engine exec"

### DIFF
--- a/src/tools/sfdk/sdkmanager.cpp
+++ b/src/tools/sfdk/sdkmanager.cpp
@@ -1316,6 +1316,7 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
         QProcessEnvironment *environment) const
 {
     QTC_ASSERT(hasEngine(), return false);
+    *program = QDir::fromNativeSeparators(*program);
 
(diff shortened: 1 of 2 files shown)


vigejolla:
Only last commit belongs to PR.
Yes, I will rebase before merging.

martyone:
It would be better to re-open the PR using sdk-3.2 as the target branch - it will be merged to master later so it is more convenient to fix it in the release branch.

vigejolla:
> Hm, unfortunately this doesn't eliminate the `cleanSharedTarget + "/"` case - doesn't it still double the slashes? I think we should store mappings in a std::vector a populate it dynamically.

No, it doesn't double the slashes. It just replaces empty + "/" with "/", i.e. does nothing. But of course it does extra work there. I agree, storing mappings a dynamic container would be better. I would have picked something from Qt, but std::vector works just as well.

> While here we can also drop the `Q_CC_GNU <= 504` case IIUC.

I also wondered why do we still have that - if there's no reason, I'll drop it.

I'll close this PR and open a new one, targeting sdk-3.2 branch.

PR #333, opened 2020-07-07
Sfdk: Don't map shared target path if it's empty

vigejolla:


src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1318,7 +1318,7 @@ QString SdkManager::cleanSharedTarget() const
     QString errorString;
     const BuildTargetData target = SdkManager::configuredTarget(&errorString);
     if (!target.isValid()) {
-        qerr() << errorString << endl;
+        qCDebug(sfdk) << errorString << endl;
```
martyone, line 1321: I would prefer to let `cleanSharedTarget` accept and forward `errorString` argument, then print the debug message only where it is used, i.e. in `mapEnginePath`.

Now when `qCDebug` is used, it would be better to add some contextual information too, so maybe

```
qCDebug(sfdk).noquote() << "Not mapping shared target path:" << errorString
```

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1353,21 +1353,18 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
         QString hostPath;
         QString enginePath;
         Qt::CaseSensitivity cs;
-    } const mappings[] = {
-#if Q_CC_GNU <= 504 // Let's check if it is still needed with GCC > 5.4
-        {cleanSharedTarget + QLatin1String("/"), QLatin1String("/"), Qt::CaseSensitive},
-        {cleanSharedTarget, QLatin1String("/"), Qt::CaseSensitive},
-        {cleanSharedHome, QLatin1String(Constants::BUILD_ENGINE_SHARED_HOME_MOUNT_POINT), Qt::CaseSensitive},
-        {cleanSharedSrc, QLatin1String(Constants::BUILD_ENGINE_SHARED_SRC_MOUNT_POINT), caseInsensitiveOnWindows}
-#else
```
martyone, line 1362: It would be nice to have the <=504 case removal in a separate commit :)

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1353,21 +1353,18 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
...
+    if (!cleanSharedTarget.isEmpty()) {
+        mappings.push_back({cleanSharedTarget + "/", "/", Qt::CaseSensitive});
+        mappings.push_back({cleanSharedTarget, "/", Qt::CaseSensitive});
+    }
+    mappings.push_back({cleanSharedHome, Constants::BUILD_ENGINE_SHARED_HOME_MOUNT_POINT, Qt::CaseSensitive});
+    mappings.push_back({cleanSharedSrc, Constants::BUILD_ENGINE_SHARED_SRC_MOUNT_POINT, caseInsensitiveOnWindows});
 
     for (const Mapping &mapping : mappings) {
+        if (mapping.hostPath.isEmpty())
+            continue;
```
martyone, line 1367: Not needed anymore, but maybe worth to keep in form of QTC_ASSERT.

src/tools/sfdk/sdkmanager.cpp
vigejolla: Fixed

src/tools/sfdk/sdkmanager.cpp
vigejolla: Moved to separate commit

src/tools/sfdk/sdkmanager.cpp
vigejolla: Replaced with QTC_ASSERT

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1312,13 +1312,12 @@ QString SdkManager::cleanSharedSrc() const
...
 
-QString SdkManager::cleanSharedTarget() const
+QString SdkManager::cleanSharedTarget(QString *errorMessage) const
 {
+    Q_ASSERT(errorMessage);
     QTC_ASSERT(hasEngine(), return {});
-    QString errorString;
-    const BuildTargetData target = SdkManager::configuredTarget(&errorString);
+    const BuildTargetData target = SdkManager::configuredTarget(errorMessage);
     if (!target.isValid()) {
```
martyone, line 1320: extra braces now

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1332,7 +1331,8 @@ bool SdkManager::mapEnginePaths(QString *program, QStringList *arguments, QStrin
 
     const QString cleanSharedHome = this->cleanSharedHome();
     const QString cleanSharedSrc = this->cleanSharedSrc();
-    const QString cleanSharedTarget = this->cleanSharedTarget();
+    QString errorString;
```
martyone, line 1334: maybe move that closer to the beginning so that it is available in case we need that for preceding calls in future

src/tools/sfdk/sdkmanager.cpp
vigejolla: removed

src/tools/sfdk/sdkmanager.cpp
vigejolla: I prefer declaring variables close to where they are used. But then again, having the cleanShared.*() calls grouped together is also preferable. I moved the declaration now so that it's just before cleanSharedHome() call - it's still relatively close but the calls are neatly together.

PR #336, opened 2020-07-10
Mer: Check that package list is not empty before validation

vigejolla:


src/plugins/mer/merdeploysteps.cpp
```diff
@@ -692,9 +692,10 @@ void MerMb2RpmBuildStep::processFinished(int exitCode, QProcess::ExitStatus stat
     }
 }
 
-QStringList MerMb2RpmBuildStep::packagesFilePath() const
+QString MerMb2RpmBuildStep::packagesFilePath() const
```
martyone, line 695: I would rename it e.g. as `mainPackageFileName`.

PR #338, opened 2020-07-14
Jb50115

martyone:


martyone:
> LGTM - even though I can't quite see how this affects only macOS

same here

PR #341, opened 2020-07-17
Jb50454

martyone:


src/tools/merssh/command.cpp
```diff
@@ -405,3 +402,20 @@ void Command::maybeUndoCMakePathMapping()
         }
     }
 }
+
+void Command::updateOrAddToCMakeCacheIf(QString *data, const QString &name, const QStringList &types,
```
vigejolla, line 409: Please consider adding updateOrAddToCMakeCacheIf in a separate commit

vigejolla: Also, I think the function name is a bit misleading now - it is not really CMakeCache specific

martyone: yup I struggled to do the refactorization with a separate commit but that wouldn't really make much sense and given the size of the change I chose to keep it in one commit. The function is definitely CMakeCache specific.

PR #342, opened 2020-07-29
Jb50618

martyone:


src/plugins/mer/mersdkmanager.cpp
```diff
@@ -444,7 +444,8 @@ bool MerSdkManager::addKit(const BuildEngine *buildEngine, const QString &buildT
     Kit* kitptr = kit(buildEngine, buildTargetName);
     std::unique_ptr<Kit> kit(nullptr);
     if (!kitptr) {
-        kit = std::make_unique<Kit>();
+        kit = std::make_unique<Kit>(Core::Id::fromSetting(QVariant(buildTargetName)));
```
martyone, line 447: It would be cleaner to encapsulate this in something like `Core::Id MerSdkManager::kitId(buildEngine, buildTargetName)` but I prefer minimizing the change here for easier merging as the code diverged a lot in master.

PR #349, opened 2020-08-18
Jb50857

martyone:


martyone:
> LGTM, except for a small typo in commit message. There is an extra '(' in the last line.

a weird typo :)

PR #353, opened 2020-09-22
Update and clean up Sailfish OS related documentation

martyone:


vigejolla:
There are still some mer related images left, I don't think these are actually used anywhere, so they could be simply removed:
creator-mer-options-targets.png
qtcreator-vm-mer.png
qtcreator-device-mer-port.png

doc/qtcreator/src/mer/creator-deploying-mer.qdoc
```diff
@@ -29,18 +29,54 @@
...
+    Three deployment methods are available.
 
-    The \gui {Deploy by Copying Binaries} step determines that \QC uses the
-    rsync file copying tool to transfer the files to the device or to the
-    emulator.
+    \list
+
+    \li Deploy As RPM Package
+
+        This is the default method. An RPM package will be build and
```
vigejolla, line 46:
```suggestion
        This is the default method. An RPM package will be built and
```

doc/qtcreator/src/mer/creator-deploying-mer.qdoc
```diff
@@ -29,18 +29,54 @@
...
+
+        It comes at a cost though. In order to avoid overwritting system files
+        accidentaly, the application will be deployed under the \c
+        /opt/sdk/<project-name> prefix when this method is used. This will
+        inevitably affect visibily of files that are expected to be installed
+        under fixed locations. The common approach to deal with this is to
+        deploy with RPM package initially and every time any of the affected
+        files change.
+
+        For QML based application development a more convenient option exists
```
vigejolla, line 62: Maybe "For developing the QML parts of the application..."

doc/qtcreator/src/mer/creator-developing-mer.qdoc
```diff
@@ -24,175 +24,83 @@
...
+    You can connect Sailfish OS devices to the development PC to run, debug,
+    and analyze applications on from \QC.
 
-    The Sailfish SDK installation program installs targets on the
-    development host and configures the appropriate \l{glossary-buildandrun-kit}
-    {kit} for you in \QC. After the installation, you are ready to start
-    developing applications, as described in
-    \l{Creating a Sailfish OS Application}.
+    On Sailfish SDK side, various combinations of particular Sailfish OS
+    versions ran on particular HW architectures are supported with so called
```
vigejolla, line 31: Maybe break this up a little bit to make it easier to read, e.g. "Various combinations of Sailfish OS versions and HW architectures are supported by so called SDK Build Targets. Each build target corresponds to a specific Sailfish OS version/HW architecture combination."

doc/qtcreator/src/mer/creator-developing-mer.qdoc
```diff
@@ -24,175 +24,83 @@
...
-            to show the VirtualBox user interface.
+    On the \gui Connection page, fill in the device hostname or IP address.
+    Device's IP address may be retrieved under the \gui Settings application,
+    under the \gui {Developer tools} category.
 
-        \li In the \gui Connection field, specify the SSH key to use to protect
-            the connection between \QC and the build engine:
+    The traditional device username was \c nemo. In recent Sailfish OS version
+    this changed to \c defaultuser. You may need to adjust the \gui Username
+    field to match your device username.
```
vigejolla, line 81: It might be worth mentioning that the username is visible in the same view as the IP address on the device

vigejolla (commented):
General feedback: Good stuff! I wrote some suggestions for minor changes though.

doc/qtcreator/src/mer/creator-deploying-mer.qdoc
martyone: Tried something else.

doc/qtcreator/src/mer/creator-deploying-mer.qdoc
martyone: Fixed

doc/qtcreator/src/mer/creator-developing-mer.qdoc
martyone: Shortened by removing some words.

Note the important detail here which was more visible in the original text: This is a Qt Creator manual, not Sailfish SDK manual. From Qt Creator point of view, Sailfish SDK is an external, optional part. Compare with Android SDK etc. I did some cleanup in this respect, dropping the instructions like adding build engine and emulators manually - not because these are managed by the SDK installer but because these should be discovered automatically by QtC in the hypothetical scenario when upstream QtC integrates our plugin.

martyone: And why I mention that: this is why the leading words "On Sailfish SDK side, " are important :)

doc/qtcreator/src/mer/creator-developing-mer.qdoc
martyone: Did some reorganization

martyone:
Dropped the unused image files. Thanks, I was already tired of finding my way through the maze.. where are the old good tree structured docs gone? I can't imagine using any of Qt docs without Google.

vigejolla (approved):
LGTM

PR #356, opened 2020-09-25
Jb51082

martyone:


src/tools/sfdk/command.cpp
```diff
@@ -1748,12 +1749,22 @@ Worker::ExitStatus BuiltinWorker::setProperties(SetPropertiesTask *task,
...
         QString errorString;
-        if (!task->prepareSet(property, value, &errorString)) {
+        if (!task->prepareSet(normalizedProperty, value, &errorString)) {
             *exitCode = EXIT_FAILURE;
             qerr() << property << ": " << errorString << endl;
             return NormalExit;
         }
+
+        if (normalizedProperty != property) {
+            qCInfo(sfdk).noquote() << tr("The \"%1\" property is deprecated. Use \"%2\" instead.")
```
vigejolla, line 1765: maybe worth of qCWarning?

vigejolla (commented):
I can see what this commit does, but question remains - why?

vigejolla:
> I can see what this commit does, but question remains - why?

Never mind. This comment was for the second to last commit - I thought I was already on the last commit. The last commit explains why :)

vigejolla (approved):
LGTM. The comment about change qCInfo to qCWarning is just a suggestion, do as you like.

src/tools/sfdk/command.cpp
martyone: This is actually less a question of severity and more of the type/source of the event reported https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/tools/sfdk/sfdkglobal.cpp#L69 . I chose to not categorize this event as "Bad usage error" as I feel that category more suitable for fatal errors while the `[I]` gives the clear feedback that the command still succeeded but changes will be needed in future (I am thinking about reading logs from programmatic use of sfdk).

PR #357, opened 2020-09-29
Jb50706

martyone:


martyone:
I am undecided. Should it be cleaned on input or the code model should be more robust in this respect? Let's see what upstream thinks https://codereview.qt-project.org/c/qt-creator/qt-creator/+/315748

vigejolla:
> I am undecided. Should it be cleaned on input or the code model should be more robust in this respect? Let's see what upstream thinks https://codereview.qt-project.org/c/qt-creator/qt-creator/+/315748

I think you made the right choice - let the upstream decide :)

martyone:
Updated according to upstream review. It is not likely more changes will be required, so changing destination branch to sdk-3.3 and merging now for integration with next RC.

PR #358, opened 2020-09-30
Jb51169

vigejolla:


src/plugins/mer/mersdkmanager.cpp
```diff
@@ -447,19 +447,22 @@ bool MerSdkManager::addKit(const BuildEngine *buildEngine, const QString &buildT
...
+        sdkProvided = true;
     }
 
     ToolChainManager::registerToolChain(toolchain.get());
     ToolChainManager::registerToolChain(toolchain_c.get());
     QtVersionManager::addVersion(version.get());
 
     auto initializeKit = [&](Kit *kit) {
-        finalizeKitCreation(buildEngine, buildTargetName, kit);
+        finalizeKitCreation(buildEngine, buildTargetName, kit, sdkProvided);
```
martyone, line 465: Maybe make 'kit' the very first argument when touching this.

martyone (approved):
More correctly "results in unnecessary assertion failures" :)

src/plugins/mer/mersdkmanager.cpp
vigejolla: > More correctly "results in unnecessary assertion failures" :)

yes of course :)

PR #360, opened 2020-10-07
Jb51492

vigejolla:
Instead of crashing, display error dialog when trying to add unsuitable VM as an emulator

### DIFF
--- a/src/libs/sfdk/buildengine.cpp
+++ b/src/libs/sfdk/buildengine.cpp
@@ -1,7 +1,7 @@
 /****************************************************************************
 **
 ** Copyright (C) 2012-2019 Jolla Ltd.
-** Copyright (C) 2019 Open Mobile Platform LLC.
+** Copyright (C) 2019-2020 Open Mobile Platform LLC.
 ** Contact: http://jolla.com/
 **
 ** This file is part of Qt Creator.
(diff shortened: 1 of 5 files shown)


src/plugins/mer/meremulatoroptionswidget.cpp
```diff
@@ -376,6 +376,11 @@ void MerEmulatorOptionsWidget::onAddButtonClicked()
     };
     Sdk::createEmulator(dialog.selectedVmUri(), &loop, whenDone);
     loop.exec();
+    if (!emulator) {
+        QMessageBox::warning(this, tr("Could not add emulator"),
+                             tr("The selected virtual machine is not suitable to be used as an emulator"));
```
martyone, line 381:
```suggestion
        QMessageBox::warning(this,
                tr("Could not add a %1 Emulator").arg(Sdk::osVariant()),
                tr("The selected virtual machine is not suitable to be used as a %1 Emulator")
                    .arg(Sdk::osVariant()));
```

I would also consider saying something like "could not be configured as" instead of "is not suitable to be used as" - as a less specific judgement (maybe the VM is an emulator VM and just some small bits are missing).

src/plugins/mer/meremulatoroptionswidget.cpp
```diff
@@ -376,6 +376,13 @@ void MerEmulatorOptionsWidget::onAddButtonClicked()
...
     Sdk::createEmulator(dialog.selectedVmUri(), &loop, whenDone);
     loop.exec();
+    if (!emulator) {
+        QMessageBox::warning(this,
+                tr("Could not add a %1 Emulator").arg(Sdk::osVariant()),
+                tr("The selected virtual machine could not be configured to be used as a %1 Emulator")
+                    .arg(Sdk::osVariant()));
+        return;
+    }
     QTC_ASSERT(emulator, return);
```
martyone, line 386: I forgot to mention this can be removed now.

src/libs/sfdk/buildengine.cpp
```diff
@@ -497,6 +497,14 @@ void BuildEnginePrivate::updateVmProperties(const QObject *context, const Functo
     });
 }
 
+bool BuildEnginePrivate::isValid() const
+{
+    QTC_ASSERT(!sharedConfigPath.isEmpty(), return false);
+    QTC_ASSERT(!sharedSshPath.isEmpty(), return false);
+    QTC_ASSERT(virtualMachine->sshParameters().port(), return false);
+    return true;
```
martyone, line 505: We should check all shared paths and other properties as well (if any).

src/libs/sfdk/emulator.cpp
```diff
@@ -442,6 +442,14 @@ void EmulatorPrivate::updateVmProperties(const QObject *context, const Functor<b
     });
 }
 
+bool EmulatorPrivate::isValid() const
+{
+    QTC_ASSERT(!sharedConfigPath.isEmpty(), return false);
+    QTC_ASSERT(!sharedSshPath.isEmpty(), return false);
+    QTC_ASSERT(virtualMachine->sshParameters().port(), return false);
+    return true;
```
martyone, line 450: Similarly as with engine, the other properties like mac, subnet, ... should be checked too I think.

src/plugins/mer/meremulatoroptionswidget.cpp
vigejolla: How on earth did I not notice that myself...

src/libs/sfdk/buildengine.cpp
vigejolla: I guess we should. My original approach was to just check those properties which are needed for establishing the ssh connection, which should be enough to prevent the crash. Saying that the build engine or emulator is valid is a slightly different thing.

martyone (approved):

PR #362, opened 2020-10-08
Jb51492-2

martyone:


martyone:
Merge to master after #361 is merged and mb2 is fixed

martyone:
Replaced with #364

PR #363, opened 2020-10-09
Jb51522

martyone:


share/qtcreator/sfdk/modules/50-testing-mb2.json
```diff
@@ -18,14 +18,14 @@
             "name": "deploy",
             "trSynopsis": "{--pkcon|--rsync|--sdk|--zypper|--manual} [--all] [--debug] [--] [[+|-]<pattern>...]",
             "trBrief": "Deploy to a device",
-            "configOptions": ["device!", "output-dir", "output-prefix", "target!", "[no-]task"],
+            "configOptions": ["*", "device!", "-[no-]fix-version", "-[no-]search-output-dir", "target!"],
```
vigejolla, line 21: There's a lot more happening here than just adding the [no-]snapshot option. What does the asterisk do here? Why are you adding -[no-]fix-version and -[no-]search-output-dir? And removing output-dir, output-prefix and [no-]task?

martyone: "*" - use all module options, "-" prefix - exclude, "!" suffix - mandatory

So this is just switching from the "start with none, then add" style to "start with all, then exclude" style.

vigejolla: Of course...

vigejolla (approved):
LGTM

PR #369, opened 2020-10-16
gdb: Package libncursesw-6.dll and libgcc_s_dw2-1.dll from MinGW

vigejolla:


dist/gdb/Makefile.mingw
```diff
@@ -15,6 +15,7 @@ gdbtargets=$(addprefix ${targetdir}/bin/gdb-, ${targets})
 packageparts=${targetdir}/bin/lib ${targetdir}/bin/lib ${targetdir}/bin/libiconv-2.dll ${targetdir}/bin/python27.dll ${targetdir}/bin/libexpat-1.dll
 arch=`uname -sm | sed 's/ /-/g' | tr A-Z a-z`
 packagename=sailfish-gdb-${version}-windows.7z
+mingwbin=/c/MinGW/bin
```
martyone, line 18: No way to compute this somehow?

martyone (approved):

dist/gdb/Makefile.mingw
vigejolla: I was thinking this myself as well, but could figure out an easy way. Of course we could hack something together, like check all directories in PATH if they happen to contain the files we want, but I don't really think that's a good idea. I can still try to think of something while waiting for Jenkins to build...

martyone: maybe... isn't that the dir where gcc executable exists? `mingwbin=$(dirname "$(which gcc-exact-name)")`?

vigejolla: Yes. I'll try that after I have confirmed that this really fixes the issue.

PR #370, opened 2020-10-16
Jb51617

martyone:


vigejolla (approved):
LGTM. I think you shouldn't worry too much about undetected changes - if those are to happen, we should catch them after the build has finished, and we swap the files again. At least I think so...

martyone:
> [...] we should catch them after the build has finished, and we swap the files again. At least I think so...

That's what I am a bit worried about. I think this may or may not fail depending on the timing.

The desired order:

sfdk: swap files
sfdk: exit
qtc: handle build finished
qtc: handle changed files

May change to:

sfdk: swap files
qtc: handle changed files
sfdk: exit
qtc: handle build finished

quite easily

PR #373, opened 2020-10-26
Jb51595

martyone:


vigejolla (approved):
LGTM, however I would feel more confident if we merged this only after the 3.3 release

martyone:
> LGTM, however I would feel more confident if we merged this only after the 3.3 release

This goes to a different branch :)

vigejolla:
> This goes to a different branch :)

Even better :)

PR #375, opened 2020-11-20
Set OPENSSL_armcap in default debug environment

vigejolla:


martyone:
Hm, I need to ask (myself) why upstream does not allow to configure debug environment separately? Or do they? If not, then this would be likely the only use case ever seen for it. Having a check box under "Debugger settings" would be a better UI. But do we need a UI for this at all? In 99.9% cases there would be no need to clear that env and if so, it could be done by adding "unset environment ..." to "Additional startup commands". I would rather go without UI - the less code written the better :)

And we also need to set it for `sfdk debug`. Wouldn't it be possible to set it somehow globally for gdb instead?

vigejolla:
> Hm, I need to ask (myself) why upstream does not allow to configure debug environment separately? Or do they? If not, then this would be likely the only use case ever seen for it. Having a check box under "Debugger settings" would be a better UI. But do we need a UI for this at all? In 99.9% cases there would be no need to clear that env and if so, it could be done by adding "unset environment ..." to "Additional startup commands". I would rather go without UI - the less code written the better :)

Last week I tried setting the environment variable with "set environment" in "Additional startup commands" - it didn't work. I don't know why. I could try if unsetting works though.

> And we also need to set it for `sfdk debug`. Wouldn't it be possible to set it somehow globally for gdb instead?

Right, I forgot about sfdk. I'll fix that one as well. I'm quite sure I tried already adding the "set environment" command in .gdbinit, but as I'm not 100% sure I could try again.

vigejolla:
> Hm, I need to ask (myself) why upstream does not allow to configure debug environment separately? Or do they? If not, then this would be likely the only use case ever seen for it. Having a check box under "Debugger settings" would be a better UI. But do we need a UI for this at all? In 99.9% cases there would be no need to clear that env and if so, it could be done by adding "unset environment ..." to "Additional startup commands". I would rather go without UI - the less code written the better :)

Tested and it was as I expected - adding "unset environment" command to "Additional startup command" does not work. So I don't really see any other option than adding an UI for not setting the environment variable.

Btw. Upstream does not allow setting environment variables for debugging. They do however allow setting them for running apps (when started from QtC). In theory we could just modify that environment, it would probably not do much harm.

> And we also need to set it for `sfdk debug`. Wouldn't it be possible to set it somehow globally for gdb instead?

After a little bit more investigation, I feel there is no way to "set it globally for gdb". Qt Creator does quite a good job at removing all environment variables which it doesn't know about when it starts the debugger. So even if we managed to set it up in gdb, we would still need to add code to QtC to not remove it, and then we would need to add UI to remove it. So it would be even more complicated than it already is. 

In the end, I find the current implementation the most convenient. Unless you can come up with an alternative solution which actually works, I'll just add similar functionality to sfdk.

src/plugins/mer/merrunconfigurationaspect.cpp
```diff
@@ -129,7 +134,13 @@ class MerRunConfigWidget : public QWidget
                 MerQmlLiveBenchManager::offerToStartBenchIfNotRunning();
         });
 
-        layout()->addWidget(qmlLiveWidget);
+        qmlLiveLayout->addWidget(qmlLiveWidget);
+        outerLayout->addLayout(qmlLiveLayout);
+
+        QCheckBox* debuggerEnvironmentCheckbox = new QCheckBox(tr("Bypass OPENSSL ARM capability detection while debugging"));
```
martyone, line 140:
```suggestion
        QCheckBox* debuggerEnvironmentCheckbox = new QCheckBox(tr("Bypass OpenSSL ARM capability detection while debugging"));
        debuggerEnvironmentCheckbox->setToolTip(tr("When checked, \"OPENSSL_armcap=1\" (ARMV7_NEON) "
                "will be set in the process environment under debugger. "
                "The way OpenSSL detects CPU capabilities would lead to receiving SIGILL at startup."));
```

src/plugins/mer/merdevicedebugsupport.cpp
```diff
@@ -113,6 +113,12 @@ MerDeviceDebugSupport::MerDeviceDebugSupport(RunControl *runControl)
 
     setUsePortsGatherer(isCppDebugging(), isQmlDebugging());
 
+    auto aspect = runControl->runConfiguration()->aspect<MerRunConfigurationAspect>();
+    if (aspect->isDebugEnvironmentEnabled()) {
```
martyone, line 117: I would prefer using more specific naming for this and all the related identifiers, e.g. "isDebugBypassOpenSslArmCapEnabled".

src/plugins/mer/merrunconfigurationaspect.cpp
```diff
@@ -129,7 +134,13 @@ class MerRunConfigWidget : public QWidget
...
+        debuggerEnvironmentCheckbox->setChecked(m_aspect->isDebugEnvironmentEnabled());
+        connect(debuggerEnvironmentCheckbox, &QCheckBox::stateChanged, m_aspect, &MerRunConfigurationAspect::setDebugEnvironmentEnabled);
```
martyone, line 142: long line

src/tools/sfdk/sfdkconstants.h
```diff
@@ -41,6 +41,7 @@ const int EXIT_ABNORMAL_DEFAULT_CODE = 120;
 const char NO_QPA_PLATFORM_MINIMAL_ENV_VAR[] = "SFDK_NO_QPA_PLATFORM_MINIMAL";
 const char CONNECTED_TO_TERMINAL_HINT_ENV_VAR[] = "SFDK_CONNECTED_TO_TERMINAL";
 const char NO_TEMP_REPOSITORY_ENV_VAR[] = "SFDK_NO_TEMP_REPOSITORY";
+const char NO_DEBUG_OPENSSL_ARMCAP_ENV_VAR[] = "SFDK_NO_DEBUG_OPENSSL_ARMCAP";
```
martyone, line 44: I would prefer SFDK_NO_BYPASS_OPENSSL_ARMCAP

src/plugins/mer/merdevicedebugsupport.cpp
vigejolla: I agree - for some reason it didn't occur to me to change the names when removing the generic debug environment functionality

martyone: hm, I still see some occurrences of the original name (setter, member variable, constant IIRC)

vigejolla: :( sorry about that, I'll fix them asap

src/plugins/mer/merrunconfigurationaspect.cpp
```diff
@@ -129,7 +134,18 @@ class MerRunConfigWidget : public QWidget
                 MerQmlLiveBenchManager::offerToStartBenchIfNotRunning();
         });
 
-        layout()->addWidget(qmlLiveWidget);
+        qmlLiveLayout->addWidget(qmlLiveWidget);
+        outerLayout->addLayout(qmlLiveLayout);
```
martyone, line 138: Is the extra layout really (still?) needed?

vigejolla: Yes

martyone: Please elaborate. After dropping it I didn't see a difference in the UI.

martyone (approved):
Yup and can we also drop the `layout` variable and use `layout()` as before? And consider renaming also the check box.

PR #379, opened 2020-12-03
Mer: Check updates daily if EA repository is enabled

vigejolla:


src/plugins/mer/merplugin.cpp
```diff
@@ -409,5 +412,60 @@ void MerPlugin::ensureCustomRunConfigurationIsTheDefaultOnCompilationDatabasePro
...
+        .toList();
+    foreach (const QVariant &variant, variants) {
+        QInstaller::Repository repo = variant.value<QInstaller::Repository>();
+        if (repo.m_enabled && repo.m_url.toString().endsWith("updates-ea")) {
+            QSettings *settings = ICore::settings();
+            settings->beginGroup(QLatin1String("Updater"));
+            settings->setValue(QLatin1String("CheckUpdateInterval"), QLatin1String("DailyCheck"));
+            settings->endGroup();
+        }
+    }
```
martyone, line 436: I would add a safety QTC_CHECK that (e.g.) the "updates-ea" repo is in the list, i.e., that we successfully parsed the .ini, which may change in future.

I would consider reverting to weekly check when EA updates are disabled later.

src/plugins/mer/merplugin.h
```diff
@@ -83,4 +85,27 @@ private slots:
...
+    bool m_enabled;
+};
+
+QDataStream &operator>>(QDataStream &istream, QInstaller::Repository &repository);
+QDataStream &operator<<(QDataStream &ostream, const QInstaller::Repository &repository);
+
+}
+
+Q_DECLARE_METATYPE(QInstaller::Repository)
+
```
martyone, line 110: I would prefer moving this to the .cpp file.

src/plugins/mer/merplugin.cpp
vigejolla: QTC_CHECK sounds like a good idea.

I thought about reverting back to weekly, but then decided against it: We would have to store the information that the change to daily was done by us, and not by the user. We don't want to change the setting if the user has modified it, unless the user has enabled the EA repo, in which case we want to force it to daily. I don't find the extra effort in storing the information about us modifying the setting worth it.

martyone: Yeah, I thought it was like that :)

martyone (approved):
Consider also inlining Repository implementation now that it is all in cpp.

PR #380, opened 2020-12-09
Jb49109

martyone:


src/tools/sfdk/dbus.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+namespace Sfdk {
+
+namespace {
+const char DBUS_SERVICE_NAME_TEMPLATE[] = "org.sailfishos.sfdk.I%1";
+const char DBUS_SFDK_ERROR[] = "org.sailfishos.sfdk.Error";
+const char DBUS_CONFIGURED_DEVICE_PATH[] = "/device";
+} // namespace anonymous
+
+class AbstractAdaptor : public QDBusAbstractAdaptor
```
vigejolla, line 45: I somewhat dislike the name of the class AbstractAdaptor. It's derived from QDBusAbstractAdaptor, and it's less abstract than QDBusAbstractAdaptor. But the name implies the opposite! Well, I don't have any better naming suggestion, so feel free to ignore this comment ;)

vigejolla (approved):
Tentative approvement: I reserve the right to change revoke it after reading the other PRs... :)

PR #383, opened 2020-12-25
Fix display of the validator suites

etaishev:
If a long address is specified in the suites description the profile name may completely overlap


martyone:
Wasn't there a related bz item? It would be nice to have the bug number mentioned in branch name.

etaishev:
I didn't create a bug for so small a problem. Should I create a bug?

martyone:
Yes, please, required for administrative purposes.

etaishev:
https://github.com/sailfishos/sailfish-qtcreator/pull/385

PR #385, opened 2021-01-14
JB52737

etaishev:


src/plugins/mer/merdeploysteps.cpp
```diff
@@ -153,7 +153,7 @@ class MerRpmValidationStepConfigWidget : public ProjectExplorer::BuildStepConfig
             item->setCheckState(0, step->selectedSuites().contains(suite.id) ? Qt::Checked : Qt::Unchecked);
             item->setData(0, Qt::UserRole, suite.id);
         }
-        m_ui.suitesTreeWidget->header()->setSectionResizeMode(0, QHeaderView::Stretch);
+        m_ui.suitesTreeWidget->header()->setSectionResizeMode(0, QHeaderView::ResizeToContents);
         m_ui.suitesTreeWidget->header()->setSectionResizeMode(1, QHeaderView::ResizeToContents);
         m_ui.suitesTreeWidget->header()->setSectionResizeMode(2, QHeaderView::ResizeToContents);
```
martyone, line 158: I think we should set index 1 to Stretch now. We want the table to use all available width and at the same to avoid horizontal scrollbars to appear.

Could you please attach two screenshots - one with short URL, one with URL which overflows - after changing this?

etaishev: Screenshots with stretch on website column:
`m_ui.suitesTreeWidget->header()->setSectionResizeMode(1, QHeaderView::Stretch);`


strech_with_long_url
strech_with_short_url

etaishev: So, if we want to avoid horizontal scrollbars we need more changes.

martyone: And how does it behave *without* changing resize mode for column 1?

```
         m_ui.suitesTreeWidget->header()->setSectionResizeMode(0, QHeaderView::ResizeToContents); 
         m_ui.suitesTreeWidget->header()->setSectionResizeMode(2, QHeaderView::ResizeToContents); 
```

etaishev: In this case the address overlaps the last column
no_resize

martyone: pity. If you do not manage to fix it, we could work it around by swapping the second and third column, .i.e., making url the last column. But I would like to avoid that.

src/plugins/mer/merdeploysteps.cpp
```diff
@@ -123,6 +123,26 @@ class MerPrepareTargetStepConfigWidget : public BuildStepConfigWidget
...
+        QLabel("", parent, f)
+    {
+        m_text = text;
+    }
+    void resizeEvent(QResizeEvent *event) {
+        QWidget::resizeEvent(event);
+
+        QFontMetrics metrics(font());
+        QString elidedText = metrics.elidedText(m_text, Qt::ElideRight, width());
+        setText(QString("<a href='%1'>%2</a>").arg(m_text, elidedText));
```
martyone, line 139: Tricky. If the font used for links differs...

src/plugins/mer/merdeploysteps.cpp
```diff
@@ -144,7 +164,7 @@ class MerRpmValidationStepConfigWidget : public ProjectExplorer::BuildStepConfig
             hasSelectedSuite |= suite.essential;
             auto *item = new QTreeWidgetItem(m_ui.suitesTreeWidget);
             item->setText(0, suite.name);
-            QLabel *websiteLabel = new QLabel(QString("<a href='%1'>%1</a>").arg(suite.website));
+            ElidedWebsiteLabel *websiteLabel = new ElidedWebsiteLabel(suite.website);
```
martyone, line 167: `auto *websiteLabel = ...`

src/plugins/mer/merdeploysteps.cpp
```diff
@@ -123,6 +123,26 @@ class MerPrepareTargetStepConfigWidget : public BuildStepConfigWidget
...
 };
 
+class ElidedWebsiteLabel : public QLabel
+{
+public:
+    explicit ElidedWebsiteLabel(const QString &text, QWidget *parent=nullptr, Qt::WindowFlags f=Qt::WindowFlags()) :
+        QLabel("", parent, f)
+    {
+        m_text = text;
+    }
```
martyone, line 133: if we never pass flags... style

```suggestion
    explicit ElidedWebsiteLabel(const QString &text, QWidget *parent = nullptr)
        : QLabel(QString(), parent)
        , m_text(text)
    {
    }

```

src/plugins/mer/merdeploysteps.cpp
```diff
@@ -123,6 +123,26 @@ class MerPrepareTargetStepConfigWidget : public BuildStepConfigWidget
...
+    void resizeEvent(QResizeEvent *event) {
+        QWidget::resizeEvent(event);
+
+        QFontMetrics metrics(font());
+        QString elidedText = metrics.elidedText(m_text, Qt::ElideRight, width());
+        setText(QString("<a href='%1'>%2</a>").arg(m_text, elidedText));
+        setOpenExternalLinks(true);
```
martyone, line 140: We already set that externally and unless you see a reason to do it internally I would prefer keeping it outside.

src/plugins/mer/merdeploysteps.cpp
```diff
@@ -123,6 +123,26 @@ class MerPrepareTargetStepConfigWidget : public BuildStepConfigWidget
...
+    }
+private:
+    QString m_text;
```
martyone, line 143: const

src/plugins/mer/merdeploysteps.cpp
```diff
@@ -123,6 +123,26 @@ class MerPrepareTargetStepConfigWidget : public BuildStepConfigWidget
...
 
+class ElidedWebsiteLabel : public QLabel
+{
+public:
+    explicit ElidedWebsiteLabel(const QString &text, QWidget *parent=nullptr, Qt::WindowFlags f=Qt::WindowFlags()) :
+        QLabel("", parent, f)
+    {
+        m_text = text;
+    }
+    void resizeEvent(QResizeEvent *event) {
```
martyone, line 134: brace after line break

martyone (commented):
nice idea

src/plugins/mer/merdeploysteps.cpp
etaishev: and so we put the setOpenExternalLinks outside

src/plugins/mer/merdeploysteps.cpp
martyone: This is still here

src/plugins/mer/merdeploysteps.cpp
```diff
@@ -144,7 +163,7 @@ class MerRpmValidationStepConfigWidget : public ProjectExplorer::BuildStepConfig
             hasSelectedSuite |= suite.essential;
             auto *item = new QTreeWidgetItem(m_ui.suitesTreeWidget);
             item->setText(0, suite.name);
-            QLabel *websiteLabel = new QLabel(QString("<a href='%1'>%1</a>").arg(suite.website));
+            auto websiteLabel = new ElidedWebsiteLabel(suite.website);
```
martyone, line 166: Tooltip with full url visible would be nice.

src/plugins/mer/merdeploysteps.cpp
martyone: Let's assume this won't happen

martyone (approved):
Rebase please

PR #387, opened 2021-01-22
Jb52481

martyone:


vigejolla (commented):
typo in commit message: SshProcessRunnner -> SshProcessRunner

vigejolla (approved):
LGTM. An interesting fix to an interesting problem. But shouldn't we upstream the fix?

martyone:
Yup, I am going to do so.

PR #392, opened 2021-02-19
Jb42918

martyone:


src/libs/sfdk/emulator.cpp
```diff
@@ -204,14 +204,20 @@ void Emulator::restoreFactoryState(const QObject *context, const Functor<bool> &
...
-            QTC_CHECK(updatedOk);
-            functor(restoredOk && updatedOk);
-        });
+        *allOk &= restoredOk;
+    });
+
+    d_func()->updateVmProperties(this, [=](bool updatedOk) {
+        QTC_CHECK(updatedOk);
+        *allOk &= updatedOk;
+        callIf(context_, functor, *allOk);
```
vigejolla: How do you make sure restoreSnapshot() has finished and it can be trusted that allOk already contains also restoredOk before this?

src/libs/sfdk/asynchronous.cpp
```diff
@@ -121,6 +137,7 @@ void CommandQueue::cancelBatch(BatchId batchId)
 void CommandQueue::enqueue(std::unique_ptr<CommandRunner> &&runner)
 {
     QTC_ASSERT(runner, return);
+    QTC_CHECK(!CommandRunner::isPostprocessing()); // use enqueueImmediate for postprocessing!
```
vigejolla, line 140: I like the idea! :)

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -110,9 +110,10 @@ CommandQueue *commandQueue()
  * \class DockerVirtualMachine
  */
 
-DockerVirtualMachine::DockerVirtualMachine(const QString &name, QObject *parent)
+DockerVirtualMachine::DockerVirtualMachine(const QString &name, VirtualMachine::Features featureMask,
+        QObject *parent)
     : VirtualMachine(std::make_unique<DockerVirtualMachinePrivate>(this), staticType(),
-            staticFeatures(), name, parent)
+            staticFeatures() & featureMask, name, parent)
```
vigejolla, line 116: Feels a bit silly to me, as staticFeatures() already returns NoFeatures.

src/libs/sfdk/buildengine.cpp
```diff
@@ -439,7 +439,9 @@ bool BuildEnginePrivate::initVirtualMachine(const QUrl &vmUri)
     Q_Q(BuildEngine);
     Q_ASSERT(!virtualMachine);
 
-    virtualMachine = VirtualMachineFactory::create(vmUri);
+    VirtualMachine::Features unsupportedFeatures = VirtualMachine::Snapshots;
+
+    virtualMachine = VirtualMachineFactory::create(vmUri, ~unsupportedFeatures);
```
vigejolla, line 444: For some reason the old logic was easier to understand (for me). But I *think* this works.

src/libs/sfdk/virtualmachine_p.h
```diff
@@ -189,6 +186,9 @@ class VirtualMachinePrivate
 private:
     void enableUpdates();
 
+signals:
+    void aboutToRestoreSnapshot(const std::shared_ptr<bool> &ok);
```
vigejolla, line 198: Somehow it feels a bit wrong to have a signal which, based on the given name, seems to just indicate that we are about to restore a snapshot, to have a return value as a parameter. I mean, it's not like we are going to not restore the snapshot because the slot failed to do whatever it wants to do? I do understand that it's there because we want to forward the error reporting, but still it feels a bit weird.

martyone: Yes, the intention here is not just signaling but calling back and getting the return value(s) collected from all observers. Moreover an asynchronous way - hence the `shared_ptr`. (It would be plain `bool *ok` in synchronous code). Compare this e.g. with Boost's combiners https://www.boost.org/doc/libs/1_61_0/doc/html/signals2/tutorial.html#idp394578384 . Qt is not that powerful here, so the combiner work is offloaded to observers.

src/libs/sfdk/asynchronous.cpp
martyone: Yeah, hopefully it is sufficient substitution for the missing documentation :) thanks! :)

src/libs/sfdk/virtualmachine_p.h
vigejolla: Interesting read, thanks for the link. The signals in boost are indeed a bit different than the signals in Qt. You are now twisting the Qt signals to act more like boost signals. I'm not saying it's wrong, it just feels a bit weird. And now I'm also a bit sad the Qt didn't take the opportunity to add signal return values (and combiners!) in Qt 6. Because those would be useful, like we see here!

src/tools/sfdk/dbus.cpp
```diff
@@ -72,24 +82,36 @@ class ConfiguredDeviceAdaptor : public AbstractAdaptor
...
+                    || test.exec() != EXIT_SUCCESS) {
+                replyError(message, tr("Failed to prepare device"));
+            }
+        });
+    }
+
+private:
+    template<typename Fn>
+    auto withDevice(const QDBusMessage &message, Fn fn) const
+        -> decltype(fn(std::declval<Device *>()))
```
vigejolla, line 113: I do have to wonder if there's an easier way to do this. Maybe there isn't, but this just feels over complicated.

src/tools/sfdk/dbus.cpp
```diff
@@ -82,6 +87,13 @@ class ConfiguredDeviceAdaptor : public AbstractAdaptor
     }
 
 public slots:
+    bool isEmulator(const QDBusMessage &message)
```
vigejolla, line 90: I don't think this is actually used anywhere

src/tools/sfdk/dbus.cpp
martyone: What do you mean more exactly?

src/tools/sfdk/dbus.cpp
vigejolla: never mind, of course it's called from dbus.

martyone: Is it? I didn't find it quickly so I thought I actually forgot to utilize it on sdk-setup side - now it would be nice to know if you found it somewhere :)

src/tools/sfdk/dbus.cpp
vigejolla: Now I know what I mean, it took me a while to even realize it myself, as I had to start writing this comment in order to get my thoughts in order! :-D
The way I see this, the return value of withDevice() is only used by isEmulator(), which returns a bool. Using decltype and std::declval to deduce the return type feels a bit excessive when the the only used return values are bool or void. We could use e.g. std::optional for this. But then again, this is of course more future proof. So yeah, I don't have any real suggestions for improvements here, it's just trying to figure out what's happening here is not that easy. It's not really your fault but rather the C++ syntax which is not always as readable as I wish it was.

vigejolla (approved):
Huge PR, so it's quite hard to be 100% sure about everything, especially when there are the cross-dependencies to other PRs in other components. Anyway, good addition(s) to the functionalities of the SDK.

martyone:
Rebased

src/tools/sfdk/dbus.cpp
vigejolla: No I actually didn't.

martyone: Yeah and I forgot to fix it before merging. Anyway, noted down and I will fix it while working on the autotests.

PR #394, opened 2021-03-04
Jb50188

etaishev:


martyone:
This PR was split from PR #325

martyone (commented):
Now I realized if we are adding UI for deploying RPM keys, then we should provide also UI for listing and removing existing keys. But those should be primarily provided on the device - it should be possible to check and remove user-deployed keys under settings.

etaishev:
> Now I realized if we are adding UI for deploying RPM keys, then we should provide also UI for listing and removing existing keys. But those should be primarily provided on the device - it should be possible to check and remove user-deployed keys under settings.

I thought there would be an item in the settings menu on the device for key management

martyone:
Yup, that's what I meant. Or should I understand it so that you thought there already was one? At least I didn't find it on my 4.0.1 device.

etaishev:
No, it's not ready yet

PR #395, opened 2021-04-08
Jb50531

martyone:


src/plugins/mer/merdeploysteps.cpp
```diff
@@ -259,22 +259,27 @@ bool MerProcessStep::init(InitOptions options)
     const FilePath projectDirectory = isShadowBuild
         ? bc->rawBuildDirectory()
         : project()->projectDirectory();
-    const FilePath toolsPath = engine->buildTarget(target).toolsPath;
-    CommandLine deployCommand(toolsPath.pathAppended(Sfdk::Constants::WRAPPER_DEPLOY));
-    deployCommand.addArgs(arguments(), CommandLine::Raw);
-
-    ProcessParameters *pp = processParameters();
+    // FIXME needed?
```
vigejolla: So.. is it needed? Just pointing this out in case this is something you intended to look at before merging.

martyone: That's indeed an open issue that I didn't found crucial to deal with immediately, so I left the comment just as a reminder for future. Thanks.

vigejolla (approved):
After spending considerable amount of time browsing through this PR now, I have the feeling that we should just merge it and try to find possible bugs via testing. It's just way too big to be able to  say that "everything in this is great" with confidence.

martyone:
Yup, there definitely are some.

PR #402, opened 2021-04-13
Jb53872

martyone:


vigejolla (approved):
I approve, but I refuse to say that this looks good. Honestly, I think this should be fixed on the Qt side. I don't think this is "unsolvable" like they say. Sure, they can never be 100% that the window is still they send the message, but they could detect that we are shutting down and ignore the return value in that case. Well, since they have determined that it's unsolvable, I guess it is not worth it to make a PR there, so this ugly hack is the best we can do.

PR #407, opened 2021-04-21
Jb49857

martyone:


vigejolla:
Generally speaking, I like the approach. I didn't fully review each and every change, but rather picked some code paths at random, and they seem to be working like they should. I'm not going to approve this yet as I assume you will do more debugging which might cause changes - let me know when you think this is ready to be merging and I'll have another look.

martyone:
@vigejolla Debugged, ready for review.

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -206,77 +199,52 @@ void DockerVirtualMachinePrivate::start(const QObject *context, const Functor<bo
     Q_ASSERT(context);
     Q_ASSERT(functor);
 
-    const QPointer<const QObject> context_{context};
+    BatchComposer composer = BatchComposer::createBatch("DockerVirtualMachinePrivate::start");
+    composer.batch()->setPropagateFailure(true);
```
vigejolla, line 203: It seems to me that createBatch() is really often followed by composer.batch()->setPropagateFailure(true). Perhaps you could consider having an alternative version of createBatch, which takes propagateFailure as parameter?

src/libs/sfdk/sdk.cpp
```diff
@@ -118,7 +118,7 @@ Sdk::Sdk(Options options)
 Sdk::~Sdk()
 {
     Q_D(Sdk);
-    d->commandQueue_->wait();
+    QTC_CHECK(d->commandQueue_->isEmpty()); // call shutDown()!
```
vigejolla, line 121: should we call shutDown()?

src/libs/sfdk/virtualmachine.cpp
```diff
@@ -571,19 +574,20 @@ void VirtualMachine::refreshConfiguration(const QObject *context, const Functor<
     });
 }
 
-void VirtualMachine::refreshState(Sfdk::VirtualMachine::Synchronization synchronization)
+void VirtualMachine::refreshState(const QObject *context, const Functor<bool> &functor)
```
vigejolla, line 602: Something to think about (definitely not fix now!): We seem to have this pattern of passing these (context, functor) pairs around a lot. Wouldn't it be convenient if we had a type which contained both?

src/libs/sfdk/virtualmachine.cpp
```diff
@@ -746,10 +750,8 @@ void VirtualMachinePrivate::enableUpdates()
         // BuildEngine or Emulator classes.
         qCDebug(vms) << "VM info cache updated. Refreshing configuration of" << q->uri().toString();
 
-        // FIXME Not ideal
-        bool ok;
-        execAsynchronous(std::tie(ok), std::mem_fn(&VirtualMachine::refreshConfiguration), q);
-        QTC_CHECK(ok);
+        // TODO Is that ideal?
```
vigejolla, line 778: Close enough if you ask me :)

src/plugins/mer/mercmakebuildconfiguration.cpp
```diff
@@ -105,10 +106,16 @@ MerCMakeBuildConfigurationFactory::MerCMakeBuildConfigurationFactory()
...
-    MerSdkKitAspect::buildEngine(target()->kit())->virtualMachine()->connectTo(
-        Sfdk::VirtualMachine::AskStartVm|Sfdk::VirtualMachine::Block);
+    BuildEngine *const engine = MerSdkKitAspect::buildEngine(target()->kit());
+    if (engine->virtualMachine()->state() == VirtualMachine::Connected)
+        return;
+
+    bool ok;
+    execAsynchronous(std::tie(ok), std::mem_fn(&VirtualMachine::connectTo),
+            engine->virtualMachine(), VirtualMachine::AskStartVm);
+    // FIXME handle failure?
```
vigejolla, line 118: An idea: override isEnabled() and disabledReason(), with something like "Could not start build engine" as the reason for being disabled?

src/libs/sfdk/buildengine.cpp
```diff
@@ -319,8 +319,10 @@ void BuildEngine::importPrivateGpgKey(const QString &id,
         const Functor<bool, QString> &functor)
 {
     QString errorString;
-    QTC_ASSERT(isGpgAvailable(&errorString),
-            QTimer::singleShot(0, context, std::bind(functor, false, errorString)); return);
+    QTC_ASSERT(isGpgAvailable(&errorString), {
+        BatchComposer::enqueueCheckPoint(context, std::bind(functor, false, errorString));
```
vigejolla, line 323: "Fix async returns from asserts" - not really the best possible commit message. Why where they broken? I would appreciate "Sfdk: use enqueueCheckPoint() instead of QTimer::singleShot" and then on the next line "QTimer::singleShot no longer works because of x" more.

vigejolla (approved):
The comments I wrote were mostly just food for thought. Nothing clearly bad (except maybe that one comment about commit message, that one I would really like to see fixed). I think this is good enough to be merged as is.

src/libs/sfdk/virtualmachine.cpp
martyone: I also thought about, but on caller side it would be less convenient - it would require an extra pair of braces - while the profit on implementation side would be just subtle I think.

src/libs/sfdk/virtualmachine.cpp
martyone: :)

src/plugins/mer/mercmakebuildconfiguration.cpp
martyone: Noted down

src/libs/sfdk/buildengine.cpp
martyone: They never worked. I realized it some time ago already but only now there was an opportunity to fix it. They fulfilled the requirement of being asynchronous, but failed at fulfilling the requirement of being ordered.

I will add that info to the commit message but I will leave the summary as is to describe the "what" rather than the "how" and fit into 50 chars limit more easily :)

src/libs/sfdk/sdk.cpp
martyone: Fixed

PR #412, opened 2021-04-27
Workarounded podman support

SfietKonstantin:
This is a simple workaround to support podman.
For local images, podman always prepend "localhost".
This messes up with the sfdk engine detection that
assume image name == container name.

This workaround simply removes the prefix "localhost/"
from the image name (if found).

### DIFF
--- a/src/libs/sfdk/dockervirtualmachine.cpp
+++ b/src/libs/sfdk/dockervirtualmachine.cpp
@@ -491,7 +491,9 @@ QStringList DockerVirtualMachinePrivate::listedImages(const QString &output)
 {
     QStringList images;
     const QStringList lines = output.split(QRegularExpression("[\r\n]"), Qt::SkipEmptyParts);
-    for (const QString &line : lines) {
+    for (QString line : lines) {
+        if (line.startsWith("localhost/"))
+            line = line.right(line.size() - 10);
         if (line != "<none>")
             images.append(line);
     }


martyone:
It would be nice to see whether we can make docker behave equally with some common set of options. I.e. pass some additional options to 'docker image ls' that would make it print names prefixed with host name unconditionally. With the current options it does so for non-local images only.

vigejolla:
> It would be nice to see whether we can make docker behave equally with some common set of options. I.e. pass some additional options to 'docker image ls' that would make it print names prefixed with host name unconditionally. With the current options it does so for non-local images only.

I don't think it's possible to make docker behave like podman, and I don't think that would even make sense. It would make more sense to try to make podman behave more like docker, which it already quite well does. But currently even that doesn't seem possible (at least I couldn't quickly find a way).

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -491,7 +491,9 @@ QStringList DockerVirtualMachinePrivate::listedImages(const QString &output)
 {
     QStringList images;
     const QStringList lines = output.split(QRegularExpression("[\r\n]"), Qt::SkipEmptyParts);
-    for (const QString &line : lines) {
+    for (QString line : lines) {
+        if (line.startsWith("localhost/"))
+            line = line.right(line.size() - 10);
```
vigejolla, line 496: This seems like a good workaround to me, but I'm wondering if it would be better to only do it if we detect that docker is really podman? One way to detect could be running "docker --version". Real docker calls itself "Docker", while podman calls itself just "docker".

martyone:
> I don't think it's possible to make docker behave like podman, and I don't think that would even make sense. It would make more sense to try to make podman behave more like docker, which it already quite well does. But currently even that doesn't seem possible (at least I couldn't quickly find a way).

Whichever way of course :) Just get same output for same inputs from both. Isn't there some other command for listing images? Something suitable for machine parsing?

SfietKonstantin:
I have tried searching making podman image list similar to docker but didn't found anything either.
Podman seems to enforce the hostname (https://github.com/containers/buildah/issues/1034).

Now about trying to identify podman and differentiate it from docker that would be a good idea, and I might come back and revisit this. Right now, trying to make the docker/podman build system work under Fedora, and I'm piling workarounds one on top of another ...

martyone:
> I'm piling workarounds one on top of another

So this image list is not going to be the only difference to care about? It might be more convenient to write a shell wrapper around podman to fight the differencies then. You can either store it on PATH or set it via https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/libs/sfdk/sdk.cpp#L427

SfietKonstantin:
Oh, interesting !
I will see what I can do about that.

Differences I have are now also related to the "rootless" state of podman and cgroups v2 right now.
While a wrapper might help me get around the rootless thing (and the image prefix), it might not be able to fix the cgroup v2 issues.

martyone:
Regarding cgroups v2 this might help you https://forum.sailfishos.org/t/issues-with-installing-sailfish-sdk-docker-with-debian-bullseye-11/4454/9 .

SfietKonstantin:
I want to actually keep cgroups v2, so that's unfortunately not an option right now :(
Might become one if it is getting too painful.

martyone:
Maybe you just missed the detail that `systemd.unified_cgroup_hierarchy=false` option does not disable cgroups v2. It just forces systemd to use the hybrid hierarchy where both v1 a v2 coexist, the former at /sys/fs/cgroup/, the latter at /sys/fs/cgroup/unified.

SfietKonstantin:
Oh wow, interesting indeed. Will try this !
In the meantime, I will close this ticket. The wrapper script works quite well (even if I didn't manage to achieve what I want with the Docker based SDK)

PS: is this expected that inside the docker SDK, installing gcc in a target results in a non-working gcc ?

martyone:
Hm, I wouldn't expect it to break. It is not a thing you should normally do (compiler/toolchain is to be installed under the tooling instead) but I vaguely remember seeing packages that specify `BuildRequires: gcc` (or so), so it should not break at least. It should just ignore gcc from target and use the one from tooling implicitly. But you may be luckily invoking gcc a way that is not intercepted by sb2 and then you end up running gcc from target, which may break. So, are you sure you need to install gcc under a target?

PS: looking forward to seeing your results with the podman wrapper etc.

vigejolla:
> In the meantime, I will close this ticket. The wrapper script works quite well (even if I didn't manage to achieve what I want with the Docker based SDK)

Could you share your wrapper script? It could be useful for others using podman, I'm sure you are not the only one :)

SfietKonstantin:
I will put the wrapper script online.
It's not a script per se, more a Rust executable ;)

SfietKonstantin:
@vigejolla I have shared my investigations and some details in this little write up https://sfietkonstantin.github.io/2021/04/29/Sailfish-SDK-Fedora.html. I also shared the scripts in it. I hope that it's clear enough.

vigejolla:
Thanks for the write up! I enjoyed reading it, even if the end result wasn't exactly what we had hoped for. However, there's still one thing that you could perhaps try: Since you already have the docker wrapper script, perhaps you could add to it the --uidmap parameter, mapping from your host user id to the mersdk user id inside the container. I also completely understand if you feel that you have already tried enough hacks :)

SfietKonstantin:
Hello @vigejolla 
I have actually tried uidmap ! The problem with that is that it changes all files owned by root to make them (eg) owned by mersdk. This prevent systemd from booting in the container ...
I think that, without a change in design for the container (do not boot SSHD, use docker exec to perform operations, use root as default user), it will be hard to use podman ...
Fortunately I have other solutions :)

PR #416, opened 2021-04-29
Jb53194

martyone:


martyone:

src/plugins/mer/merbuildengineoptionswidget.cpp
```diff
@@ -643,6 +644,7 @@ void MerBuildEngineOptionsWidget::onSrcFolderApplyButtonClicked(const QString &n
     buildEngine->virtualMachine()->lockDown(false, this, IgnoreAsynchronousReturn<bool>);
 
     if (ok) {
+        Core::DocumentManager::setProjectsDirectory(buildEngine->sharedSrcPath());
```
vigejolla, line 647: Wouldn't it make sense to first check if current projectsDirectory already happens to be under workspace?

martyone: Yup, done.

vigejolla (approved):
LGTM

PR #418, opened 2021-05-03
Jb53314

martyone:


martyone:

src/tools/sfdk/commandlineparser.cpp
```diff
@@ -168,6 +176,18 @@ CommandLineParser::CommandLineParser(const QStringList &arguments)
...
+    for (const QString &value : parser.values(COption)) {
+        // setCurrentWorkingDirectory() cannot be used before SdkManager is instantiated
+        m_cwdChanges += [=]() -> bool {
+            QString errorString;
+            if (!setCurrentWorkingDirectory(value, &errorString)) {
+                badUsage(invalidArgumentToOptionMessage(errorString, COption.names().first(), value));
+                return false;
+            }
+            return true;
+        };
```
vigejolla, line 188: Wouldn't it be simpler to store just the directories in m_cwdChanges, as a stringlist? Feels kind of silly to store these functions instead, when it is the same function for each and every directory.

martyone: It needs to access the COption variable.

martyone:
Updated so that it only recognizes "$WORKSPACE" when it matches the full argument. Specific paths under workspace may be referred by passing -C multiple times.

src/tools/sfdk/commandlineparser.cpp
vigejolla: In order to not just hardcode "C" here? I'm a bit undecided here which one is worse, the disease or the cure. Well, it's your code. I just wish I don't have to fix bugs here.

martyone: It's just a task executed asynchronously, waiting for a precondition. Compared to spreading the data and code throughout the CommandLineParser class, this way the task is nicely encapsulated in its own class (lambdas are kind of functor classes).

src/tools/sfdk/sfdkconstants.h
```diff
@@ -35,6 +35,8 @@ const char GENERAL_DOMAIN_NAME[] = "general";
 const char BUILD_ENGINE_SYSTEM_BUS_CONNECTION[] = "build-engine-system-bus";
 const char SAILFISH_SDK_SFDK_DBUS_SERVICE[] = "SAILFISH_SDK_SFDK_DBUS_SERVICE";
 
+const char WORKSPACE_PSEUDO_VARIABLE[] = "$WORKSPACE";
```
vigejolla, line 38: I'm not sure if using $ in the pseudo variable name is a good idea. It requires escaping if used from shell, and it's easy to confuse with environment variables. On the other hand, you probably had good reasons for choosing this approach - I would just like to hear them :)

src/tools/sfdk/commandlineparser.cpp
vigejolla: Yeah, there is value there. It definitely tips the scale towards your approach. I'm still not sure if it tips it enough that I would choose the approach though ;)

src/tools/sfdk/sfdkconstants.h
martyone: Yup, I considered `%{WORKSPACE}`, which is the syntax used by qtc's macro expander, `@WORKSPACE@`, which is used by Qt IFW, but I avoided that exactly for the possibility for confusion and need to escape in installer code. I chose sh-like syntax as the same approach is used for `$ORIGIN` in RPAHT.

vigejolla (approved):
LGTM

PR #419, opened 2021-05-04
Remove QML imports which are not supported by Sailfish OS 4.1.0

vigejolla:


martyone:
There is the opposite issue as well. Not only that it offers modules it shouldn't, it also does not offer modules it should.

Try to type "import <Ctrl-SPACE>" in a QML file - it offers various Qt modules but none of our modules.

So we should add the info for the modules we actually provide - it supports that per-target. It looks for `$[QT_INSTALL_QML]/*-bundle.json` and only if none of the loaded bundles (if any) provide the `QtQuick 2.x` module, it falls back to the one coming with qtc. https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/plugins/qmljstools/qmljsbundleprovider.cpp#L126

vigejolla:
> There is the opposite issue as well. Not only that it offers modules it shouldn't, it also does not offer modules it should.
> 
> Try to type `import<SPACE><Ctrl-SPACE>` in a QML file - it offers various Qt modules but none of our modules.
> 
> So we should add the info for the modules we actually provide - it supports that per-target. It looks for `$[QT_INSTALL_QML]/*-bundle.json` and only if none of the loaded bundles (if any) provide the `QtQuick 2.x` module, it falls back to the one coming with qtc. https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/plugins/qmljstools/qmljsbundleprovider.cpp#L126

Now that's an interesting find. I'll dig a bit deeper.

vigejolla:
> > There is the opposite issue as well. Not only that it offers modules it shouldn't, it also does not offer modules it should.
> > Try to type `import<SPACE><Ctrl-SPACE>` in a QML file - it offers various Qt modules but none of our modules.
> > So we should add the info for the modules we actually provide - it supports that per-target. It looks for `$[QT_INSTALL_QML]/*-bundle.json` and only if none of the loaded bundles (if any) provide the `QtQuick 2.x` module, it falls back to the one coming with qtc. https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/plugins/qmljstools/qmljsbundleprovider.cpp#L126

Yes, we could indeed use *-bundle.json files in targets. However:
 - that would require changes in the targets
  - preferably so that each package which provides imports includes a bundle file
 - This PR is specifically about fixing the bug about extra imports. I think adding the bundle files to the targets is no longer even just a bug fix, it's a task, or maybe even a story.
 - We can't add bundle files to old targets, so it makes sense to modify the default anyway.
 - I suggest we merge this as is and create a task or a story about the bundle files in targets.

martyone:
I was rather thinking of generating those on the fly. It should be possible to query the information from the plugins.qmltypes files or even more easily from rpmvalidator where we already do similarly with the docs. Should be quite easy, what do you think?

martyone (approved):
Feel free to merge but I would choose to quickly do the proper fix.

vigejolla:
> I was rather thinking of generating those on the fly. It should be possible to query the information from the plugins.qmltypes files or even more easily from rpmvalidator where we already do similarly with the docs. Should be quite easy, what do you think?

It's not as easy as you think. Even going through those manually, which I had to do in order to make this PR, was a bit difficult, as the plugins.qmltypes files don't actually give you the exact information required for the task.

rpmvalidator indeed has correct information, but unfortunately only for those imports which we allow in Harbour. Most of the APIs are not allowed, but still it would be nice to support them in QtC.

martyone:
Omitting them from the bundle does not mean they would not be supported in QtC. They would just not be suggested for inclusion, but once included, the code assistance would work normally - this is what we have nowadays anyway. Similarly we bundle documentation for harbour-allowed APIs only. So it would be an advantage in this sense.

vigejolla:
> Omitting them from the bundle does not mean they would not be supported in QtC. They would just not be suggested for inclusion, but once included, the code assistance would work normally - this is what we have nowadays anyway. Similarly we bundle documentation for harbour-allowed APIs only. So it would be an advantage in this sense.

I still think it would be better to also suggest those. And we do bundle documentation for APIs which are not allowed in Harbour - e.g. sailfish secrets. Many use those APIs, they just don't submit their apps to Harbour. And that's just fine.

PR #420, opened 2021-05-05
Jb54141

martyone:


share/qtcreator/sfdk/modules/20-building-mb2.json
```diff
@@ -163,5 +168,34 @@
...
             "trDescription": "In case a passphrase-protected signing key is selected for signing packages, read the passphrase from the <file>. Only the first line will be read."
         }
+    ],
+
+    "hooks":
+    [
+        {
+            "name": "pre-pull-build-requires",
+            "trSynopsis": "[<dependency>...]",
+            "trDescription": "This hook is invoked by commands that take care of updating build time dependencies – prior to doing so.\n\nThe arguments and available environment variables are the same as with the 'pull-build-requires' hook."
```
vigejolla, line 178: Consider e.g. "This hook is invoked by commands that take care of updating build time dependencies, before performing the update or install of the dependencies."

src/tools/sfdk/command.h
```diff
@@ -51,6 +51,18 @@ class Domain;
 class Emulator;
 class Module;
 
+class Hook
```
vigejolla, line 54: Why not struct? This looks a lot like a struct to me...

share/qtcreator/sfdk/modules/20-building-mb2.json
martyone: hm, here the first sentence would use the word update in the general sense of getting the requirements fulfilled (somehow), while the second sentence would use it in the specific sense of updating packages that fulfill the requirements. I would rather leave it as is - it should be clear enough and should there be any doubt, the "build-requires pull" command states more details.

src/tools/sfdk/command.h
martyone: It looks like a struct but it is not meant to be thought of as such. It is related to the Command, Option, Domain, Module, etc. classes - some look more like structs, some less, but all are used in the same context, so mixing class/struct would not be good.

share/qtcreator/sfdk/modules/20-building-mb2.json
vigejolla: Well it's not clear to me what the "doing so" refers to.

src/tools/sfdk/command.h
vigejolla: fair enough

vigejolla (approved):
LGTM, however:

 - As this introduces quite big changes, I suggest we postpone merging until after the SDK 3.5 release
 - I think this feature deserves a blog post or something when we release this. Giving examples how to use and such.

share/qtcreator/sfdk/modules/20-building-mb2.json
martyone: It refers to "updating build time dependencies". Does this help? https://ludwig.guru/s/prior+to+doing+so

vigejolla: Perhaps you would like to add that link to the description then, to help make it clear what you mean?

martyone:
Rebased and did minor renaming in accordance with upcoming changes.

PR #421, opened 2021-05-06
Jb53939

martyone:


src/tools/sfdk/main.cpp
```diff
@@ -88,6 +88,10 @@ void initQSsh()
 
 int main(int argc, char **argv)
 {
+    // Avoid redirecting output to system logs, which happens (not only) when
+    // sfdk is invoked by GUI applications on Windows.
+    qputenv("QT_FORCE_STDERR_LOGGING", "1");
```
vigejolla, line 93: After using Debugview on Windows for decades, I really dislike this change. Please make it at least configurable instead of forcing it on users.

martyone: This is about output that is desired to always appear on stderr. This is the same class of messages as qmake's or gcc's reports on build errors. Not something to go to debugview.

vigejolla: I disagree (partly). Yes, the usual output stuff should indeed appear on stderr. However, possible debug prints should appear on Debugview. That's the whole purpose of the tool. The preferred behaviour would be to have qDebug() stuff go to DebugView while qWarning() and qError() would go to stderr. However, I don't think that's possible with Qt :( Therefore, I repeat my plea: Please make this configurable (maybe with yet another environment variable).

martyone: Done, but tell me: what would be an use case for this?

qDebug output needs to be enabled explicitly with the `--debug` option. When I use that option I certainly want to see the output interleaving other messages and not detached into a separate view. This is a CLI build tool and I fail to see an analogy with some other CLI build tool that would behave like you suggest (considering journal/syslog on Linux). At the same time it's crucial that the more urgent messages appear on stderr and as you said, separating those is not out of box supported with Qt.

vigejolla (approved):
LGTM. Thank you for the if!

PR #422, opened 2021-05-10
Jb53317

martyone:


src/tools/sfdk/task.cpp
```diff
@@ -424,12 +424,24 @@ BOOL WINAPI SignalHandler::ctrlHandler(DWORD fdwCtrlType)
             qCDebug(sfdk) << "Ctrl-C event ignored";
             return TRUE;
         }
-        Q_FALLTHROUGH();
-    case CTRL_CLOSE_EVENT:
+        TaskManager::process(TaskManager::Terminate);
```
vigejolla, line 427: Just curious: Why did you previously choose to use Q_FALLTHROUGH, and now instead duplicate code?

vigejolla: Yup, it's just one additional copy (two lines if you count the return). I'm mostly wondering if you have some magic rule of not copying more than X lines of code or something?

PR #425, opened 2021-05-17
Jb54269

martyone:


martyone:

src/tools/sfdk/sdkmanager.cpp
```diff
@@ -1387,9 +1387,13 @@ bool SdkManager::stopReliably(VirtualMachine *virtualMachine)
 {
     QTC_ASSERT(virtualMachine, return false);
     bool ok;
+    if (isRunningReliably(virtualMachine)) {
```
vigejolla, line 1390: As far as I understand, if the vm is just starting, isRunningReliably returns false. Should we care?

martyone: Actually isRunningReliably and the related isEngineRunning, isEmulatorRunning etc only care about the basic VM state. sfdk (the CLI) makes as much assumptions as possible in order to minimize startup time, so it never tries to really connect to the VM in the LBYL manner and instead it works with the basic VM state as long as possible.

vigejolla: I think I was confused by the excessive amounts of negations in the logic. I'm just going to assume that it's ok.

vigejolla (approved):
LGTM

PR #426, opened 2021-05-17
Sfdk: Forward --snapshots-of argument to sdk-assistant

vigejolla:


src/tools/sfdk/command.cpp
```diff
@@ -1660,12 +1660,24 @@ Worker::ExitStatus BuiltinWorker::runTools(const QStringList &arguments_, int *e
     }
 
     if (arguments.first() == "remove") {
-        if (!P::checkPositionalArgumentsCount(arguments, 2, 2))
+        QCommandLineParser parser;
+        QCommandLineOption snapshotsOfOption(QStringList{"snapshots-of"});
+        parser.addOption(snapshotsOfOption);
```
martyone, line 1665: I would prefer using the plural version and separate with blank line for consistency :)

```suggestion

        parser.addOptions({snapshotsOfOption});
```

src/tools/sfdk/command.cpp
```diff
@@ -1660,12 +1660,24 @@ Worker::ExitStatus BuiltinWorker::runTools(const QStringList &arguments_, int *e
...
+        parser.addPositionalArgument("name", QString(), "[name]");
+
+        if (!parser.parse(arguments)) {
+             qerr() << parser.errorText() << endl;
+             return BadUsage;
+        }
+
+        QStringList positionalArguments = parser.positionalArguments();
+        if (positionalArguments.count() != 1)
```
martyone, line 1674: Unless I missed something, `P::checkPositionalArguments` should be used as it also prints an appropriate message in case of error. I would also omit the variable for consistency - `QCommandLineParser::positionalArguments` is likely an inline function anyway.

```suggestion
        if (!P::checkPositionalArgumentsCount(parser.positionalArguments(), 1, 1))
```

src/tools/sfdk/command.cpp
vigejolla: Right, now I remember you complained about the same thing the last time I did changes here...

src/tools/sfdk/command.cpp
```diff
@@ -1660,12 +1660,24 @@ Worker::ExitStatus BuiltinWorker::runTools(const QStringList &arguments_, int *e
     }
 
     if (arguments.first() == "remove") {
-        if (!P::checkPositionalArgumentsCount(arguments, 2, 2))
+        QCommandLineParser parser;
+        QCommandLineOption snapshotsOfOption(QStringList{"snapshots-of"});
```
martyone, line 1664: This should work too https://doc.qt.io/qt-5/qcommandlineoption.html#QCommandLineOption

```suggestion
        QCommandLineOption snapshotsOfOption("snapshots-of");
```

martyone (approved):
One more hint, otherwise LGTM.

src/tools/sfdk/command.cpp
vigejolla: I'm using QStringList there for consistency ;) No, just kidding, will change

PR #427, opened 2021-05-17
Jb54282

martyone:


martyone:

martyone:
https://codereview.qt-project.org/c/qt-creator/qt-creator/+/349540

vigejolla (approved):
This is used only for displaying paths? The discussion on upstream review was interesting. I'm slightly worried about what happens when someone has a path which actually has a backslash in the name. But if it's good enough for upstream it's good enough for me. And it's nice to know that the "real" fix, i.e. pathnames already knowing which separators they use, is already cooking in upstream.

martyone:
Nope, it's not only for displaying. It modifies the real data.

PR #429, opened 2021-05-19
Jb54306

martyone:


src/tools/sfdk/command.cpp
```diff
@@ -1852,9 +1852,7 @@ bool BuiltinWorker::listTools(SdkManager::ListToolsOptions options, bool listToo
     }
 
     TreePrinter::Tree tree = TreePrinter::build(table, 0, 1);
-    TreePrinter::sort(&tree, 0, 0, true);
-    TreePrinter::sort(&tree, 1, 0, true);
-    TreePrinter::sort(&tree, 2, 0, true);
+    TreePrinter::sort(&tree, 0, true);
```
vigejolla, line 1855: I believe after these changes TreePrinter::sort(Tree* tree, int level, int column, bool ascending) is no longer ever called with level != 0. It could be simplified, as the else branch is dead code now.

martyone: True, but a new use case may easily appear in future, so I preserved the functionality.

vigejolla: Well, git does preserve history, so in that sense it does not make sense to keep old dead code around just because it might be needed someday. But your call, just wanted to point that out in case you didn't notice.

vigejolla (approved):
LGTM

PR #430, opened 2021-05-20
Jb53937

martyone:


martyone:

martyone:
https://codereview.qt-project.org/c/qt-creator/qt-creator/+/350182

vigejolla (approved):
LGTM, with one minor nitpick: fullLocalCommandLine() no longer lives up to its expectations, as it does not contain the ssh executable. Something like "fullLocalArguments" might be a more appropriate name.

martyone:
I was a bit undecided about this too, but then I found this is the original name that was used before the return type was changed to CommandLine - see commit c36b1e40b0db99c5f9903cfb0aa614842c210b38, which is basically what this commit reverts (does not apply cleanly anymore).

PR #431, opened 2021-05-20
Jb54310

martyone:


martyone:
The very first commit belongs to another PR

martyone:
Initially I tried to remove the need to connectTo prior to lockDown, but that would not be a trivial extension to the state machine.

src/tools/sfdk/command.cpp
```diff
@@ -1674,7 +1674,7 @@ Worker::ExitStatus BuiltinWorker::runTools(const QStringList &arguments_, int *e
         if (!P::checkPositionalArgumentsCount(parser.positionalArguments(), 1, 1))
             return BadUsage;
 
-        const QString name = positionalArguments.first();
+        const QString name = parser.positionalArguments().first();
```
vigejolla, line 1677: How on earth did this end up in the commit :-o talking about the non-compiling code of course

vigejolla (approved):
LGTM. Please file a bug about checking if the configuration has changed, unless you have already done so.

PR #432, opened 2021-05-21
Jb53937

martyone:


martyone:
Only later I found the function is actually used outside the QSsh lib in the ProjectExplorer plugin, so the signature may not change :\

This unfortunately also means the fix will not apply as broadly as originally assumed. At least it doesn't seem to affect frequently used features - only remote commands executed in console are affected.

martyone:
Wrong branch name - I will correct it in the summary message when merging.

vigejolla (approved):
LGTM

PR #434, opened 2021-05-28
Jb54236

martyone:


vigejolla (commented):
"Refactor build target handling" as a commit message doesn't really say anything. "Use BuildTargetData where suitable" would be more meaningful. But that's just my opinion..

vigejolla (approved):
LGTM

PR #443, opened 2021-06-24
Jb54737

martyone:


martyone:

vigejolla (approved):
LGTM

vigejolla (commented):
... when EA and latest are of the same version. LGTM.

martyone:
Here I really meant latest non-EA, but later, when working on alignment of the listing with and without --available, I found it better to keep the same order and list EA first (later PR does it)

PR #445, opened 2021-07-02
Semiautomatic vm storage resizing

vigejolla:


src/libs/sfdk/virtualmachine.cpp
```diff
@@ -402,6 +402,67 @@ void VirtualMachine::setStorageSizeMb(int storageSizeMb, const QObject *context,
...
+}
+
+void Sfdk::VirtualMachine::setFreeSizeMb(int freeSizeMb, int incrementMb, bool allowShrink,
+                                         const QObject *context, const Functor<bool> &functor)
+{
+    const QPointer<const QObject> context_{context};
+
+    auto lockAndResize = [=](int storageSizeMb) {
+        lockDown(true, context, [=](bool ok) {
+            if (ok)
```
martyone, line 440: We should do `functor(false); return;` on `!ok`. More occurrences elsewhere.

src/libs/sfdk/virtualmachine.h
```diff
@@ -126,6 +126,10 @@ class SFDK_EXPORT VirtualMachine : public QObject
     void setStorageSizeMb(int storageSizeMb, const QObject *context,
             const Functor<bool> &functor);
 
+    void currentFreeSize(const QObject *context, const Functor<int> &functor);
+    void setFreeSizeMb(int freeSizeMb, int incrementMb, bool allowShrink,
+            const QObject *context, const Functor<bool> &functor);
+
```
martyone, line 132: It makes me ask how does `incrementMb` affect shrinking? So I would rather keep this API grow-only and let sfdk fail if '+' is not used right now (shrinking not supported by any of our VMs yet).

I would prefer `storageFreeSizeMb` and `ensureStorageFreeSizeMb` as the names and stick them together with the above functions as in:

```
storageSizeMb
freeStorageSizeMb
setStorageSizeMb
ensureStorageSizeMb
```

src/libs/sfdk/virtualmachine.cpp
```diff
@@ -402,6 +402,67 @@ void VirtualMachine::setStorageSizeMb(int storageSizeMb, const QObject *context,
...
+                        freeSize = -2;
+                }
+                if (context)
+                    functor(freeSize);
+            });
+        }
+    };
+
+    if (isOff()) {
+        connectTo(NoConnectOption, this, vmStarted);
```
martyone, line 427: This is problematic. So far the API is explicit about starting/stopping VMs, so we should not break that. We should either cache the free space, so that the VM does not need to be started in order to query it (preferred) or assert that the VM is started (connected) when this function is used. With the latter option we would then need to shut the VM down for the resize and that would hit similar issue. So the former option seems to be not only the preferred but also the only viable option.

src/tools/sfdk/command.cpp
```diff
@@ -198,7 +198,7 @@ class SetPropertiesTask : public Task
         bool lockDownOk = false;
 
         if (m_needsVmOff) {
-            if (SdkManager::isRunningReliably(m_virtualMachine)) {
+            if (!qEnvironmentVariableIsSet(Constants::SFDK_AUTO_STOP_VMS_ENV_VAR) && SdkManager::isRunningReliably(m_virtualMachine)) {
```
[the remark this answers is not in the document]
vigejolla: sure thing

src/libs/sfdk/virtualmachine.h
vigejolla: > It makes me ask how does `incrementMb` affect shrinking? So I would rather keep this API grow-only and let sfdk fail if '+' is not used right now (shrinking not supported by any of our VMs yet).

incrementMb disables shrinking. Or at least that was my plan. When I realized shrinking isn't supported I stopped worrying about it. Sure, I can remove shrinking from the API, it certainly makes things simpler.

src/libs/sfdk/virtualmachine.cpp
vigejolla: I find caching problematic as well, as pretty much everything we do in the vm affects the free space. We should update the cache every time we shut down the vm, and I don't like the idea of doing anything extra when shutting down.

martyone: That was my first idea too, to simply save it at shutdown. In addition to that, most actions go through mb2 and sdk-manage, so those two could trigger that too easily. I wouldn't be too scare of getting out of sync as much as to actually break things easily.

vigejolla: I'll try with the "save at shutdown" approach. It should be relatively easy to implement.

src/libs/sfdk/buildengine.h
```diff
@@ -124,6 +124,10 @@ class SFDK_EXPORT BuildEngine : public QObject
     void importPrivateGpgKey(const QString &id, const Utils::FilePath &passphraseFile,
             const QObject *context, const Functor<bool, QString> &functor);
 
+    int freeSize() const;
+    void setFreeSizeMb(int freeSizeMb, int incrementMb, const QObject *context,
+            const Functor<bool> &functor);
+
```
martyone, line 130: Compare this with `setStorageSizeMb`, which is also not supported on emulator. Still it is exposed on `VirtualMachine` and we have those `GrowStorageSize` and `ShrinkStorageSize` feature flags to indicate its availability. Similarly we should expose `freeStorageSizeMb` and `reserveStorageSizeMb` (*) on `VirtualMachine` and add the feature flag `ReserveStorageSize` to indicate their availability.

(*) Before I suggested to use "ensure" in the name, but "reserve" will work better for the feature flag, so let's use that instead.

vigejolla: The reason I moved it to `BuildEngine` is that `VirtualMachine` does no consept of shared folders - those are implemented on `BuildEngine` and `Emulator` level. Are you sure you want me to move the shared folders to `VirtualMachine`? Well, `sharedConfigPath` at least. Both emulators and build engines have those, so it would make kind of sense.

martyone: Actually the information is available at that level, even that it is not exposed in the public interface of `VirtualMachine` - you can get the shared path as `d->cachedInfo().sharedConfig` there. See how it is used in `DockerVirtualMachinePrivate::makeCreateArguments()`.

You will need to add the `ReserveStorageSize` flag to `VBoxVirtualMachine::staticFeatures` and to `EmulatorPrivate::initVirtualMachine()::unsupportedFeatures` to indicate its availability just on vbox-based build engine.

vigejolla: Thanks! I do have to wonder how I missed that.

src/libs/sfdk/virtualmachine.h
```diff
@@ -126,6 +127,10 @@ class SFDK_EXPORT VirtualMachine : public QObject
     void setStorageSizeMb(int storageSizeMb, const QObject *context,
             const Functor<bool> &functor);
 
+    int freeStorageSizeMb() const;
```
martyone, line 130: I would prefer ordering `freeStorageSizeMb` directly after `storageSizeMb` and keep these four stuck together without separating blank line. Same ordering in .cpp.

src/libs/sfdk/virtualmachine.cpp
```diff
@@ -402,6 +402,55 @@ void VirtualMachine::setStorageSizeMb(int storageSizeMb, const QObject *context,
     });
 }
 
+int Sfdk::VirtualMachine::freeStorageSizeMb() const
+{
+    Q_D(const VirtualMachine);
+    QTC_ASSERT(d->initialized(), return false);
```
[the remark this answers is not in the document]
vigejolla: Sure. And probably return something else than false.

src/libs/sfdk/virtualmachine.cpp
```diff
@@ -402,6 +402,55 @@ void VirtualMachine::setStorageSizeMb(int storageSizeMb, const QObject *context,
...
+        lockDown(true, context, [=](bool ok) {
+            if (!ok) {
+                callIf(context_, functor, false);
+                return;
+            }
+            setStorageSizeMb(storageSizeMb, context, functor);
+        });
+    };
+
+    int currentFreeSizeMb = currentFreeSize / 1024;
```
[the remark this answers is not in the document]
vigejolla: Yes, it *should*

src/libs/sfdk/virtualmachine.cpp
```diff
@@ -402,6 +402,55 @@ void VirtualMachine::setStorageSizeMb(int storageSizeMb, const QObject *context,
...
+
+    if (currentFreeSizeMb < minFreeStorageSizeMb) {
+        lockAndResize(storageSizeMb() + (incrementMb ? incrementMb : minFreeStorageSizeMb - currentFreeSizeMb));
+    } else {
+        BatchComposer::enqueueCheckPoint(context, std::bind(functor, true));
+    }
```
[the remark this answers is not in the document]
vigejolla: Actually, we shouldn't even be locking down at this stage...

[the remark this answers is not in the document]
vigejolla: Please check the current version if that's ok in your opinion

src/libs/sfdk/virtualmachine.cpp
```diff
@@ -372,6 +372,23 @@ int VirtualMachine::storageSizeMb() const
...
 
+int VirtualMachine::freeStorageSizeMb() const
+{
+    Q_D(const VirtualMachine);
+    QTC_ASSERT(d->initialized(), return -1);
+    QTC_ASSERT(d->features & ReserveStorageSize, return -1);
+    const Utils::FilePath dfPath = Utils::FilePath::fromString(d->cachedInfo().sharedConfig)
+            .pathAppended("df.cache");
+    FileReader reader;
+    if (!reader.fetch(dfPath.toString(), QIODevice::ReadOnly))
```
martyone, line 383: We should say something qCritical here and below.

src/tools/sfdk/command.cpp
```diff
@@ -325,6 +326,26 @@ class VirtualMachinePropertiesAccessor : public PropertiesAccessor
...
+                *errorString = valueCannotBeDecreasedMessage();
+                return Failed;
+            }
+            if (!parsePositiveInt(&m_freeSizeMb, freeSize, errorString))
+                return Failed;
+            if (commaIndex > 0) {
+                if (!parsePositiveInt(&m_incrementMb, value.mid(commaIndex + 1), errorString))
+                    return Failed;
+                if (m_incrementMb < m_freeSizeMb) {
+                    *errorString = tr("Increment can't be smaller than minimum free storage size");
```
martyone, line 344: Maybe `<step> cannot be smaller than <size>`, provided that the accepted argument would be documented as `<size>[+][,<step>]`. BTW, the docs need to be updated in the module definition.  (There was also a rule that contractions shouldn't be used in messages IIRC, hence "can't" -> "cannot".)

src/tools/sfdk/command.cpp
```diff
@@ -362,7 +383,12 @@ class VirtualMachinePropertiesAccessor : public PropertiesAccessor
                     m_vm.data(), m_storageSizeMb);
             ok &= stepOk;
         }
-
```
martyone, line 363: Keep separated above/below please. Similarly on line 582.

src/libs/sfdk/virtualmachine.cpp
vigejolla: sure

src/tools/sfdk/command.cpp
vigejolla: Previously we discussed that this wouldn't be documented, at least for the time being. This was supposed to be used by the installer only.

src/libs/sfdk/virtualmachine.cpp
```diff
@@ -372,6 +372,27 @@ int VirtualMachine::storageSizeMb() const
...
+int VirtualMachine::freeStorageSizeMb() const
+{
+    Q_D(const VirtualMachine);
+    QTC_ASSERT(d->initialized(), return -1);
+    QTC_ASSERT(d->features & ReserveStorageSize, return -1);
+    const Utils::FilePath dfPath = Utils::FilePath::fromString(d->cachedInfo().sharedConfig)
+            .pathAppended("df.cache");
+    FileReader reader;
+    if (!reader.fetch(dfPath.toString(), QIODevice::ReadOnly)) {
+        qCritical(lib) << "Unable to read df.cache: " << reader.errorString();
```
martyone, line 384:
```suggestion
        qCritical(lib).noquote() << "Unable to read df.cache:" << reader.errorString();
```

src/tools/sfdk/command.cpp
```diff
@@ -363,6 +384,12 @@ class VirtualMachinePropertiesAccessor : public PropertiesAccessor
             ok &= stepOk;
         }
 
+        if (m_freeSizeMb > 0) {
+            bool stepOk;
+            execAsynchronous(std::tie(stepOk), std::mem_fn(&VirtualMachine::reserveStorageSizeMb),
+                             m_vm.data(), m_freeSizeMb, m_incrementMb);
+            return stepOk;
+        }
```
martyone, line 392:
```suggestion
        }

```

src/tools/sfdk/command.cpp
```diff
@@ -72,6 +72,7 @@ const char VM_MEMORY_SIZE_MB[] = "vm.memory-size";
 const char VM_SWAP_SIZE_MB[] = "vm.swap-size";
 const char VM_CPU_COUNT[] = "vm.cpu-count";
 const char VM_STORAGE_SIZE_MB[] = "vm.storage-size";
+const char VM_FREE_STORAGE_SIZE_MB[] = "vm.free-storage-size";
```
martyone, line 75: > Previously we discussed that this wouldn't be documented, at least for the time being. This was supposed to be used by the installer only.

Right, and that is great because only now I realize this would not be easy to document consistently with the existing properties. So yes, please omit the docs change.

```suggestion
const char VM_FREE_STORAGE_SIZE_MB[] = "vm.free-storage-size"; // intentionally undocumented, write-only
```

martyone (approved):
Last suggestions. I would also squash the two commits to sfdk, the lib and the tool.

PR #447, opened 2021-08-10
Mer: Do not update vm storage limits before value updated. JB#55104

etaishev:


src/plugins/mer/mervirtualmachinesettingswidget.cpp
```diff
@@ -73,7 +73,6 @@ void MerVirtualMachineSettingsWidget::setVmFeatures(VirtualMachine::Features fea
     else
         setToolTip(ui->cpuInfoLabel, tr("Virtual Machine does not allow changing cpu count"));
 
-    setStorageSizeLimits();
```
martyone, line 76: I am afraid this is definitely not a valid fix for the issue you observe. `setStorageSizeLimits` must be called here.

etaishev: What for? Nowhere above is the current disk size set and only the limits are set

martyone: It sets the limits for the *current* disk size (as it is known to the SDK) at the time the widget is initialized and every time the disk size is changed using the widget. Check the code. It is not only done to restrict it to growing/shrinking depending on the available VM features, but also to restrict the maximum increment in case of growing.

etaishev: The `setStorageSizeLimits();` calls twice: from `setVmFeatures()` and `setStorageSizeMb()`. When `setStorageSizeLimits` is called from `setVmFeatures` the disk value is not set and it is set as `0+MAX_STORAGE_SIZE_INCREMENT_GB`. At the moment when `setStorageSizeMb` is called there is already a limit, which is based on wrong data

martyone: This is because the order of calling `setVmFeatures` and `setStorageSizeMb` inside `MerEmulatorDetailsWidget::setEmulator` and `MerBuildEngineDetailsWidget::setBuildEngine` is wrong - this is what should be fixed instead. Be sure to test very well the effect of correcting the order as there may be conflicting requirements!

PR #450, opened 2021-08-18
Jb55177

martyone:


martyone:
Under remotelinux they also omit '--' https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/plugins/remotelinux/remotelinuxsignaloperation.cpp#L52

src/tools/sfdk/remoteprocess.cpp
```diff
@@ -240,7 +240,7 @@ void RemoteProcess::kill(const QString &signal, void (RemoteProcess::*callback)(
         m_killRunner.reset(nullptr);
     });
 
-    QString killCommand = QString("kill -%1 -- -%2 %2").arg(signal).arg(m_processId);
+    QString killCommand = QString("kill -%1 -%2 %2").arg(signal).arg(m_processId);
```
vigejolla, line 243: Why is the process ID there twice, with and without a preceding minus?

martyone: Negation of a PID results in killing the process group the process leads.

martyone: Negation of a PID results in killing the process group the process leads.

vigejolla: TIL :)

PR #467, opened 2021-09-08
Jb55462

martyone:


martyone:

vigejolla (approved):
LGTM, with a small caveat: I didn't try how this actually feels, as I didn't have a recent enough build of QtC and building takes ages. It's building now so I'll test when it's done.

martyone:
Polished once more. Now it delays displaying the progress dialog when there is no pop up dialog open - the assumption is that in this case the action was triggered using the buttons in bottom left corner and those already provide necessary feedback. The progress bar will only show if it takes too long - like before. Also fixed stacking the progress bar on top of other windows when a pop up dialog is open (e.g. when starting/stopping the VM from options dialog).

vigejolla:
tested now, both before and after polishing. My "lgtm" still stands :)

PR #468, opened 2021-09-09
Jb55451

martyone:


martyone:

vigejolla (approved):
Looks almost good to me. One thing worries me slightly: After this change, the added file won't be visible in diff. However, that is quite minor issue - I think we could file that as a separate issue and fix later (if at all).

martyone:
Yeah I hoped QtC would allow to diff the staged changes as well, but it does not seem to. At least it allows to invoke gitk and git-gui and those allow that. Otherwise I unfortunately do not see a way to improve that. I used to hit this issue occasionally in a completely different context ...until I gave up using --intent-to-add.

vigejolla:
> Yeah I hoped QtC would allow to diff the staged changes as well, but it does not seem to. At least it allows to invoke gitk and git-gui and those allow that. Otherwise I unfortunately do not see a way to improve that. I used to hit this issue occasionally in a completely different context ...until I gave up using --intent-to-add.

I didn't look into the details on how diff is implemented inside QtC. I assumed it wouldn't be impossible to diff also the staged changes, in addition to those which are not staged.

PR #469, opened 2021-09-10
Jb55307

martyone:


src/libs/sfdk/emulator.cpp
```diff
@@ -929,6 +924,16 @@ void EmulatorManager::fromMap(const QVariantMap &data, bool fromSystemSettings)
...
 }
 
+int EmulatorManager::doAddEmulator(std::unique_ptr<Emulator> &&emulator)
+{
+    QTC_ASSERT(emulator, return -1);
+
+    m_emulators.emplace_back(std::move(emulator));
+    const int index = m_emulators.size() - 1;
+    emit emulatorAdded(index);
+    return index;
```
vigejolla, line 957: Why do we bother returning the index? I couldn't find the return value actually used anywhere. The same applies to the build engine.

martyone: It's a library API, so it cares about _possible_ use cases, not just those actually realized by us. See how the behavior changes now: before it was appending, now it is inserting at some position. Potential code written with the assumption that new item appears at the end of the list would broke now. Potential code working with the returned index with no assumptions would happily continue working.

vigejolla: worrying about not existing use cases feels premature overengineering to me. But I'm fine with that, as long as it's a conscious decision, and not just some accidental leftover.

vigejolla (approved):
LGTM

PR #470, opened 2021-09-13
Jb55451

martyone:


martyone:

src/plugins/git/settingspage.ui
```diff
@@ -123,6 +123,16 @@
         </property>
        </widget>
       </item>
+      <item row="2" column="0">
+       <widget class="QCheckBox" name="addImmediatelyCheckBox">
+        <property name="toolTip">
+         <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Do not use the &lt;span style=&quot; font-family:'Hack';&quot;&gt;--intent-to-add&lt;/span&gt; option to &lt;span style=&quot; font-family:'Hack';&quot;&gt;git-add&lt;/span&gt;.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
vigejolla, line 129: The html is not only unnecessary, I think it's also responsible for the text to be wrapped to three lines, which looks ugly. Without the html the end result looks better.

martyone: Yeah the way it wraps it is really unfortunate in this case.

martyone: updated

vigejolla (approved):
LGTM

PR #482, opened 2021-10-21
Jb49234

martyone:


vigejolla (approved):
LGTM. Bonus point for not doing it as unreadable oneliner.

PR #499, opened 2021-11-25
Jb56299

martyone:


share/qtcreator/sfdk/modules/80-ide-qmltypes.mjs
```diff
@@ -11,8 +11,10 @@ export function formatOption(name, argument, needsArgument, defaultFormat) {
         "task"
     ].flatMap(utils.expandCompacted);
 
-    if (mb2Options.indexOf(name) >= 0)
-        return [true, ["-Xmb2"].concat(defaultFormat)];
+    if (mb2Options.indexOf(name) >= 0) {
+        var format = defaultFormat.flatMap(argument => ["-Xmb2", argument]);
```
vigejolla, line 15: This is not exactly easy to follow, so please confirm that I have understood this correctly:
The old code returned an array with "-Xmb2" as the first element, and the arguments (which were passed as "defaultFormat") as the rest of the elements, i.e. ["-Xmb2", "arg1", "arg2"]. The new code instead appends -Xmb2 to all of the arguments, resulting in ["-Xmb2arg1", "-Xmb2arg2"]. Am i even close?

martyone: Nearly there. The result is `["-Xmb2", "arg1", "-Xmb2", "arg2"]`. Read `arr.flatMap(op)` as `arr.map(op).flat()` (required newer JS).

vigejolla: right, of course

vigejolla (approved):
LGTM

PR #502, opened 2021-12-07
Jb45571

martyone:


martyone:

src/libs/sfdk/emulator.cpp
```diff
@@ -377,9 +377,12 @@ bool EmulatorPrivate::fromMap(const QVariantMap &data)
 
     DeviceModelData deviceModel;
     EmulatorManager::fromMap(&deviceModel, data.value(Constants::EMULATOR_DEVICE_MODEL).toMap());
-    setDisplayProperties(deviceModel,
-            static_cast<Qt::Orientation>(data.value(Constants::EMULATOR_ORIENTATION).toInt()),
-            data.value(Constants::EMULATOR_VIEW_SCALED).toBool());
+    Qt::Orientation orientation = static_cast<Qt::Orientation>(
+            data.value(Constants::EMULATOR_ORIENTATION, Qt::Vertical).toInt());
+    if (orientation == 0)
+        orientation = Qt::Vertical;
```
vigejolla, line 383: Given that there are only two allowed values, I'd write this as:
`Qt::Orientation orientation = data.value(Constants::EMULATOR_ORIENTATION, Qt::Vertical).toInt() == Qt::Horizontal = Qt::Horizontal ? Qt::Vertical`

martyone: The `if ...` part is there only to fix user configuration broken by this bug in past. It is meant to be removed after a few releases. Hence the separation.

vigejolla: Ok, in that case it makes sense.

vigejolla (approved):
LGTM

PR #512, opened 2022-01-04
Sfdk: enqueue checkpoint in VirtualMachinePrivate::doInitGuest()

vigejolla:
The UI is waiting for the done signals from initGuest, so we need to do
something, even when we are doing nothing.


Thaodan (approved):
LGTM

PR #513, opened 2022-01-10
Sfdk: Move checkpoint enqueuing from VirtualMachinePrivate

vigejolla:
...to VmConnection, where the BatchComposer is instantiated.

### DIFF
--- a/src/libs/sfdk/vmconnection.cpp
+++ b/src/libs/sfdk/vmconnection.cpp
@@ -1082,6 +1082,7 @@ bool VmConnection::sshStmStep()
             BatchComposer composer = BatchComposer::createBatch("VmConnection::initGuest");
             m_guestInitBatch = composer.batch();
             emit initGuest();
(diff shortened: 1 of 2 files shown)


src/libs/sfdk/vmconnection.cpp
```diff
@@ -1082,6 +1082,7 @@ bool VmConnection::sshStmStep()
             BatchComposer composer = BatchComposer::createBatch("VmConnection::initGuest");
             m_guestInitBatch = composer.batch();
             emit initGuest();
+            composer.enqueueCheckPoint(this, [](){});
```
martyone, line 1085: It's a static member, so it would be better to call it as such to not suggest to the reader that it works on the `composer` instance. (It works on the currently opened batch.)

```suggestion
            BatchComposer::enqueueCheckPoint(this, [](){});
```

martyone (approved):

PR #520, opened 2022-01-28
Jb57232

martyone:


src/libs/sfdk/virtualmachine.cpp
```diff
@@ -677,6 +675,18 @@ VirtualMachinePrivate::~VirtualMachinePrivate()
...
 }
 
+void VirtualMachinePrivate::initGuest(const QObject *context, const Functor<> &functor)
+{
+    BatchComposer composer = BatchComposer::createBatch("VirtualMachine::initGuest");
+    const QPointer<const QObject> context_{context};
+
+    // Errors not considered fatal here
+    doInitGuest();
+    emit initGuest();
```
vigejolla, line 697: I'm quite sure this signal is no longer connected anywhere, and could be just removed

vigejolla (approved):
LGTM; you might want to check if that initGuest signal is still needed though

src/libs/sfdk/virtualmachine.cpp
martyone: Yup I am aware of that - preserving it as a guidance for possible future use.

PR #522, opened 2022-02-10
Sfdk: Add --no-rpmlint to build arguments

vigejolla:


vigejolla:
Added checking $SAILFISH_SDK_FRONTEND to mb2 PR, so this is actually not needed

PR #524, opened 2022-03-02
Jb57615

martyone:


share/qtcreator/sfdk/modules/20-building-mb2/doc/module.adoc
```diff
@@ -85,15 +85,31 @@ Similarly it is possible to supply any custom build-time dependencies by simply
 
 === Clean Builds
 
-Build time requirements of one package may interfere with those of another one. In order to avoid polluting build targets with dependencies of a particular package, sfdk uses build targets through writable snapshots. See the description of the 'snapshot' configuration option for details on how a snapshot is chosen.
+Build time requirements of one package may interfere with those of another one. Packages modified and deployed under the build environment in scope of one task may not be compatible with changes worked on in scope of other task. As time goes the build environment accumulates various changes, leading to generally non-reproducible build results.
```
vigejolla, line 152: I think it could be mentioned, preferably quite early, that most of the time developer does not have to think about the snapshots or how they work, as long as all the build time requirements are defined in the RPM 'SPEC' file.

share/qtcreator/sfdk/modules/20-building-mb2/doc/module.adoc
```diff
@@ -3,18 +3,19 @@
...
-    └─%{OsVariant:NoSpace}-3.0.0.8-i486
-    %{OsVariant:NoSpace}-3.0.2.8
-    ├─%{OsVariant:NoSpace}-3.0.2.8-armv7hl
-    └─%{OsVariant:NoSpace}-3.0.2.8-i486
+    %{OsVariant:NoSpace}-4.1.0.24
+    └─%{OsVariant:NoSpace}-4.1.0.24-i486
+    %{OsVariant:NoSpace}-4.2.0.21
+    ├─%{OsVariant:NoSpace}-4.2.0.21-aarch64
+    ├─%{OsVariant:NoSpace}-4.2.0.21-armv7hl
+    └─%{OsVariant:NoSpace}-4.2.0.21-i486
```
vigejolla, line 11: Why not go straight to 4.3.0.12?

share/qtcreator/sfdk/modules/20-building-mb2/doc/module.adoc
```diff
@@ -12,49 +12,113 @@ Start by choosing the build target. List the available build tools.
 
 Each build target lets you build software for a specific operating system version running on a specific hardware. Each target is listed under one tooling supporting the given OS version (see MANAGING SDK BUILD TOOLS below for more information).
 
-As with other configuration options, you can decide if you want to apply your choice globally or just for the current (shell) session. Let's pick the latest ARM target for global use and its i486 counterpart for this session only.
+Let's pick the latest 64-bit ARM target for global use.
```
vigejolla, line 15: 4.2.0.21 is not the latest target, maybe it's better to just drop the word.

vigejolla (approved):
A few comments, which you can ignore if you don't agree with them. Approving anyway, as I think this is already a clear improvement, even if you choose to ignore my comments.

share/qtcreator/sfdk/modules/20-building-mb2/doc/module.adoc
martyone: This is mean to refer to the latest target on user's list of installed targets. Added the words there.

share/qtcreator/sfdk/modules/20-building-mb2/doc/module.adoc
martyone: I use older example version numbers to make it more obvious that these are static examples hanging behind the reality.

share/qtcreator/sfdk/modules/20-building-mb2/doc/module.adoc
vigejolla: Yes, of course. I would have realized that if github review tool had not decided to cut the displayed text at just after the list. But I guess the added words don't harm.

share/qtcreator/sfdk/modules/20-building-mb2/doc/module.adoc
vigejolla: makes sense

share/qtcreator/sfdk/modules/20-building-mb2/doc/module.adoc
martyone: I was thinking how/where to fit such a note but have not found a feasible way. Also, it should not be needed. What are the issues with missing BuildRequires should be something the user learns very soon, both from the preceding parts of this manual and from their own experience. It is more likely to make troubles to power users or new users advised by old power users to manually deliver locally built dependencies, which is covered by

>  Use a shared output directory to supply locally built build-time dependencies without disturbance by (possibly implied) 'build-requires reset'. See WORKING ON MULTIPLE PACKAGES above.

It also speak in terms of 'build-requires reset', as snapshots are intentionally mentioned only below the "Detailed information follows." line.

PR #536, opened 2022-10-04
Sfdk: Hide snapshots in the listing unless told otherwise

martyone:


src/tools/sfdk/command.cpp
```diff
@@ -1930,22 +1930,35 @@ Worker::ExitStatus BuiltinWorker::runTools(const QStringList &arguments_, int *e
...
         if (!parser.parse(arguments)) {
             qerr() << parser.errorText() << endl;
             return BadUsage;
         }
 
+        if (!P::checkExclusiveOption(parser, {&availableOption, &snapshotsOption}))
+            return BadUsage;
+
+        if (parser.isSet(snapshotsOption) && typeHint == SdkManager::ToolingHint) {
+            qerr() << tr("The '%1' option can only be used with targets")
```
vigejolla, line 1946: As the `--snapshots` option can be used also without `targets`, i.e. `sfdk tools list --snapshots`, I would rephrase this as "The '%1' option can not be used with toolings".

share/qtcreator/sfdk/modules/20-building-mb2/doc/option.snapshot.adoc
```diff
@@ -4,8 +4,6 @@ By default, i.e., when this option is not set, the snapshot with suffix 'default
 
 When this option is set but no <suffix> was specified, '%pool' is assumed for compatibility with older sfdk versions. (Deprecated)
 
-When <suffix> is '%pool[,<N>]', a temporary snapshot will be used from a pool of the optionally given size <N> (default: 5).  The actual suffix will follow the pattern 'pool.XXX'.
+When <suffix> is '%pool[,<N>]', a temporary snapshot will be used from a pool of the optionally given size <N> (default: 5).  The actual suffix will follow the pattern 'pool.XXX'. Each temporary snapshot is used exclusively with single build tree at a time and it is reset automatically whenever it is taken into use with another build tree.
 
-Snapshot state is preserved across sfdk invocations unless 1) the original target was updated meanwhile (suppressed by the 'no-pull-build-requires' configuration option) or 2) a temporary snapshot is taken into use with other build tree than it was used with before.
-
-A snapshot can be reset forcefully using the 'build-requires reset' command.
+See the CLEAN BUILDS section in the domain help for more information.
```
vigejolla, line 9: Where can I find this "domain help" which has the "CLEAN BUILDS" section?

martyone: It's displayed on the same page. Which makes me realize the words "in the domain help" are redundant here. Those make sense for commands help pages, which are displayed separately from the all-in-one or per-domain manuals. Dropped.

src/tools/sfdk/command.cpp
martyone: yeah I wish sdk-assistant was never using the strange UI with the optional [tooling|target] "subcommand", which is actually an option and so should use the natural --option syntax instead. A few things would have been much easier and clearer, this being just one example. Fixed and aligned on sdk-assistant side as well.

share/qtcreator/sfdk/modules/20-building-mb2/doc/option.snapshot.adoc
vigejolla: I must be blind, because for me the only sections on that page are: "NAME", "SYNOPSIS", "DESCRIPTION", "GLOBAL OPTIONS", "CONFIGURATION OPTIONS", "HOOKS", "ENVIRONMENT VARIABLES" and "EXIT STATUS". No "CLEAN BUILDS" anywhere.

martyone: You are not blind, just looking for something else :) Look one level deeper under DESCRIPTION. Though the "typographical convention" throughout the text is to use uppercase to refer to the (sub)sections of the description, the titles themselves are Title Case. I didn't realize you would look that up the hard way - I expected you to use search and nowadays the pager does usually case-insensitive search by default :)

vigejolla: Found it! :)

Well, at least for me the pager is less, and it does case-sensitive searches by default. Only when it's run by man it uses case-insensitive searches. And since I wanted to have a look at the changes you had made, I just removed the previously generated man pages, thus forcing sfdk to display the help without the help from man, thus running less in case-sensitive mode... after regenerating the man pages, the search is indeed case-insensitive, assuming you don't write in caps. I mean, searching for "clean builds" finds the heading, but searching for "CLEAN BUILDS" doesn't. 

I think there's still something we can do to improve the probability of finding the section, but from the top of my head I don't know what.

vigejolla (approved):
LGTM

share/qtcreator/sfdk/modules/20-building-mb2/doc/option.snapshot.adoc
martyone: Indeed I never realized it is man which enables the case-insensitive search. Added a commit to let sfdk enable it with the plain less.

I also considered to pass the stronger option `--IGNORE-CASE` but that wouldn't be trivial to align to with the man viewer - it would require a bit of thinking and testing on mac and win, which is not a good time for now.

vigejolla (approved):
Beautiful! I trust that you tested this, I'm not going to.

martyone:
Yup, I did.

PR #538, opened 2022-11-14
Sfdk: Unify references to RPM SPEC files

vigejolla:


share/qtcreator/sfdk/modules/20-building-mb2/doc/command.apply.adoc
```diff
@@ -1,3 +1,3 @@
-Apply all patches defined in the RPM 'SPEC' file to the source tree.
+Apply all patches defined in the RPM SPEC file to the source tree.
```
martyone, line 1: Hm, did you really not like that the "SPEC" word is rendered as a keyword? (It is actually the markup for "Emphasized text" https://asciidoc-py.github.io/userguide.html#X51 but we use that for inline code/commands/keywords/... as it works better in some respects)

vigejolla: :facepalm: I just grepped through the sources and didn't look at how the text is actually rendered.

I didn't find any other references which wouldn't have been already "compliant". So I suppose we can just close this PR.

martyone: Ah that way :) There are actually two well hidden occurrences - in the manifest https://github.com/sailfishos/sailfish-qtcreator/blob/jb58855/share/qtcreator/sfdk/modules/20-building-mb2/manifest.json#L30 here and a bit below.

PR #542, opened 2022-11-23
Sfdk: Update paragraph about YAML

vigejolla:


martyone (commented):
I am not sure whether we should drop it immediately. There are existing packages which use YAML, so keeping at least some hint on what's going on would be good.

vigejolla:
> I am not sure whether we should drop it immediately. There are existing packages which use YAML, so keeping at least some hint on what's going on would be good.

Maybe you are right. However, I think we can't keep it with the current wording, as the files won't be there after sfdk init like the text implies. I'll try to come up with better wording.

martyone (approved):

PR #546, opened 2022-12-07
Sfdk: Set the nofile limits under Docker. JB#57802

martyone:


vigejolla (approved):
Approving, but I do have some questions: What are those apps which don't work well? Their behavior sounds quite weird to me. Maybe even buggy.

martyone:
Sounds buggy to me too. I just found a few issues reported about processes that fork and iterate through all possible FDs to close them, one report mentioned dnsmasq https://lists.thekelleys.org.uk/pipermail/dnsmasq-discuss/2020q1/013821.html, but in any case the effort to ensure we don't have any of those would hardly pay off, so I didn't look for more cases.

PR #549, opened 2023-01-11
Sfdk: Allow optional access to host devices. JB#59670

martyone:


vigejolla (approved):
LGTM. But thinking a bit further, perhaps instead of using privileged, we could allow just the loop device access with `--device-cgroup-rule="b 7:* rmw"` ?

martyone:
Nice tip. Strange I never hit a reference on this option while looking for possible solutions. There is also a need to use LVM, which would need something more flexible for proper filtering - haven't checked yet.

PR #550, opened 2023-02-01
Sfdk: Set GUI/LastNormalWindowPosition on setVideoMode

vigejolla:
VirtualBox 7.0 no longer uses GUI/LastGuestSizeHint, so we need to set both in order to make sure window size is correct

### DIFF
--- a/src/libs/sfdk/vboxvirtualmachine.cpp
+++ b/src/libs/sfdk/vboxvirtualmachine.cpp
@@ -523,7 +523,6 @@ void VBoxVirtualMachinePrivate::doSetStorageSizeMb(int storageSizeMb, const QObj
     qCDebug(vms) << "Changing storage size of" << q->uri().toString() << "to" << storageSizeMb << "MB";
 
     BatchComposer composer = BatchComposer::createBatch("VBoxVirtualMachinePrivate::doSetStorageSizeMb");
-    composer.batch()->setPropagateFailure(true);
 
     const QPointer<const QObject> context_{context};
     fetchInfo(VirtualMachineInfo::StorageInfo, Sdk::instance(),
--- a/src/libs/sfdk/vboxvirtualmachine.cpp
+++ b/src/libs/sfdk/vboxvirtualmachine.cpp
@@ -428,6 +428,22 @@ void VBoxVirtualMachinePrivate::setVideoMode(const QSize &size, int depth,
         setExtraData(key, value, Sdk::instance(), [=](bool ok) { *allOk &= ok; });
     };
 
+    {
+        BatchComposer composer = BatchComposer::createBatch("setLastNormalWindowPosition");
+        fetchExtraData("GUI/LastNormalWindowPosition", Sdk::instance(),
+                [=, batch = composer.batch()](const QString &data, bool ok) {
+            *allOk &= ok;
+            if (!ok)
+                return;
+            QTC_ASSERT(data.startsWith("Value:"), { *allOk = false; return; });
+            const QStringList splitData = data.section(':',1,1).trimmed().split(',');
+            QTC_ASSERT(splitData.size() > 1, { *allOk = false; return; });
+            const QString lastPosition = splitData[0] + ',' + splitData[1] + ','
+                + QString::number(size.width()) + ',' + QString::number(size.height());
+            BatchComposer composer = BatchComposer::extendBatch(batch);
+            enqueue("GUI/LastNormalWindowPosition", lastPosition);
+        });
+    }
     enqueue("CustomVideoMode1", videoMode);
     enqueue("GUI/LastGuestSizeHint", hint);
     enqueue("GUI/AutoresizeGuest", QLatin1Literal("false"));
--- a/src/libs/sfdk/vboxvirtualmachine.cpp
+++ b/src/libs/sfdk/vboxvirtualmachine.cpp
@@ -414,7 +414,6 @@ void VBoxVirtualMachinePrivate::setVideoMode(const QSize &size, int depth,
         << "depth" << depth;
 
     BatchComposer composer = BatchComposer::createBatch("VBoxVirtualMachinePrivate::setVideoMode");
-    composer.batch()->setPropagateFailure(true);
 
     QString videoMode = QStringLiteral("%1x%2x%3")
         .arg(size.width())


src/libs/sfdk/vboxvirtualmachine.cpp
```diff
@@ -429,6 +428,19 @@ void VBoxVirtualMachinePrivate::setVideoMode(const QSize &size, int depth,
...
+        QTC_ASSERT(data.startsWith("Value:"), return);
+        data = data.section(':',1,1).trimmed();
+        QStringList splitData = data.split(',');
+        QTC_ASSERT(splitData.size() > 1, return);
+        QString lastPosition = splitData[0] + ',' + splitData[1] + ','
+                + QString::number(size.width()) + ',' + QString::number(size.height());
+        BatchComposer composer = BatchComposer::extendBatch(batch);
+        enqueue("GUI/LastNormalWindowPosition", lastPosition);
+    });
+
```
martyone, line 443: This pair of `fetchExtraData` and (optional) `setExtraData` makes a batch and so we need to be explicit about it or the `setExtraData` would get enqueued after the checkpoint below.

```
{
    BatchComposer composer = BatchComposer::createBatch("setLastNormalWindowPosition");
    
    fetchExtraData("GUI/LastNormalWindowPosition", Sdk::instance(),
            [=, batch = composer.batch()](QString data, bool ok) {
        ... unchanged ...
    });
}
```

Compare with, e.g., https://github.com/sailfishos/sailfish-qtcreator/blob/master/src/libs/sfdk/vboxvirtualmachine.cpp#L785.

vigejolla: I did look at doSetReservedPortListForwarding() earlier, but I didn't notice that there were two createBatch()es :(

vigejolla: Now it seems to be working as it should :)

src/libs/sfdk/vboxvirtualmachine.cpp
```diff
@@ -429,6 +428,21 @@ void VBoxVirtualMachinePrivate::setVideoMode(const QSize &size, int depth,
         setExtraData(key, value, Sdk::instance(), [=](bool ok) { *allOk &= ok; });
     };
 
+    {
+        BatchComposer composer = BatchComposer::createBatch("setLastNormalWindowPosition");
+        fetchExtraData("GUI/LastNormalWindowPosition", Sdk::instance(), [=, batch = composer.batch()](QString data, bool ok) {
+            if (!ok)
```
martyone, line 434: Let's fix error reporting now. We need

```
*allOk &= ok;
```

prior to this line and then `{ *allOk = false; return; }` instead of just `return` in assertions.

src/libs/sfdk/vboxvirtualmachine.cpp
```diff
@@ -429,6 +428,21 @@ void VBoxVirtualMachinePrivate::setVideoMode(const QSize &size, int depth,
         setExtraData(key, value, Sdk::instance(), [=](bool ok) { *allOk &= ok; });
     };
 
+    {
+        BatchComposer composer = BatchComposer::createBatch("setLastNormalWindowPosition");
+        fetchExtraData("GUI/LastNormalWindowPosition", Sdk::instance(), [=, batch = composer.batch()](QString data, bool ok) {
```
martyone, line 433: Too long, I would prefer starting the lambda on next line.

src/libs/sfdk/vboxvirtualmachine.cpp
```diff
@@ -429,6 +428,21 @@ void VBoxVirtualMachinePrivate::setVideoMode(const QSize &size, int depth,
...
+            if (!ok)
+                return;
+            QTC_ASSERT(data.startsWith("Value:"), return);
+            data = data.section(':',1,1).trimmed();
+            QStringList splitData = data.split(',');
```
martyone, line 438: I would prefer joining this and making the three variables const (and ref in case of `data`).

martyone (commented):
Error handling and some nitpicks yet.

src/libs/sfdk/vboxvirtualmachine.cpp
vigejolla: That was left out on purpose. Even if this part of setVideoMode fails, the others might be enough, depending on VirtualBox version.

src/libs/sfdk/vboxvirtualmachine.cpp
vigejolla: fixed

src/libs/sfdk/vboxvirtualmachine.cpp
vigejolla: fixed

src/libs/sfdk/vboxvirtualmachine.cpp
martyone: Ah, I guess you assume it would cancel the other parths. That's not the case here. It would just indicate that not all went as planned to the user. So I would prefer to be very specific about the two scenarios we expect here, branch for old and new version of vbox we currently know about and be verbose in other cases.

Note: When I tried with vbox 6.1.36r152435, the GUI/LastNormalWindowPosition property is set for my emulators, so this should not fail normally. Also when I tried with nonexisting property, it still returned zero, just the message is "No value set!". Older emulators will get reinstalled this time, because we are touching the emulator package, so our emulators should have that property set after upgrade.

vigejolla: > Ah, I guess you assume it would cancel the other parths. That's not the case here. It would just indicate that not all went as planned to the user. So I would prefer to be very specific about the two scenarios we expect here, branch for old and new version of vbox we currently know about and be verbose in other cases.

You are again right. I wonder how I could think that, when the code is right in front of me?

> Note: When I tried with vbox 6.1.36r152435, the GUI/LastNormalWindowPosition property is set for my emulators, so this should not fail normally. Also when I tried with nonexisting property, it still returned zero, just the message is "No value set!". Older emulators will get reinstalled this time, because we are touching the emulator package, so our emulators should have that property set after upgrade.

Thanks for testing!

martyone (approved):
Would be nice to fix the misplaced `&` in lambda argument list and wrong indentation of that line :)

PR #561, opened 2026-04-30
CMake: Do not use Ninja when it is available on host. JB#63214

martyone:


martyone:

rainemak (approved):

PR #563, opened 2026-05-06
Sfdk: Use correct path to gdb on macOS

martyone:


martyone:

rainemak (approved):

PR #565, opened 2026-05-11
Sfdk: Look up docker path in registry too

martyone:


martyone:
Just the second commit belongs to this PR.

src/libs/sfdk/dockervirtualmachine.cpp
```diff
@@ -61,6 +62,19 @@ QString dockerPath()
...
+    if (HostOsInfo::isWindowsHost()) {
+        // Not found in environment? Look up registry.
+        QSettings s(QLatin1String("HKEY_LOCAL_MACHINE\\SOFTWARE\\Docker Inc.\\Docker\\1.0"),
+                    QSettings::NativeFormat);
+        path = s.value(QLatin1String("BinPath")).toString();
+        if (path.startsWith(QLatin1Char('"')) && path.endsWith(QLatin1Char('"')))
+            path = path.mid(1, path.length() - 2); // remove quotes
+        if (!path.isEmpty())
+            path.append(QDir::separator() + QLatin1String(DOCKER));
+    }
```
direc85, line 77:
```suggestion
#if defined(Q_OS_WIN)
    if (path.isEmpty())
        // Not found in environment? Look up registry.
        QSettings s(QLatin1String("HKEY_LOCAL_MACHINE\\SOFTWARE\\Docker Inc.\\Docker\\1.0"),
                    QSettings::NativeFormat);
        path = s.value(QLatin1String("BinPath")).toString();
        if (path.startsWith(QLatin1Char('"')) && path.endsWith(QLatin1Char('"')))
            path = path.mid(1, path.length() - 2); // remove quotes
        if (!path.isEmpty())
            path.append(QDir::separator() + QLatin1String(DOCKER));
    }
#endif
```
And ifdef for the QSettings include as well. A bit cleaner this way I think and avoids unnecessary include on other OS'es, and one extra return.

direc85 (approved):
One suggestion to get a little bit cleaner implementation, but fine as-is as well.

src/libs/sfdk/dockervirtualmachine.cpp
martyone: It is actually a coding convention in qtc IIRC and good practice in general to avoid use of preprocessor conditionals. It makes life easier because all code is checked for compile-time errors no matter which platform you develop on (imagine you change an API used under a per-platform #if-else and then you need to try to compile on all platforms to ensure... And given that the condition is constant, the code will be omitted in the output equally.

